PR #6, opened 2018-01-26
[11.0][ADD] easy_switch_user

tarteo:
This module lets administrators and developers quickly change user to test e.g. access rights.

### DIFF
--- a/easy_switch_user/README.rst
+++ b/easy_switch_user/README.rst
@@ -1,0 +1,59 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+================
+Easy Switch User
+================
+
+This module lets administrators and developers quickly change user to test e.g. access rights.
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Click on the caret in the system tray
+#. Select the user you want to switch to
+#. Login (only required once)
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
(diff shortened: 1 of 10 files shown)


LoisRForgeFlow:
Probably not due to this module, but I can't access runbot right now.

StefanRijnhart (commented):
I like this module! The code looks clean as well. What I don't like is that the entered passwords are stored unencrypted in the session storage on the server, including the password of the original user when they switch back. Can you make it so that the cookie only contains the list of authenticated users and bypass the password check when switching users?

tarteo:
@StefanRijnhart The passwords are not stored server side. I use html5 session storage which is stored client side. This storage is cleared when you close the browser. https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage

StefanRijnhart:
Indeed, I misunderstood the concept of session storage, and the cleartext password of the current user is put in the server side session by Odoo itself. Thanks for clarifying! I guess this is basically alright for security, but maybe this mechanism can be made explicit and/or optional? I'm thinking of a checkbox

    [x] Store this password in your browser until you close this tab

So that you cannot trick other users to enter their password to switch to their account temporarily and harvest their passwords.

Another thing: can you clear this data from the session storage upon logout?

tarteo:
@StefanRijnhart Yes sounds good! I'll add it. Thanks for your feedback. :)

tarteo:
@StefanRijnhart Hey, I've added the checkbox and clearing the session storage opun logout. Can you re-review ;) when you have time?

StefanRijnhart (approved):
Thank you for the quick followup and the nice module!

LoisRForgeFlow (approved):
Tested locally, great one, thanks! :+1:

astirpe (approved):
Code review: OK

PR #9, opened 2018-03-17
[11.0] [BUG] Fix date range type unlink.

feketemihai:


date_range/models/date_range_type.py
```diff
@@ -26,3 +26,10 @@ def _default_company(self):
     _sql_constraints = [
         ('date_range_type_uniq', 'unique (name,company_id)',
          'A date range type must be unique per company !')]
+
+    @api.multi
+    def unlink(self):
```
pedrobaeza: Instead of this, use `ondelete='restrict'` in the many2one in date.range.

feketemihai: @pedrobaeza I know, but i wanted to track the exact error, and to benefit from it to super method in https://github.com/OCA/account-financial-tools/pull/636, because there the super was not covered in tests, red build, and i only had an postgres integrity error, and wanted to track it as odoo error..

pedrobaeza: You can test it the same. Just use:

```python
with self.assertRaises(Exception):
    with self.env.cr.savepoint():
        self.date_range_type.unlink()
```

feketemihai: @pedrobaeza Updated, but i think is fine if i leave also the unlink method, to have a nice error message for the user.

pedrobaeza: But this is code to maintain. Generic error is clear enough IMO.

MiquelRForgeFlow (approved):

pedrobaeza:
It would be convenient to split in 2 commits: one for the unlink problem, and another for pylint warnings, coding removal, etc.

sbidoul (approved):
Splitting commits would is nice to have, but otherwise :+1:

feketemihai:
I will do it later. Thanks for review.

MiquelRForgeFlow:
:green_apple:

pedrobaeza:
Thanks

PR #10, opened 2018-03-23
[11.0][IMP] date_range: adapt to multicompany

MiquelRForgeFlow:
This PR adapts the `date_range` module to a multicompany environment.

It also provides tests for this improvement and migration files.

I hope all you like it.


MiquelRForgeFlow:
@etobella @jbeficent

JordiBForgeFlow (approved):
:+1: Tested. Works like a charm!

date_range/models/date_range_type.py
```diff
@@ -25,3 +27,19 @@ def _default_company(self):
...
         ('date_range_type_uniq', 'unique (name,company_id)',
          'A date range type must be unique per company !')]
+
+    @api.constrains('company_id')
+    def _check_company_id(self):
+        if not self.env.context.get('bypass_company_validation', False):
+            for rec in self:
+                if not rec.company_id:
+                    continue
+                field = self.env['date.range'].search(
```
lmignon: @mreficent  you could maybe avoid a search into a loop  by using a `read_group` (not bloking)

lmignon (approved):
Thank you for this improvement. (Code review only)

date_range/models/date_range_type.py
feketemihai: Probably it's better to have a one2many to date_range and use mapped on company_id.

MiquelRForgeFlow: Yeah, I like this one2many approach. I will move this way.

pedrobaeza:
@jbeficent please wait until CIs are green before merging.

PR #12, opened 2018-04-14
[FIX][mass_editing] v11 Fixing error when removing value for fields b…

BT-rmartin:
…eing translatable, as their translations were not removed.

### DIFF
--- a/mass_editing/__manifest__.py
+++ b/mass_editing/__manifest__.py
@@ -11,7 +11,7 @@
     'license': 'AGPL-3',
     'summary': 'Mass Editing',
     'uninstall_hook': 'uninstall_hook',
-    'depends': ['base', 'mail'],
+    'depends': ['base'],
     'data': [
         'security/ir.model.access.csv',
         'views/mass_editing_view.xml',
--- a/mass_editing/__manifest__.py
+++ b/mass_editing/__manifest__.py
@@ -5,6 +5,7 @@
     'version': '11.0.1.0.0',
     'author': 'Serpent Consulting Services Pvt. Ltd., '
               'Tecnativa, '
+              'brain-tec AG, '
               'Odoo Community Association (OCA)',
     'category': 'Tools',
     'website': 'http://www.serpentcs.com',
--- a/mass_editing/__manifest__.py
+++ b/mass_editing/__manifest__.py
@@ -12,7 +12,7 @@
     'license': 'AGPL-3',
     'summary': 'Mass Editing',
     'uninstall_hook': 'uninstall_hook',
-    'depends': ['base', 'mail'],
+    'depends': ['base'],
     'data': [
         'security/ir.model.access.csv',
         'views/mass_editing_view.xml',
--- a/mass_editing/__manifest__.py
+++ b/mass_editing/__manifest__.py
@@ -5,7 +5,6 @@
     'version': '11.0.1.0.0',
     'author': 'Serpent Consulting Services Pvt. Ltd., '
               'Tecnativa, '
-              'brain-tec AG, '
               'Odoo Community Association (OCA)',
     'category': 'Tools',
     'website': 'http://www.serpentcs.com',
--- a/mass_editing/__manifest__.py
+++ b/mass_editing/__manifest__.py
@@ -11,7 +11,7 @@
     'license': 'AGPL-3',
     'summary': 'Mass Editing',
     'uninstall_hook': 'uninstall_hook',
-    'depends': ['base'],
+    'depends': ['base', 'mail'],
     'data': [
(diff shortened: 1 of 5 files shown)


BT-rmartin:
@pedrobaeza Hi Pedro, could you please have a look at this? It's exactly the same functionality as the one you merged right now. Thanks

mass_editing/models/mass_object.py
```diff
@@ -77,6 +77,7 @@ def unlink(self):
         self.unlink_action()
         return super(MassObject, self).unlink()
 
+    @api.multi
```
pedrobaeza: No need for this. It's the default

mass_editing/__manifest__.py
```diff
@@ -5,6 +5,7 @@
     'version': '11.0.1.0.0',
     'author': 'Serpent Consulting Services Pvt. Ltd., '
               'Tecnativa, '
+              'brain-tec AG, '
```
pedrobaeza: For this newest version, I don't consider your contribution to be upgraded to co-authorship. Please keep only the contributor in the README part.

mass_editing/tests/test_mass_editing.py
```diff
@@ -16,6 +16,9 @@ def setUp(self):
         self.mass_wiz_obj = self.env['mass.editing.wizard']
         self.mass_object_model = self.env['mass.object']
         self.res_partner_model = self.env['res.partner']
+        self.ir_translation_model = self.env['ir.translation']
+        self.lang_model = self.env['res.lang']
+
```
pedrobaeza: Don't add empty extra lines in the middle of a method.

mass_editing/tests/test_mass_editing.py
```diff
@@ -38,33 +53,52 @@ def _create_partner(self):
             'email': 'example@yourcompany.com',
             'phone': 123456,
             'category_id': [(6, 0, categ_ids)],
+            'notify_email': 'always'
+        })
+
+    def _create_partner_title(self):
+        """Create a Partner Title."""
+
```
pedrobaeza: Remove all the empty lines

mass_editing/tests/test_mass_editing.py
```diff
@@ -38,33 +53,52 @@ def _create_partner(self):
...
             'email': 'test@test.com',
         })
 
-    def _create_mass_editing(self, model, fields):
+    def _create_mass_editing(self, model, fields, model_name):
         """Create a Mass Editing with Partner as model and
         email field of partner."""
         mass = self.mass_object_model.create({
-            'name': 'Mass Editing for Partner',
+            'name': u'Mass Editing for {0}'.format(model_name),
```
pedrobaeza: In Python 3, there's no more `u''`, as everything is unicode.

pedrobaeza (changes requested):

mass_editing/models/mass_object.py
BT-rmartin: Travis was complaining about it was missing, at least in previous versions

mass_editing/tests/test_mass_editing.py
BT-rmartin: Why not? We separate model connections, to the logic itself. IMHO it improves legibility.

mass_editing/tests/test_mass_editing.py
BT-rmartin: Again, I'm for them, to improve legibility. Any reason behind this suggestion?

mass_editing/tests/test_mass_editing.py
BT-rmartin: Ok, thanks

mass_editing/__manifest__.py
BT-rmartin: Ok, I understand this. Thanks for the suggestion

mass_editing/tests/test_mass_editing.py
pedrobaeza: No, it only adds unneeded lines. PEP8 recommends not to put them. If you want to add a line with a comment, OK. That's the way to separate things inside the same method. Adding empty lines, not.

mass_editing/models/mass_object.py
pedrobaeza: I'm sure it doesn't.

mass_editing/tests/test_mass_editing.py
BT-rmartin: I already removed the empty lines. Thanks for the suggestion

pedrobaeza (approved):
Fast-tracking as it has been already merged in v10.

PR #13, opened 2018-04-30
[MIG][11.0] base_optional_quick_create

fmdl:


fmdl:
@alnslang

elicoidal (approved):
LGTM

pedrobaeza:
@fmdl trying to fix the commit history, this has result in an empty PR and auto-closed. Follow-up is on #14. Please never make PRs from the main branch (11.0), which has caused the problem. Use an auxiliary branch.

PR #15, opened 2018-05-09
[11.0][MIG] base_tier_validation

LoisRForgeFlow:
Migration to v11. Based on https://github.com/OCA/server-tools/pull/1203

### DIFF
--- a/base_tier_validation/__init__.py
+++ b/base_tier_validation/__init__.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from . import models
--- a/base_tier_validation/__init__.py
+++ b/base_tier_validation/__init__.py
@@ -1,0 +1,4 @@
+# -*- coding: utf-8 -*-
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from . import models
--- a/base_tier_validation/__manifest__.py
+++ b/base_tier_validation/__manifest__.py
@@ -1,12 +1,11 @@
-# -*- coding: utf-8 -*-
 # Copyright 2017 Eficent Business and IT Consulting Services S.L.
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 {
     "name": "Base Tier Validation",
     "summary": "Implement a validation process based on tiers.",
-    "version": "10.0.1.0.0",
+    "version": "11.0.1.0.0",
     "category": "Tools",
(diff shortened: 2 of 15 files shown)


base_tier_validation/__init__.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
```
MiquelRForgeFlow: in all `__init__` files you should remove full header

base_tier_validation/__manifest__.py
```diff
@@ -1,10 +1,9 @@
-# -*- coding: utf-8 -*-
 # Copyright 2017 Eficent Business and IT Consulting Services S.L.
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 {
     "name": "Base Tier Validation",
     "summary": "Implement a validation process based on tiers.",
-    "version": "10.0.1.0.0",
+    "version": "11.0.1.0.0",
     "category": "Tools",
     "website": "https://github.com/OCA/server-tools",
```
MiquelRForgeFlow: -> `server-ux`

MiquelRForgeFlow (changes requested):

base_tier_validation/models/tier_validation.py
```diff
@@ -90,7 +89,7 @@ def _compute_need_validation(self):
     def evaluate_tier(self, tier):
         try:
             res = safe_eval(tier.python_code, globals_dict={'rec': self})
-        except Exception, error:
+        except Exception as error:
```
MiquelRForgeFlow: you can remove `as error`

LoisRForgeFlow: I use it below

lreficent:
WIP

base_tier_validation/__init__.py
LoisRForgeFlow: License must be kept: https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py

base_tier_validation/__manifest__.py
LoisRForgeFlow: right!

base_tier_validation/models/tier_validation.py
MiquelRForgeFlow: Ah, true, I checked too fast.

MiquelRForgeFlow (approved):

JordiBForgeFlow (approved):

JordiBForgeFlow:
Tested. Test Coverage is now very good.

PR #18, opened 2018-06-28
[IMP] date_range: new README structure

lmignon:
same as https://github.com/OCA/server-tools/pull/1305


pedrobaeza (approved):
Fast-tracking it when CIs finish

PR #21, opened 2018-07-04
[ADD] barcode_action

etobella:
This module allows to use barcodes as launchers of actions.

The action will launch a function that uses the barcode in order to return an action.

Usage
=====
Actions must be configured with the following data in the context:
* model: Model where we can find the method (required)
* method: Method to execute (required)
* res_id: Id as base (optional)

The method must return an action. Installing this module with demo data will
install a demo application that allows the system administrator to find a
partner by the external reference encoded in a barcode.

Go to *Settings / Find partners* and scan a barcode that contains the
internal reference of an existing partner. As soon as you read the barcode
the system will redirect you to that partner's form view.

Technical implementation of this example:

Action::

        <act_window id="res_partner_find"
            name="Find Partner"
            res_model="barcode.action"
            view_mode="form"
            view_type="form"
            context="{'default_model': 'res.partner', 'default_method': 'find_res_partner_by_ref_using_barcode'}"
            target="new"/>

        <menuitem id="menu_orders_customers" name="Find partners"
            action="res_partner_find"
            parent="base.menu_administration"/>

Python code::

    import json
    from odoo import api, models, _
    from odoo.tools.safe_eval import safe_eval

    class ResPartner(models.Model):
        _inherit = 'res.partner'

        @api.multi
        def find_res_partner_by_ref_using_barcode(self, barcode):
            partner = self.search([('ref', '=', barcode)], limit=1)
            if not partner:
                action = self.env.ref('res_partner_find')
                result = action.read()[0]
                context = safe_eval(result['context'])
                context.update({
                    'default_state': 'warning',
                    'default_status': _('Partner with Internal Reference '
                                        '%s cannot be found') % barcode
                })
                result['context'] = json.dumps(context)
                return result
            action = self.env.ref('base.action_partner_form')
            result = action.read()[0]
            res = self.env.ref('base.view_partner_form', False)
            result['views'] = [(res and res.id or False, 'form')]
            result['res_id'] = partner.id
            return result


JordiBForgeFlow (approved):
:+1: technical and functional review. Amazing stuff!

JordiBForgeFlow:
@hbrunn @tarteo You'll love this one

hbrunn:
looks nice indeed. Tests are missing though.

etobella:
@hbrunn Tests have been added.

barcode_action/README.rst
```diff
@@ -0,0 +1,21 @@
+**This file is going to be generated by oca-gen-addon-readme.**
```
tarteo: Can you generate a temporary readme?

barcode_action/static/src/js/action_barcode_form.js
```diff
@@ -0,0 +1,28 @@
...
+            return new $.Deferred().reject();
+        }
+        var record = this.model.get(this.handle);
+        var self = this;
+        return self._rpc({
+                    model: record.data.model,
+                    method: record.data.method,
+                    args: [[record.data.res_id], barcode],
+                }).done(function (action) {
+                    if (action !== undefined){
```
tarteo: action is here always defined, right? Maybe it's better to change it to `if(action) { }`

tarteo (changes requested):
Very cool module!!

etobella:
@tarteo Changes applied

tarteo (approved):

barcode_action/models/res_partner.py
```diff
@@ -0,0 +1,31 @@
...
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+import json
+from odoo import api, models, _
+from odoo.tools.safe_eval import safe_eval
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    @api.multi
```
astirpe, line 12: This should be `@api.model`.
(also in the readme)

barcode_action/__manifest__.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2017 Creu Blanca
+# Copyright 2017 Eficent Business and IT Consulting Services, S.L.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+{
+    'name': 'Barcode action launcher',
+    'version': '11.0.1.0.0',
+    'category': 'Reporting',
```
astirpe: Any other appropriate category? Maybe `'Extra Tools' `?

barcode_action/wizard/barcode_action.py
```diff
@@ -0,0 +1,22 @@
...
+    )
+    res_id = fields.Integer()
+    method = fields.Char(
+        required=True, readonly=True
+    )
+    state = fields.Selection([
+        ('waiting', 'Waiting'),
+        ('warning', 'Warning')
+    ], default='waiting', readonly=True, )
+    status = fields.Text(
```
astirpe, line 19: Fields names are: `state` and `status` ... why not using just one field? :thinking:

barcode_action/wizard/barcode_action_view.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0"?>
```
astirpe: Missing `encoding="utf-8"`

barcode_action/static/src/js/action_barcode_form.js
```diff
@@ -0,0 +1,28 @@
+odoo.define('barcode_action.form', function (require) {
+"use strict";
+
+var FormController = require('web.FormController');
+
+FormController.include({
+    _barcodeHandleAction: function (barcode, activeBarcode) {
+        if (this.mode === 'readonly') {
+            this.do_warn(_t('Error : Document not editable'),
```
astirpe: Is this check necessary?

astirpe (changes requested):

barcode_action/wizard/barcode_action.py
etobella: Well, `status` should be used in order to expand why you have a `state`. At the end, `state` will be used in order to select a color and `status` as the text inside the colored box.

barcode_action/static/src/js/action_barcode_form.js
etobella: Mmmmh, good point, I followed the same logic used on [odoo](https://github.com/odoo/odoo/blob/11.0/addons/barcodes/static/src/js/barcode_form_view.js#L62), but I could remove it if you feel that it is unnecessary.

barcode_action/models/res_partner.py
etobella: I do not agree, it should be multi, because we would like to use the res_id as part of our process. Usually it could be a @api.model (As in the example), but in some cases, it will be a @api.multi with a single record. For example, If we have a barcode for products, inside a SO we could add the "Add product". When you scan the product, it will be added on your SO, so the function will be assigned in `sale.order` and as a `res_id` you will send the active_id.

astirpe: Interesting! Then @api.multi is fine.

barcode_action/wizard/barcode_action.py
astirpe: OK

barcode_action/static/src/js/action_barcode_form.js
astirpe: I feel like it's not necessary. Not blocking anyway.

astirpe (approved):

barcode_action/static/src/js/action_barcode_form.js
etobella: Removed

PR #25, opened 2018-10-01
[12.0][MIG] date_range

astirpe:


espo-tony (approved):
I tested it and works fine. Code looks good too. :+1:

date_range/models/date_range.py
```diff
@@ -22,7 +22,7 @@ def _default_company(self):
         ondelete='restrict', domain="['|', ('company_id', '=', company_id), "
                                     "('company_id', '=', False)]")
     type_name = fields.Char(
-        string='Type', related='type_id.name', readonly=True, store=True)
+        related='type_id.name', readonly=True, store=True, string="Type Name")
```
lmignon, line 25: @astirpe  you must remove the string since "Type Name" is the default string for a field named type_name.

lmignon (changes requested):

lmignon:
@astirpe Thank you for the migration. A little change to do otherwise  LGTM

date_range/models/date_range.py
pedrobaeza: I have put it precisely because now Odoo checks for the existence of 2 labels with the same name, so you have "Name" and "Name", and this avoids the warning.

astirpe: @pedrobaeza the conflicts was on string='Type', not on Name. Removing all the string="..." fixes the warning to me.

astirpe: Oh and... now I know who is the one that is changing the code :laughing: @pedrobaeza

pedrobaeza: Yes, I have squashed a bit the commit history and fixed that problem for moving forward with this. I hope you don't bother for that.

astirpe: @pedrobaeza no problem of course! I'm going to remove string="Type Name" anyway. Do you have objections?

lmignon: @pedrobaeza PLZ don't squash the history on module where you are not the maintainer. :smirk:

date_range/models/date_range.py
```diff
@@ -0,0 +1,93 @@
...
+        default=_default_company)
+    active = fields.Boolean(
+        help="The active field allows you to hide the date range without "
+        "removing it.", default=True)
+
+    _sql_constraints = [
+        ('date_range_uniq', 'unique (name,type_id, company_id)',
+         'A date range must be unique per company !')]
+
+    @api.onchange('company_id')
```
SimoRubi: should depend on `type_id` too

date_range/models/date_range_type.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @api.model
+    def _default_company(self):
+        return self.env['res.company']._company_default_get('date.range')
+
+    name = fields.Char(required=True, translate=True)
+    allow_overlap = fields.Boolean(
+        help="If sets date range of same type must not overlap.",
+        default=False)
+    active = fields.Boolean(
```
SimoRubi, line 20: [...] to hide the date range **type** without [...]

date_range/models/date_range_type.py
```diff
@@ -0,0 +1,44 @@
...
+        ('date_range_type_uniq', 'unique (name,company_id)',
+         'A date range type must be unique per company !')]
+
+    @api.constrains('company_id')
+    def _check_company_id(self):
+        if not self.env.context.get('bypass_company_validation', False):
+            for rec in self.sudo():
+                if not rec.company_id:
+                    continue
+                if bool(self.date_range_ids.filtered(
```
SimoRubi: I think you don't need the cast to `bool`.
Check on `rec` instead of `self`.

date_range/models/date_range_type.py
```diff
@@ -0,0 +1,44 @@
...
+                        lambda r: r.company_id and
+                        r.company_id != rec.company_id)):
+                    raise ValidationError(
+                        _('You cannot change the company, as this '
+                          'Date Range Type is  assigned to Date Range '
+                          '(%s).') % (self.date_range_ids.name_get()[0][1]))
```
SimoRubi: Use `rec` instead of `self`.

date_range/readme/INSTALL.rst
```diff
@@ -0,0 +1,3 @@
+The addon use the daterange method from postgres. This method is supported as of postgresql 9.2
+
+
```
tarteo: Too many new lines.

date_range/static/src/js/date_range.js
```diff
@@ -0,0 +1,125 @@
...
+    get_value: function () {
+        return parseInt(this.$el.val());
+    },
+
+    on_range_selected: function (e){
+        var self = this;
+        self.domain = '';
+        framework.blockUI();
+        return rpc.query({
+                args: [this.get_value()],
```
tarteo, line 105: Can you fix the indentation here.

date_range/static/src/js/date_range.js
```diff
@@ -0,0 +1,125 @@
...
+    },
+
+    toString: function () {
+        var select = this.$el[0];
+        var option = select.options[select.selectedIndex];
+        return option.label || option.text;
+    },
+
+    get_value: function () {
+        return parseInt(this.$el.val());
```
tarteo: Can you specify the radix.

date_range/static/src/js/date_range.js
```diff
@@ -0,0 +1,125 @@
...
+            // want to put nest the returned value into an array.
+            res.attrs.domain = this.value.domain;
+        }
+        return res;
+    }
+
+});
+
+/**
+Since Odoo 11, The Field class used as base class for all specialized filter
```
tarteo, line 73: Is this still the case in Odoo 11? If so can you change it to Odoo 12

date_range/static/src/js/date_range.js
```diff
@@ -0,0 +1,125 @@
...
+            var type_id = val.replace('drt_', '');
+            this.date_range_type_operator_selected(type_id);
+            return;
+        }
+        this._super.apply(this, arguments);
+    },
+
+    date_range_type_operator_selected: function (type_id){
+        this.$value.empty().show();
+        var ds = new data.DataSetSearch(this, 'date.range', this.context, [['type_id', '=', parseInt(type_id)]]);
```
tarteo: Can you specify the radix?

SimoRubi (changes requested):
Thanks for your contribution!
Please check the warning thrown by runbot too

tarteo (changes requested):

date_range/models/date_range_type.py
lmignon: @SimoRubi I think you misunderstand the code..... 
We check if a date_range with another company `lambda r: r.company_id and  r.company_id != rec.company_id` already exists into the list of date ranges (self.date_range_ids) with the same type.

SimoRubi:
@astirpe also note that the folder migrations has to be removed as stated in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0#tasks-to-do-in-the-migration:

> Remove any possible migration script from previous version.

date_range/models/date_range_type.py
lmignon: @SimoRubi see my previous comment. This code is right.

date_range/models/date_range_type.py
SimoRubi: Thanks for the clarification @lmignon, but if `self` is a recordset, then wouldn't `self.date_range_ids` throw an Exception?

date_range/static/src/js/date_range.js
astirpe: I would keep it as it is, meaning that:
_starting from V11, the Field class used as base class for all specialized filter etc..._

astirpe:
@SimoRubi I think that the actual migration script (`11.0.2.0.0`) should be kept as it is, to handle the case of migration directly from `11.0.1.0.0` to `12.0.1.0.0`.

date_range/models/date_range_type.py
lmignon: @SimoRubi You're right, the logic here is not right.... may be something like
```python
for rec in self.sudo():
    company_ids = filtered(None, rec.date_range_ids.mapped('company_id'))
    if company_ids and company_ids != rec.company_id:
        raise ....
```
to be tested

lmignon: But replacing self by rec is even more easy and readable ;-)

astirpe: With latest commit, the self is replaced by rec. Is that enough? Do you all agree?

pedrobaeza:
@astirpe No, the rule is to always be on latest updated version of N-1 module for migrating to N. If not, you will get a lot of side effects. Trust me on this one :wink:

SimoRubi (approved):
ok for me

SimoRubi (changes requested):
Please check runbot

date_range/models/date_range.py
astirpe: @lmignon Pedro was right, we need to write string="Type Name" explicitly, otherwise runbot gives the following warning:

WARNING openerp_test odoo.addons.base.models.ir_model: Two fields (type_name, name) of date.range() have the same label: Name.

astirpe:
Migration scripts are removed.
The `string="Type Name"` is restored.
Latest commits are squashed.

ThomasBinsfeld (approved):

lmignon (approved):
@astirpe It could be nice to backport the fix into the date_range_type constrains in previous version....

tarteo (approved):
Thanks!

SimoRubi (approved):
Ok for me

astirpe:
The code review done within this PR is backported to V11: https://github.com/OCA/server-ux/pull/26

mpanarin (approved):

PR #27, opened 2018-10-02
[ADD][11.0] Add base_filter_hidden module

acsonefho:
- [x] New module "base_filter_hidden";
- [x] Add new field "is_system" into ir.filters (these filters are not loaded into web-interface (views));
- [x] Unit test added;

The main goal of this MR is to hide (from views) some `ir.filters` define as system filters. So they'll be used into a `M2O` (for example) into some recordset.

Any better ideas/improvements are welcome


sebalix:
A related module for Odoo 8 by @hbrunn : https://github.com/OCA/server-tools/tree/8.0/base_ir_filters_active
Are they complementary or yours is a better implementation?

acsonefho:
@sebalix Thank you for your comment.
The MR in v8.0 is not the same (and the added `active` field already exists in v11.0).
The goal of this MR is to hide some "technical" `ir.filters` without inactive them, because we'll use it. We just don't need to display them into the search view.

sebalix (approved):

lmignon (approved):
Great usability improvement when you use filters for some internal business processes as in https://github.com/OCA/product-attribute/pull/376/files

base_filter_hidden/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2018 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': "Base filter hidden",
+    'description': """
```
NL66278, line 5: Description key is deprecated, please remove.

base_filter_hidden/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2018 ACSONE SA/NV (<http://acsone.eu>)
```
NL66278, line 1: Use https: for copyright and licence lines. (Also in the other files).

NL66278 (changes requested):
I agree the module might be very usefull, but I think it can also be more simple and efficient.

As it is the module reads all the normal filters, then does work to remove the system filters.

It would be more efficient to call super in the get_filters method with some context key that system filters are not wanted. Then override the search method to add a criterium to the domain, if the context key is set, to not select system filters.

bealdav:
I think @NL66278 remark could be added to roadmap and merge as it : tests and approved.

PR #28, opened 2018-10-07
[12.0][FIX] date_range: at_install/post_install test decorators

astirpe:
Following https://github.com/odoo/odoo/issues/27471, here is the patch for the `at_install`/`post_install` test decorators, the V12.0 way.

### DIFF
--- a/date_range/tests/test_date_range.py
+++ b/date_range/tests/test_date_range.py
@@ -3,13 +3,13 @@
 
 import datetime
 
+import odoo
 from odoo.exceptions import ValidationError
 from odoo.tests.common import TransactionCase
 
 
+@odoo.tests.tagged('post_install', '-at_install')
 class DateRangeTest(TransactionCase):
-    post_install = True
-    at_install = False
 
     def setUp(self):
         super(DateRangeTest, self).setUp()
--- a/date_range/tests/test_date_range.py
+++ b/date_range/tests/test_date_range.py
@@ -3,12 +3,10 @@
 
 import datetime
 
-import odoo
 from odoo.exceptions import ValidationError
 from odoo.tests.common import TransactionCase
 
 
-@odoo.tests.tagged('post_install', '-at_install')
 class DateRangeTest(TransactionCase):
 
     def setUp(self):
--- a/date_range/tests/test_date_range_generator.py
+++ b/date_range/tests/test_date_range_generator.py
@@ -5,12 +5,10 @@ import datetime
 
 from dateutil.rrule import MONTHLY
 
-import odoo
 from odoo.exceptions import ValidationError
 from odoo.tests.common import TransactionCase
 
 
-@odoo.tests.tagged('post_install', '-at_install')
 class DateRangeGeneratorTest(TransactionCase):
 
     def setUp(self):
--- a/date_range/tests/test_date_range_generator.py
+++ b/date_range/tests/test_date_range_generator.py
@@ -5,13 +5,13 @@ import datetime
 
 from dateutil.rrule import MONTHLY
 
+import odoo
 from odoo.exceptions import ValidationError
 from odoo.tests.common import TransactionCase
 
 
+@odoo.tests.tagged('post_install', '-at_install')
 class DateRangeGeneratorTest(TransactionCase):
-    post_install = True
-    at_install = False
 
     def setUp(self):
         super(DateRangeGeneratorTest, self).setUp()
--- a/date_range/tests/test_date_range_type.py
+++ b/date_range/tests/test_date_range_type.py
@@ -3,14 +3,14 @@
 
 from psycopg2 import IntegrityError
 
+import odoo
 from odoo.exceptions import ValidationError
 from odoo.tests.common import TransactionCase
 from odoo.tools import mute_logger
 
 
+@odoo.tests.tagged('post_install', '-at_install')
 class DateRangeTypeTest(TransactionCase):
-    post_install = True
-    at_install = False
 
     def setUp(self):
         super(DateRangeTypeTest, self).setUp()
--- a/date_range/tests/test_date_range_type.py
+++ b/date_range/tests/test_date_range_type.py
@@ -3,13 +3,11 @@
 
 from psycopg2 import IntegrityError
 
-import odoo
 from odoo.exceptions import ValidationError
 from odoo.tests.common import TransactionCase
 from odoo.tools import mute_logger
 
 
-@odoo.tests.tagged('post_install', '-at_install')
 class DateRangeTypeTest(TransactionCase):
 
     def setUp(self):


lmignon:
@astirpe Do we need this fix? can't we just remove   `post_install = True` and `at_install = False` since tests are OK without your fix?

pedrobaeza:
@lmignon this is needed when you integrate this module with other modules that add required fields due to the way `required` implementation works (adding an SQL NOT NULL constraint at DB level), but not providing the default value yet because they are not loaded.

Any way, I think current way (through variables, not decorators), it's working without problems in v12. Decorators are the one that have changed. You can see in the log if the tests are executed just after the module loading or at the end.

astirpe:
@lmignon tests are OK without this fix because the repository is still empty, this is the only module present at the moment. The `post_install`/ `at_install` decorators have affect when other modules are installed.
According to https://github.com/odoo/odoo/commit/b356b190338e3ee032b9e3a7f670f76468965006, making use of test tags is the way it was intended to be.

lmignon:
@astirpe but is-it still required in Odoo 12. IMO this code should be deleted and reintroduced only when necessary. I am rather doubtful about the relevance of this fix which seems to me to mask a problem outside the module. I have never had any problems adding required fields that required this kind of fix.

astirpe:
@pedrobaeza I take the opportunity of this PR to ask something about the [Migration-to-version-12.0 guide](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0).

By doing the porting of my first modules to V12, I'm missing the following information in the guide:

- how to properly convert `post_install`/`at_install` (whether it's a nice to have or is a required step)
- add a reminder to check the readonly attribute for the related fields (following https://github.com/odoo/odoo/pull/27236/commits/1a69dab1d08434606e245301b917355524fa30e8)

May I ask you to keep the two points in consideration while updating the wiki? Thank you!

@lmignon I don't have any problem to remove that code. With this PR I just wanted to properly complete the porting of the module as it was already implemented.
If you all agree, I'll remove the code.

lmignon:
@astirpe Thank you very much for your work and the care you have taken in migrating the module with this PR to make it even cleaner. During the previous review I had not paid attention to this detail in the tests. The current PR gives me the opportunity to question the validity of these lines. For me they should be removed.

astirpe:
Here it is :smile:

lmignon (approved):
Thank you @astirpe

pedrobaeza:
Please don't remove these decorators/variables. I have already commented why they are there, and meanwhile here in this OCA repository they are not needed yet, we have integration tests that hits the problem, and they are totally harmless. Why do you want to remove them?

astirpe:
I'm personally open to any of the three options:

- closing this PR: the variables will be preserved
- keep the first commit only: the variables will be replaced with V12 decorators
- merge both the commits: the decorators/variables will be removed

Just let me know which option makes everyone agree.

lmignon:
@pedrobaeza I'm sorry but I honestly think it's not healthy to add code to OCA modules for problems outside OCA. I remain convinced that the case you describe must be solved in another way. We have in our projects cases similar to the one you describe and we have never had to use this kind of solution. 
I try to be careful what happens in this module but I can't find the PR that introduced this change. Do I miss something?

astirpe:
PRs are https://github.com/brain-tec/server-ux/pull/13 and this https://github.com/brain-tec/server-ux/pull/14

lmignon:
@astirpe No these are not the ones. These PR doesn't target OCA and as you can see there is no review. I've checked all the PRs (open and closed) and I can't the one adding this code. May be @pedrobaeza can point the right one.

pedrobaeza:
@lmignon it's not outside OCA. It's simply when you combine several modules, and it can happens inside or outside OCA.

Let me put an example:

- `purchase` module adds a required field `po_lead` in the company:

  https://github.com/odoo/odoo/blob/360212f2064e8bd72a3689f1e0104bfb73521f2a/addons/purchase/models/res_company.py#L9

- This module creates a company in the tests:

https://github.com/OCA/server-ux/pull/28/files#diff-cf005ed2488dad0cd047c8a33432db1bR20

- On Travis, there are 2 stages: one that install all modules and dependencies, and another that performs the tests with an update operation.

- When you have a module that requires `purchase` module in this repo (I know it's not the best example out there because you will say: purchase modules go to other repositories, but please think in a generic way) + `date_range`, on the first stage, all modules will be installed, so the CONSTRAINT NOT NULL on `po_lead` will be put on the DB.

- On the second stage, as `date_range` doesn't depend on `purchase` module, tests might be executed before, and then, when creating the company, you won't have the needed default value, making tests to fail.

This is the main reason why all tests should be by default this way IMO, as this doesn't have any drawback and prevents the problem. It's even what Odoo says about this problem. They promised to not add more required fields by this way (and put required on views for example) for not needing this workaround, but they still need to clean up remaining ones.

So please, don't be picky with this thing that is good for all without any bad counterpart.

yajo:
Just to add a little context on the subject, you can see for instance https://github.com/OCA/social/pull/258 and https://github.com/OCA/crm/pull/202 as real world examples where this happens.

I agree this is not the most beautiful solution, but it is the most practical for now indeed, and this is also true:



> is good for all without any bad counterpart

However, there's another thing that is also true: the new test labeling system gives us some flexibility that we didn't have before.

Maybe we could just tag the tests that are known to fail in integration environments with `integration-fails`, or tests that need network connectivity with `needs-network`, etc. This way, CIs that test integration environments in network-constrained environments can disable those tags and do not get false negatives, while not altering the execution context of tests in OCA itself. :thinking:

lmignon:
@pedrobaeza Where is the PR where this code has been added?
If we have no existing practical cases at the moment we must remove these lines. These lines could also lead to wrong results. Indeed when running tests at the end of installation, some others modules could have added constrains not supported by this module.

pedrobaeza:
@lmignon I have already spent the time explaining why they are needed. Now please explain me you which wrong results you get. I can assure you that's not true, and if it happens, the problem is in the test.

The PR was the module migration itself.

lmignon:
@pedrobaeza The PR introducing these changes is not the one for module migration since these lines are also into 11.0
This addition has never been discussed before.
 If the issue is not into OCA, there is no motivation to add these lines. An other way to avoid the problem you mention is to force the install of this module before running the tests.

lmignon:
@pedrobaeza regardless of the outcome on the discussion I must say I'm not too happy with commits being done without a PR in a module where I'm maintainer. The maintainer role documents says _PSC members should not merge PR concerning modules with declared maintainers, unless the maintainer is unreactive._

That being said, if the problem is the creation of companies in the tests, the problem can be solved by using existing companies in the demo data. I will change the code in that way.

astirpe:
Closing, as PR https://github.com/OCA/server-ux/pull/32 was merged.

PR #30, opened 2018-10-11
[FIX] mass_editing: remove non existent field in tests

cubells:
`notify_email` doesn't exist in res.partner model,

cc @Tecnativa


pedrobaeza (approved):
Fast-tracking this obvious one.

PR #31, opened 2018-10-12
12 mig base optional quick create mgs

GillesTephaneMeyomesse:
https://github.com/OCA/server-ux/issues/24


pedrobaeza:
Please squash together translation commits (and maybe some meaningless of the same author)

pedrobaeza:
We have been talking about deprecating this module in favor of web_m2x_options expanded with definition per model.

cc @sbidoul

i-vyshnevska (approved):
LG, possible solution  about  web_m2x_options could be mentioned in Readme and added to Roadmap

simahawk (changes requested):
LG but I'd squash a lot of useless commits like "descr" or similar together w/ the main ones.

Regarding tackling this on widget side: I agree.
I think this should be improved and instead of bumping an error in user's face it should block creation by setting `no_create` option on the views (`fields_view_get` or m2o widget override).

That being said we could track this in the roadmap and plan it for a next dev.

fmdl:
Note : I made this PR https://github.com/odoo/odoo/pull/32013 for Odoo to make this module for simpler.

alexis-via:
I proposed an improvement to this PR: https://github.com/acsone/server-ux/pull/1

alexis-via:
FYI, I will stop using base_optional_quick_create and use web_m2x_options instead. I updated my module eradicate_quick_create to depend on web_m2x_options instead of base_optional_quick_create https://github.com/akretion/odoo-usability/blob/12.0/eradicate_quick_create/__manifest__.py

SimoRubi (approved):
Tested locally and it works, thanks!
Could you please squash a bit the commits following https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests?

base_optional_quick_create/models/ir_model.py
```diff
@@ -0,0 +1,54 @@
...
+                    model_obj._revert_method(method_name)
+        return True
+
+    def _register_hook(self):
+        models = self.search([])
+        models._patch_quick_create()
+        return super()._register_hook()
+
+    @api.model
+    @api.returns('self', lambda value: value.id)
```
fmdl, line 44: should be remove

base_optional_quick_create/models/ir_model.py
```diff
@@ -0,0 +1,54 @@
...
+                if method and hasattr(method, 'origin'):
+                    model_obj._revert_method(method_name)
+        return True
+
+    def _register_hook(self):
+        models = self.search([])
+        models._patch_quick_create()
+        return super()._register_hook()
+
+    @api.model
```
fmdl, line 43: use 

```python
    @api.model_create_multi
    def create(self, vals_list):
         ir_models = super().create(vals_list)
         ir_models._patch_quick_create()
         return ir_models
```

ougc27:
Hello  @GillesTephaneMeyomesse this PR will be merged soon?

I can help you with the last comments, what do you think?

StefanRijnhart:
> I think this should be improved and instead of bumping an error in user's face

@simahawk this module does not throw an error in the user's face. Instead, it uses the mechanism in the Odoo web client that triggers the fallback to create_and_edit when whatever error occurs during quick create: https://github.com/odoo/odoo/blob/12.0/addons/web/static/src/js/fields/relational_fields.js#L391-L398

I don't see any progress on the integration of this per-module setting in web_m2x_options and I don't really see how that should work so I'm going to propose an improved version of this PR later today.

StefanRijnhart:
Continuing here: https://github.com/OCA/server-ux/pull/160. @fmdl I fixed your comments, thanks!

PR #36, opened 2018-10-16
[WIP] DO NOT MERGE: SQL constraint problem showcase

pedrobaeza:
* Add new module that depends on purchase so that it's installed and the SQL constraint is added
* Due to the module name, the module is going to be loaded later
* On date_range tests, without post_install, tests are failing on the company creation

### DIFF
--- a/date_range/tests/test_date_range.py
+++ b/date_range/tests/test_date_range.py
@@ -8,8 +8,6 @@ from odoo.tests.common import TransactionCase
 
 
 class DateRangeTest(TransactionCase):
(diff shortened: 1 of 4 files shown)


pedrobaeza:
@Yajo do you know why this is not failing now?

yajo:
As we discussed before, here's the culprit: https://github.com/odoo/odoo/pull/27883

It turns out nobody ever wondered why testing at install was the only way to test properly, and it turns out the patch is pretty simple... :roll_eyes:

PR #37, opened 2018-10-17
mass_editing: protect against fields that start with 'selection_'

CGenie:
A field can start with `selection_` string and still not be a selection :) Protect against this by checking if a dunder is in the string. This isn't perfect but works, unless you name your field `selection__...`.


legalsylvain:
hi @cgenie. Thanks for this patch. Make sense. I think that a field named 'selection__' should not occur. 

fast tracking trivial patch.

PR #29, opened 2018-10-08
[12.0][MIG] mass_editing: Migration to 12.0

aitorbouzas:
Migration to 12.0

PR #38, opened 2018-10-18
[MIG] mass_editing v12.0

hugho-ad:


pedrobaeza:
Please check #29 instead, that is already migrated.

hugho-ad:
Thanks, good to know!

pedrobaeza:
For avoiding wasted efforts, better to check first on the opened pull requests and the migration issue if there is already there or someone has expressed their interest of migrating it.

PR #39, opened 2018-10-23
[11][FIX] sequence_reset_period: incorrect python statement

tansadio:
Small correction (replaced super() -> super(IrSequence, self))


pedrobaeza:
This is not needed. You should deploy Odoo 11 with Python 3, not 2.

PR #40, opened 2018-10-23
 [ADD] email_show_only_changes

hugho-ad:


pedrobaeza:
Name should avoid plurals and follow guidelines. My suggestion is `mail_track_diff_only`.

pedrobaeza:
And this should be moved to OCA/social.

hugho-ad:
https://github.com/OCA/social/pull/311

PR #48, opened 2019-01-15
[11.0][IMP] base_tier_validation: 

LoisRForgeFlow:
* enhance views.
* register who does reviews in new field 'done_by'.
* Update README.


HviorForgeFlow (commented):
Nice feature @lreficent, does it make sense to add a request_by one??

LoisRForgeFlow:
@hveficent Yes, it also makes sense, added!

HviorForgeFlow (approved):
LGTM :+1:

base_tier_validation/views/tier_definition_view.xml
```diff
@@ -21,34 +24,59 @@
...
+        <field name="arch" type="xml">
+            <search>
+                <field name="model_id"/>
+                <field name="reviewer_id"/>
+                <field name="reviewer_group_id"/>
+                <field name="active"/>
+                <separator/>
+                <filter string="All" name="all" domain="['|', ('active', '=', False), ('active', '=', True)]" />
+                <group expand="0" string="Group By">
+                    <filter string="Model" domain="[]" context="{'group_by':'model_id'}"/>
```
naglis, line 67: I'd propose to add a `name` attribute to the group by `filter`, since without it it:
- It can't be selected with a xpath.
- It can't be used with `default_search_foo` on an action context.

[[source](https://github.com/odoo/odoo/commit/8f88c0336f881dd59d7c25b64c5bcba1989f998b#diff-150872ec366afb816e1fe98612a57634)]

And it's required starting with Odoo 12.

naglis (approved):

PR #50, opened 2019-01-28
[FIX] don't break when we have a nonexisting model

hbrunn:


base_optional_quick_create/models/ir_model.py
```diff
@@ -26,6 +26,8 @@ def wrapper(self, name):
         method_name = 'name_create'
         for model in self:
             model_obj = self.env.get(model.model)
+            if not model_obj:
```
Maartincm: I faced the same problem, and this approach did not work for me. Instead I suggest this:
```suggestion
            if not isinstance(model_obj, models.BaseModel):
```

Maartincm (changes requested):

base_optional_quick_create/models/ir_model.py
hbrunn: can you explain why you think this is necessary? For me the code fixes the problem as it is, and `get` is defined as returning something falsy if the requested key doesn't exist, so I'd be surprised if something else came back

Maartincm: Yes, get will return something falsy it the key does not exist but also if the key **does** exist because the result will be an empty recordset.

hbrunn: ah, thanks

NL66278 (approved):
:+1: LGTM

NL66278:
@Maartincm The code was updated with your suggestion. Would you refresh your review?

Maartincm (commented):
:ok_hand:

Maartincm (approved):
:ok_hand:

PR #51, opened 2019-02-07
[ADD] web_database_rollback: allow rollback database changes

mcojocaru:
there isn't any 8.0 branch, so it shouldn't be merged on 10.0


web_database_rollback/controllers/main.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class DBRollbackController(http.Controller):
+    @http.route(
+        '/web_database_rollback/activate',
+        type='json', auth='none')
+    def activate(self):
+        registry = openerp.modules.registry.RegistryManager.get(
+            request.session.db)
+        if registry.test_cr == None:
```
naglis: From [PEP8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations):
> Comparisons to singletons like `None` should always be done with `is` or `is not`, never the equality operators.

Same below in `rollback`.

web_database_rollback/controllers/main.py
```diff
@@ -0,0 +1,46 @@
...
+##############################################################################
+import openerp
+import openerp.http as http
+from openerp.http import request
+
+
+class DBRollbackController(http.Controller):
+    @http.route(
+        '/web_database_rollback/activate',
+        type='json', auth='none')
```
naglis: IIUC, this would allow anybody (even an unauthenticated user) to activate/deactivate the test mode. Maybe it would be safer to restrict it to authenticated users with a specific group?

web_database_rollback/static/src/js/db_rollback.js
```diff
@@ -0,0 +1,45 @@
...
+        template:'web_database_rollback.ButtonsWidget',
+
+        renderElement: function() {
+            var self = this;
+            this._super();
+            this.$el.show();
+            this.$el.find('.activate').on('click', function(ev) {
+                    self.$el.find('.activate').css("background-color", "green").css("color", "white");
+                    var func = '/web_database_rollback/activate';
+                    self.rpc(func, {}).done(function(res) {
```
naglis: Is it worth having `func` declared separated? If so, `func` is a bit misleading, maybe `endpoint` would be more precise? (same below)

web_database_rollback/README.rst
```diff
@@ -0,0 +1,60 @@
...
+=====
+
+On the right side of the systray there are two buttons: Activate and Rollback
+Press the Activate button (it will turn green), do any changes/actions in odoo (products, pickings, sale orders, etc) and save them.
+If you want to undo all the changes, press Rollback button.
+The database state will revert to the state prior to pressing Activate button.
+
+The number of Odoo workers has to be 0.
+
+Note that when you press Rollback button, all the chanages done by other users will be lost.
```
naglis: Typo: _changes_

web_database_rollback/controllers/main.py
```diff
@@ -0,0 +1,46 @@
...
+    def activate(self):
+        registry = openerp.modules.registry.RegistryManager.get(
+            request.session.db)
+        if registry.test_cr == None:
+            registry.enter_test_mode()
```
naglis: In the module README you mentioned it does not work in worker mode, maybe you could check that here and avoid activating in worker mode and maybe additionally return an warning to the user?

mcojocaru:
thanks for the feedback

mcojocaru:
travis checks failed because it expects to be be a 10.0 module, but this is a 8.0 module.
 Wrong Version Format "8.0.1.0.0" in manifest file. Regex to match: "(10\.0)\.\d+\.\d+\.\d+$"

PR #56, opened 2019-03-18
11.0 [ADD] hash_search

etobella:
This module implements the possibility of searching documents in Odoo
via QR code scanning.
A new QR icon will appear in the top navigation tab and by clicking on it the
scanning window will be displayed. Scan a valid QR and you will be
automatically redirected to that document.
Also allows to print the QR from the expected records through a ZPL2 label.

There is still a lot of work to do, but I think that the schema is strong enough in order to present it to OCA.

- [x] Update readme
- [x] Improve printing function
- [ ] Add some tests (probably a module `test_hash_search` will be necessary)

This module allows only to search from the Odoo interface, but I will present a module that allows to scan automatically a document with a hash qr and attach it to his record.

You can see al the work related on https://github.com/tegin/hash-search/tree/11.0-add-hash_search


etobella:
@jbeficent

MiquelRForgeFlow:
Why should the test be in another separated module?

etobella:
Because it is a mixin, no real record is generated.

etobella:
@antoniocanovas ping

etobella:
@amcor ping

JordiBForgeFlow (approved):

hash_search/models/hash_search.py
```diff
@@ -0,0 +1,81 @@
...
+            })
+            result['context'] = json.dumps(context)
+            return result
+        res = self.env[hash.model].browse(hash.res_id)
+        res.check_access_rights('read')
+
+        result = {
+            "type": "ir.actions.act_window",
+            "res_model": hash.model,
+            "views": [[res.get_formview_id(), "form"]],
```
etobella, line 77: @pedrobaeza as we talked today, this funtion returns the default form view for record and model. For example, on an invoice it will return the expected view accordibg to the result

etobella:
This module is being replaced by #72 so I will close the PR.

PR #63, opened 2019-05-08
[IMP] base_tier_validation: Add a chatter message if the model has it

etobella:
This PR adds a Chatter message if the associated record has it.

@jbeficent @lreficent @ageficent

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -154,7 +154,14 @@ class TierValidation(models.AbstractModel):
             'done_by': self.env.user.id,
             'reviewed_date': fields.Datetime.now(),
         })
-        # TODO: add message_post
+        for review in user_reviews:
+            rec = self.env[review.model].browse(review.res_id)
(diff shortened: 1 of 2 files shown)


base_tier_validation/models/tier_validation.py
```diff
@@ -154,7 +154,14 @@ def _validate_tier(self, tiers=False):
             'done_by': self.env.user.id,
             'reviewed_date': fields.Datetime.now(),
         })
-        # TODO: add message_post
+        for review in user_reviews:
+            rec = self.env[review.model].browse(review.res_id)
```
LoisRForgeFlow, line 158: Could you put this in a hook method, something like `_notify_tier_change` or similar. This way it could be disabled/changed easily in custom modules or for specific models even if they have chatter.

etobella: Done!

JordiBForgeFlow (approved):

AdriaGForgeFlow (approved):
Code review :+1:

PR #66, opened 2019-05-16
[11.0][IMP] base_tier_validation

AdriaGForgeFlow:
This PR intends to make some improvements on base_tier_validation module:

* Manage the alerts from javascript removing the need of all the code that needed to be put in the form view to show the alert banners (see README to see how it's done now)
* Moves the Reviews button to the alert banner being accessible from there all the reviews' information:
image
image
image

Thoughts:
* When state of process is set to done maybe remove alert banner and put a button in the button box that will redirect the user to a new view showing all reviews.


etobella:
Just a question, will it work with the previous forms? I mean, if this PR is approved, the other modules will keep working before you make PRs for them?

AdriaGForgeFlow:
It won’t. That’s why I created a draft PR, there’s some further work to do. But still I wanted to share this approach to see what people think.

JordiBForgeFlow (approved):
:+1: Works as expected.

etobella:
This is still Draft?

AdriaGForgeFlow:
> This is still Draft?

Yes, we thought that the best is to introduce new changes directly into version 12. We need to thing where is the best spot to represent the reviews. 
Not sure what to do with this PR.

etobella:
@jbeficent What do you think?

AdriaGForgeFlow:
Superseeded by #76

PR #72, opened 2019-06-04
[11.0][ADD] document quick access

etobella:
This module replaces #56 
Now, the name is simpler and it does not generate an extra table with the data.

@pedrobaeza @jbeficent


etobella:
The travis will go red due to base_optional_quick_create. It creates a test model that gives problems to this modules. In order to fix it, we can set `@post_install` and `@at_install`.
I made a PR with the change #73

document_quick_access/README.rst
```diff
@@ -0,0 +1,89 @@
...
+=============
+
+# Access on developer mode
+# Access `Settings > Technical > Document Quick Access`
+# Create a record selecting the model, format and priority
+
+Usage
+=====
+
+# Click on the QR button on the menu
```
pedrobaeza: I'm wondering if this can't be added as a general listener in whole web client for not needing to click in any place. Would it be an improvement? If not, I suppose this button is very accessible from any place. Isn't it?

etobella: It could be interesting, but this could blocked by the standard barcode reader of odoo. 
Maybe, we could add a shortcut? Right now, there is a button on the menu accessible everywhere.

pedrobaeza: OK, put it on the roadmap the other option and specify here in the README where the button is in the interface (`upper right in the navbar` for example).

etobella: OK! Added the roadmap and regenerated the readme.
I have been reviewing this, and the shortcut is easier to add if we use `web_responsive`. Harder with the standard odoo.

JordiBForgeFlow (approved):
:+1: Functional and code review. Works great! 

Small comment. In the readme I suggest to explain that by default the barcode should be encoded as "model,id"

etobella:
model,id is the standard behavior, but this module is ready to accept other formats. We could make an explanation, but that's not the only solution.
The best way to show it is to install printer_zpl2, generate a label for the model with a QR with code `object.get_quick_access_code()`. In order to show, you could use labellary mode.

document_quick_access/README.rst
etobella: I added a shortcut Alt + Shift + Q

jarroyomorales (approved):
LGTM :+1:

etobella:
As we are talking about QRs on this PR, it is interesting to use OCA/reporting-engine#288, not the standard QR of odoo. Quality is better with this PR. Odoo standard QR has an issue on it that makes it hard to use.

document_quick_access/README.rst
```diff
@@ -0,0 +1,95 @@
...
+
+.. contents::
+   :local:
+
+Configuration
+=============
+
+# Access on developer mode
+# Access `Settings > Technical > Document Quick Access`
+# Create a record selecting the model, format and priority
```
pedrobaeza, line 42: I think this text is outdated

pedrobaeza: Sorry, I have now seen that this is for setting rules.

PR #73, opened 2019-06-04
[FIX] base_optional_quick_create: Add tests as post_install

etobella:
This PR changes the behavior of the tests (when are executed, but not the results) in order to make #72 work.
I did not dump the version, as it only affects tests


jarroyomorales (approved):

etobella:
@jbeficent

JordiBForgeFlow (approved):

PR #74, opened 2019-06-05
[11.0][ADD] document quick access folder auto classification

etobella:
This module creates a job that scans all files from a folder and attaches them
to its record. The record is found using the document quick access rules.

Example of use for Bank Mandates:

* Create a document quick access rule for bank mandates and show the QR on the document.
* Print the mandate
* The customer signs the mandate
* You scan the mandate, the scanner will send it to the file and odoo will read it. Once he founds the QR, he will attach it directly to the mandate record.

A new module should be necessary to make this work with queue and ssh folders (that should be the expected configuration IMO)

This PR depends on #72 
@jbeficent @anajuaristi @antoniocanovas @amcor


etobella:
A dependency to queue has been added. It will be used automatically if exists the parameter `document_quick_access_auto_classification.process_path`.

Also, with this parameter, a single commit will be used for each case.
The queue job cannot be tested with travis, as it is committing.

document_quick_access_folder_auto_classification/data/cron_data.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+  © 2013-2016 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
```
jarroyomorales: Is this right?

etobella: Fixed

jarroyomorales (approved):

etobella:
I made a refactoring of the module in order to simplify some code and make it easier to include document from other sources.

LoisRForgeFlow (approved):
Works like a charm! :clap:

etobella:
Travis is :red_circle: but I cannot reproduce it locally, even doing the same than travis :cry:

LoisRForgeFlow:
I cannot reproduce locally either, but what I can say is that the error is totally unrelated to this PR: https://github.com/OCA/server-ux/issues/104.

document_quick_access_folder_auto_classification/views/document_quick_access_missing.xml
```diff
@@ -0,0 +1,77 @@
...
+        </field>
+    </record>
+
+    <record model="ir.ui.view" id="document_quick_access_missing_tree_view">
+        <field name="name">document.quick.access.missing.tree (in document_quick_access_folder_auto_classification)</field>
+        <field name="model">document.quick.access.missing</field>
+        <field name="arch" type="xml">
+            <tree create="false" delete="false">
+                <field name="name"/>
+                <field name="state"/>
```
LoisRForgeFlow, line 59: I would show the date as well in the document missing tree view. What do you think?

```suggestion
                <field name="state"/>
                <field name="create_date"/>
```

document_quick_access_folder_auto_classification/views/document_quick_access_missing.xml
```diff
@@ -0,0 +1,77 @@
...
+                    <field name="state" widget="statusbar"/>
+                </header>
+                <sheet>
+                    <h1>
+                        <field name="name"/>
+                    </h1>
+                    <group>
+                        <field name="data" filename="name"/>
+                        <field name="res_id" invisible="1"/>
+                        <field name="model" invisible="1"/>
```
LoisRForgeFlow, line 31: and in the form view.
```suggestion
                        <field name="model" invisible="1"/>
                        <field name="create_date"/>
```

LoisRForgeFlow (commented):
Little usability suggestion inline:

etobella:
@lreficent Comments attended

etobella:
The problem with travis is related to base_import_security_group. Test fails when mail is installed as you can check on:
https://travis-ci.org/etobella/server-ux/jobs/617093804
So, only two options for me, split travis or check why the module fails with mail, but I am unable to reproduce the problem locally :cry: 

Any idea?

JordiBForgeFlow:
@StefanRijnhart perhaps you know. The module 'base_import_security_group' uses phantomjs in a test. https://github.com/OCA/server-ux/blob/11.0/base_import_security_group/tests/test_base_import_security_group.py#L31

The problem is that when the phantomJS is run using as login the demo user: https://github.com/OCA/server-ux/blob/11.0/base_import_security_group/tests/test_base_import_security_group.py#L63

For some reason it must be loading the login page, because it will attempt to fetch the field 'signup_url' from module server_auth, and fails here: https://github.com/odoo/odoo/blob/11.0/addons/auth_signup/models/res_partner.py#L49

I tried adding explicit write permissions to res.users to the demo user in the test, but still fails.https://github.com/etobella/server-ux/pull/2/commits/2a2ac9bcacd81e1694eff6d5f1c781e2e81810e2

StefanRijnhart:
I checked the problem and something causes the prefetch to be filled with more than the partner of the current user. When the signup_url is read for one of them, prefetch will try to read them for all partners in the prefetch registry, and of course our demo user is not allowed to read it of those other partners. So I think base_import_security_group will have to be tested in its own run.

etobella:
Well, now it is green :green_apple: :green_apple:

LoisRForgeFlow:
It's time to merge it! before travis changes its mind! :laughing: :laughing:

PR #79, opened 2019-07-01
[ADD] base: security audit

fmdl:
Dear Community

It is hard to check if the rules and ACL are correctly configured ( for exemple : check if public user can download all invoice).
I have work on this PR https://github.com/odoo/odoo/pull/25587 to know ACL/Rules per user, but if there are lot of ACL/Rules it is hard to check.
I have also work on this PR https://github.com/odoo/odoo/pull/29122, but you need to check some time if it is good.

That why I have work on an audit system where you check that specific user/groups have not rights to read/write/unlink a specific records.
A cron check every month if everything is ok, if not an activity is created.

This module prevent to expose private data on internet, or by a malicious internal user.

image

Tell me what you think, 

Regards

@yenthe666
@alexis-via 
@sla-subteno-it
@pedrobaeza 
@mreficent
cc @odony


pedrobaeza:
Why here on server-ux? This seems something more for server-tools, the same as auditlog, but it's debatable.

fmdl:
@pedrobaeza you are right, I will move on server-tool.

fmdl:
move here https://github.com/OCA/server-tools/pull/1617

PR #81, opened 2019-07-12
[12.0][MIG] subscription from the official addons of Odoo Community v10 to OCA

alexis-via:


sbidoul:
Hi @alexis-via 

Could you do an initial commit that takes the Odoo SA code as is, and then another commit with your changes, to avoid any doubt about copyright attribution

Perhaps this module should be renamed to `recurring_document` or something, to avoid confusion with the OE subscription modules?

pedrobaeza:
And why in server-ux?

alexis-via:
@pedrobaeza feel free to suggest a better project and I'll create another PR.

pedrobaeza:
on OCA/contract?

alexis-via:
OK

PR #90, opened 2019-09-18
10.0 mig date range

daramousk:


astirpe:
It's already existing for V10, here: https://github.com/OCA/server-tools/tree/10.0

lmignon:
@daramousk Thank you for this PR. Nevertheless this addon already exists in odoo 10 but in an other repo https://github.com/OCA/server-tools/tree/10.0/date_range

daramousk:
Ah thank you, totally missed it.

lmignon:
you're welcome @daramousk :smirk:

PR #94, opened 2019-09-30
[13.0] [MIG+REF] mass_editing

legalsylvain:
**mass_editing improvment** : 
- [x] renaming ``mass.object`` to ``mass.editing`` for consistency
- [x] make options (set / add / remove) translatable
- [x] simplify ``mass.editing.wizard`` (V12: 300 lines ; V13: 160 lines)
- [x] make easier and add modularity in the ``mass.editing.wizard`` python code.
- [x] replace field_ids M2M fields by a new model``mass.editing.line`` to have the possibility to order fields in the wizard, by a sequence and to select the widget for each field.

image

- [x] migration Script.

**Add abstraction** (``mass.operation.abstract``)
- [x] Introducing a new abstract module ``mass_operation_abstract`` to provide mixin objects that allow to make mass operation on many items of any models 
- [x] make ``mass_editing`` depending on ``mass_operation_abstract``

PR #97, opened 2019-10-03
[13.0][ADD] mass action

legalsylvain:
I propose a new module ``mass_action``. It allows you to have a new button to execute python code on any items of any model.
It can be useful in many cases, when the users has to make a recurring non covered operations. (confirm and deliver and invoice many quotations, ...)

It works similarly as the module ``mass_editing`` (same repository) and depends on the new module 
``mass_operation_abstract``.

image

CC : @quentinDupont, @JayVora-SerpentCS 

~~note : requires #94 to work.~~


JayVora-SerpentCS (approved):
surely innovative!

legalsylvain:
Thanks @JayVora-SerpentCS for your review ! 

Note : Travis / runbot are red only because #94 is not merged.

mass_action/models/mass_action_wizard.py
```diff
@@ -0,0 +1,20 @@
...
+class MassActionWizard(models.TransientModel):
+    _name = 'mass.action.wizard'
+    _description = "Mass Actions Wizard"
+    _inherit = 'mass.operation.wizard.mixin'
+
+    def _apply_operation(self, items):
+        self.ensure_one()
+        mass_action = self._get_mass_operation()
+        localdict = {'self': items}
+        safe_eval(mass_action.python_code, localdict, mode="exec", nocopy=True)
```
hbrunn, line 19: I started writing here to ask to do the same as [server actions](https://github.com/OCA/OCB/blob/13.0/odoo/addons/base/models/ir_actions.py#L431), but then I ask myself: Why isn't this module a frontend for creating server actions, possibly with their own type?

legalsylvain: Hi @hbrunn. Thanks for your review !

About ``ir.actions.server`` that does the same thing as my module, the honest answer is : I just discovered servers actions I didn't know. ;-) So, I took a time to discover this "not so new model" ... and : 

1) Finally, the improvment this module makes is adding features present in the module ``mass_operation_abstract`` : 
-> possibility to add domain, to avoid to execute server actions on non desired object. (for exemple, if the state is not valid) and warn the user.
-> possibility to add a group, so the menu will not be displayed for user that has not the right to do the action, simplifying UI.

> Why isn't this module a frontend for creating server actions ?

Do you mean, 
-> keep this module, 
-> add a ir_actions_server_id on the model ``mass_action``
-> so reuse all the safe eval algorithm, & co, and remove the safe_eval part in mass_action module ?

> possibly with their own type

Not sure to understand that point. I guess, add a selection value on ``usage``[ field](https://github.com/OCA/OCB/blob/13.0/odoo/addons/base/models/ir_actions.py#L350), named ('mass_operation', 'Mass Opération) ?

Thanks.

hbrunn: Yes, keeping the module but using server actions internally, that's what I propose.

And the own type, what I meant was adding to the state selection, but that's nonsense I guess. We can just use normal server actions, just like cron does nowadays too: https://github.com/OCA/OCB/blob/13.0/odoo/addons/base/models/ir_cron.py#L52

legalsylvain: Hi @hbrunn. I tried to integrated an ``ir.actions.server`` but I failed. I'm not really mastering  ``delegate=True`` and it generates errors (field ``name`` required) and not very proud of the final design !

If I check the interest of this module : 

> -> possibility to add domain, to avoid to execute server actions on non desired object. (for exemple, if the state is not valid) and warn the user.

It is very easy to add it on an ir.actions.server with a ``filtered`` command. 

> -> possibility to add a group, so the menu will not be displayed for user that has not the right to do the action, simplifying UI.

It seems that a group is also implemented on the ``ir.actions.server``.

so the only improvment I see is that the UI for ``mass.action`` is a little better and simple that for ``ir.actions.server`` but it doesn't deserve a new module.

What do you think ? 

propose to close. 
kind regards.

hbrunn: maybe better, sorry about your work then.

legalsylvain: No problem. That's also why I share GRAP work to OCA : To hear people saying me "_Hey guy ! You reinvented the wheel !_" 

thanks for your review.

hbrunn: *Hey guy ! Thanks anyways !* ;-)

PR #95, opened 2019-09-30
[13.0][MIG] date_range

MiquelRForgeFlow:
Standard migration.

PR #99, opened 2019-10-18
[13.0] [MIG] date_range

lmignon:
fixes and supersedes #95


pedrobaeza:
Please check runbot warning:
```
2019-10-18 10:25:50,201 160 WARNING openerp_test odoo.addons.base.models.ir_module: python external dependency dateutil should be replaced by it's PyPI package name
```

lmignon:
@pedrobaeza If I use de Pypi package name, pylint fails
```
Missing external dependency "dateutil.relativedelta" from manifest. More info: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#external-dependencies
```
:unamused:

a pylint issue I suppose?

pedrobaeza:
Well, I'm only saying about the thing that makes runbot CI red. But isn't that an Odoo requirement? Do we need to declare explicitly in this module? I don't think so.

lmignon:
@pedrobaeza Yes it's an odoo requirement. I don't know why we must now declare this dependency compared to 12.0 but I'm not choked to have to do it since it's always better to not rely on transitively acquired dependencies when explicitly using one of them. I don't know how to solve the issue in this specfic case...

pedrobaeza:
@sbidoul @guewen you were dealing with this specific case, isn't it?

astirpe (approved):
In the meantime: Tested OK, code review: OK

sbidoul:
See https://github.com/OCA/pylint-odoo/pull/263 

It needs to evolve to avoid conflict with the local isort config, and handle the new python dependency mechanism of Odoo 13 (referencing PyPI distribution names instead of import name).

In the meantime it does no harm to add the dependency explicitly.

pedrobaeza:
> In the meantime it does no harm to add the dependency explicitly.

Except for the runbot warning...

sbidoul:
@lmignon The manifest must say `python-dateutil`, not `dateutil`. https://pypi.org/project/python-dateutil/. That should silence runbot.

lmignon:
@sbidoul but in this case, pylint fails...

sbidoul:
Ah yes, sorry. Then the pylint check must be temporarily disabled in `.pylintrc-mandatory`. In any case when pylint will be updated and released we'll need to push at minimum a new `.pre-commit-config.yaml`.

lmignon:
@sbidoul OK I'll do like proposed

MiquelRForgeFlow (approved):

PR #101, opened 2019-11-05
13.0 [MIG] base_tier_validation

spoerp:
Just edited one test which checked if there are any tier_validation records all ready created. Which failed when we installed purchase_tier_validation and then odoo.sh runs all the tests again.

PR #103, opened 2019-11-14
13.0 mig base tier validation formula

pegonzalezspesol:
Installable but work in progress. Also depends on #101


pegonzalezspesol:
@angelmoya please review

base_tier_validation_formula/models/tier_definition.py
```diff
@@ -0,0 +1,35 @@
...
+        "The result of executing the expression must be a res.users "
+        "recordset.",
+        default="# Available locals:\n#  - rec: current record\n"
+        "#  - Expects a recordset of res.users",
+    )
+    review_type = fields.Selection(selection_add=[("expression", "Python Expression")])
+
+    @api.onchange("review_type")
+    def onchange_review_type(self):
+        super(TierDefinition, self).onchange_review_type()
```
angelmoya, line 30: Why reviewer_expresion is updated on review_type change?

reviewer_expression has yet default value.

Maybe would be better to don't define this onchange, or check review_type to set this value or false if review_type is not formula.

What do you think?

angelmoya (approved):
LGTM :+1:

base_tier_validation_formula/models/tier_definition.py
pegonzalezspesol: Sure.. at the end of the day this info is not visible in case review_type is not expression. Take a look now.

pegonzalezspesol:
@lreficent please review

base_tier_validation_formula/models/tier_review.py
```diff
@@ -10,8 +10,11 @@ class TierReview(models.Model):
     _inherit = "tier.review"
 
     python_reviewer_ids = fields.Many2many(
+        "res.users",
+        "tier_review_python_reviewer_rel",
+        "tier_review_id",
+        "user_id",
```
LoisRForgeFlow: can you be explicit with the attributes?

base_tier_validation_formula/models/tier_definition.py
```diff
@@ -29,7 +29,11 @@ class TierDefinition(models.Model):
...
         super(TierDefinition, self).onchange_review_type()
         self.reviewer_expression = (
-            "# Available locals:\n"
-            "#  - rec: current record\n"
-            "#  - Expects a recordset of res.users"
+            (
+                "# Available locals:\n"
+                "#  - rec: current record\n"
+                "#  - Expects a recordset of res.users"
+            )
```
LoisRForgeFlow, line 37: why the double parenthesis? you could just do `"some text" if condition == True else False`

LoisRForgeFlow (commented):
Code review

base_tier_validation_formula/models/tier_definition.py
pegonzalezspesol: done

base_tier_validation_formula/models/tier_review.py
pegonzalezspesol: done

LoisRForgeFlow (approved):
Thanks! LGTM

PR #107, opened 2019-11-26
[12.0][ADD] base_tier_validation_helper

ps-tubtim:
This module helps to create, edit and delete many tier definitions.


LoisRForgeFlow:
Interesting idea. Some question on the approach:
* why don't you implement it as a wizard?
* can't it be done without the dependency to tier_validation_formula?

kittiu:
@lreficent 

Currently in base_tier_validation, there is no header model that can group tiers (which I do agree on the base simplicity). So, we like to propose this as separated set of addons.

1. To do header model for ease of maintenance
2. To make it simple to do different kind of complex use cases (1 addon per new use case, amount level is just one of them)
3. For each model, we could also have different set of tier validation, based on different department, etc.

> why don't you implement it as a wizard?

As per mentioned above. Do you think it is reasonable?

> can't it be done without the dependency to tier_validation_formula?

@ps-tubtim I do agree with @lreficent I think we can make it very generic like this,

- base_tier_validation_helper --> nothing more than the header model link to the needed model, action buttons (with no implementation yet). No need to inherit formula.
- base_tier_validation_helper_amount_level --> this is the implement for the use case that you are working on (validation by amount level). This will inherit formula.

In the future, we can have more use case on this helper, which may or may not need formula, i.e.,
- base_tier_validation_helper_xxx

What do you think?
Also, please feel free to comment on naming convention, we don't have a good name yet.

kittiu:
@ps-tubtim something you may consider,

* As we split, and base_tier_validation_helper_amount_level is just one implementation. We have helper_type, as selection field.
* To ease user understanding, we can add detailed explaination and examples, in separated "Help" tab. Help in mis_builder is good example, https://github.com/OCA/mis-builder/blob/10.0/mis_builder/views/mis_report.xml#L111

kittiu:
@lreficent if you don't mind, may be @ps-tubtim can continue. Should you have more input, do let us know anytime.

LoisRForgeFlow:
@kittiu Thanks for the detailed response and sorry for my late answer.

> As per mentioned above. Do you think it is reasonable?

Yes, I agree that it can be good to keep that record to ease management and maintenance of models with a lot of tiers defined.

> base_tier_validation_helper --> nothing more than the header model link to the needed model, action buttons (with no implementation yet). No need to inherit formula.

I like this idea, as formula extension should be avoided as much as possible to reduce security risks (quite remote risk though, but we have to be aware). In this module we may not provide any logic, just as you say, a header to easily manipulate all the tier definitions for a model. So, you should be able to edit the domain in that view per each line.

> base_tier_validation_helper_amount_level --> this is the implement for the use case that you are working on (validation by amount level). This will inherit formula.

Once formula is inhereted I would not split modules more. New implementations and logics to manage the tiers could just extend this second module.

> Also, please feel free to comment on naming convention, we don't have a good name yet.

I don't have any problem with the name!

In summary my only suggestion: base_tier_validation_helper and base_tier_validation_helper_formula would be enough segregation. Each one of them can contain different logic to handle the tiers, logic with no need to use formula goes to base_tier_validation_helper and logic using formula goes to base_tier_validation_helper_formula

What do you think?

Thanks for your contributions and ideas :wink:

kittiu:
@lreficent thanks for your reply. I do agree with most, just to add some thing further,

What I agree that, only 2 module is enough,
1. base_tier_validation_helper --> which is very basic.
2. base_tier_validation_helper_formula --> the implementation goes here.

I want to add that `base_tier_validation_helper_formula` could be the end module for formula use. Today we only have 1 use case. In the future if we have more, we can just enhance it here as another type.

This is to avoid too many modules which can be confusing. Having said that, for now, your advise is still valid. :)

@ps-tubtim agreed? anything you want to add?

LoisRForgeFlow:
@kittiu Yes, I agree. Feel free to proceed with the changes!

base_tier_validation_helper/models/tier_definition_helper.py
```diff
@@ -0,0 +1,230 @@
...
+    _name = 'tier.definition.helper.line'
+    _description = 'Tier Definition Helper Line'
+
+    parent_id = fields.Many2one(
+        comodel_name='tier.definition.helper',
+        index=True,
+        required=True,
+        ondelete='cascade'
+    )
+    approval_level = fields.Integer(
```
HviorForgeFlow, line 148: I couldn't figure out how this variable worked until I see the code, please add a help explaining that levels go from lowest one 1 to 100, I don't see any restriction to not use a number higher to 100.

base_tier_validation_helper/models/tier_definition_helper.py
```diff
@@ -0,0 +1,230 @@
...
+        string="Reviewer",
+    )
+    reviewer_group_id = fields.Many2one(
+        comodel_name="res.groups",
+        string="Reviewer group",
+    )
+    definition_domain = fields.Char(
+        string='Domain',
+        help='Expressions can be any valid python expressions.'
+             '(eg ["|", ["id", "=", 1], ["id", "=", "2"]])',
```
HviorForgeFlow:
```suggestion
             '(eg ["|", ("id", "=", 1), ("id", "=", 2)])',
```
Should it  be this way?

base_tier_validation_helper/models/tier_definition_helper.py
```diff
@@ -0,0 +1,230 @@
...
+
+class TierDefinitionHelper(models.Model):
+    _name = 'tier.definition.helper'
+    _description = 'Tier Definition Helper'
+
+    @api.model
+    def _get_default_name(self):
+        return _('New')
+
+    name = fields.Char(
```
kittiu, line 12: I think we should not allow name duplication for the helper (it will make no sense to have same name).
- Please add sql unique constraints
- When copy, should have (Copy) suffix

base_tier_validation_helper/models/tier_definition_helper.py
```diff
@@ -0,0 +1,230 @@
...
+        string='Description',
+        required=True,
+        default=lambda self: self._get_default_name(),
```
kittiu:
```suggestion
        default='New',
```
I just think it will be less code for almost the same thing.

base_tier_validation_helper/models/tier_definition_helper.py
```diff
@@ -0,0 +1,230 @@
...
+        inverse_name='parent_id',
+        string='Lines',
+        readonly=True,
+        states={'draft': [('readonly', False)]},
+    )
+
+    @api.multi
+    @api.constrains('line_ids')
+    def _check_line_ids(self):
+        if self.state == 'done' and \
```
kittiu: As you use @api.multi, we need for loop on self.

kittiu: But why do we need this constraint? line_ids already readonly on state=='done' isn't it?

base_tier_validation_helper/models/tier_definition_helper.py
```diff
@@ -0,0 +1,230 @@
...
+        models = self.env['tier.definition']._get_tier_validation_model_names()
+        return {'domain': {'model_id': [('model', 'in', models)]}}
+
+    @api.multi
+    def button_draft(self):
+        self.write({'state': 'draft'})
+
+    @api.multi
+    def button_create_tier(self):
+        if not self.line_ids:
```
kittiu: We use @api.multi, right? Please loop.
Otherwise, use self.mapped('line_ids')

base_tier_validation_helper/models/tier_definition_helper.py
```diff
@@ -0,0 +1,230 @@
...
+            raise ValidationError(_('Please add at least one line.'))
+        tier_definition_env = self.env['tier.definition']
```
kittiu: Naming convension, use Camel case for class.
```suggestion
        TierDefinition = self.env['tier.definition']
```

base_tier_validation_helper/models/tier_definition_helper.py
```diff
@@ -0,0 +1,230 @@
...
+        for rec in self.line_ids:
```
kittiu: Self can be multi, please use for loop.

base_tier_validation_helper/models/tier_definition_helper.py
```diff
@@ -0,0 +1,230 @@
...
+            if not rec.tier_definition_id:
+                tier_definition = tier_definition_env.create(
+                    rec._prepare_tier_definition())
+                rec.tier_definition_id = tier_definition
```
kittiu: Can be done in one line to shorten, i.e.,

```
res.tier_definition_id = TierDefinition.create(rec._prepare_tier_definition())
```

base_tier_validation_helper/models/tier_definition_helper.py
```diff
@@ -0,0 +1,230 @@
...
+        self.write({'state': 'done'})
+
+    @api.multi
+    def button_delete_tier(self):
+        self.line_ids.mapped('tier_definition_id').unlink()
```
kittiu:
```suggestion
        self.mapped('line_ids').mapped('tier_definition_id').unlink()
```

base_tier_validation_helper/models/tier_definition_helper.py
```diff
@@ -0,0 +1,230 @@
...
+        self.write({'state': 'cancel'})
+
+
+class TierDefinitionHelperLine(models.Model):
+    _name = 'tier.definition.helper.line'
+    _description = 'Tier Definition Helper Line'
+
+    parent_id = fields.Many2one(
```
kittiu: Naming Convention
(parent_id normally means same object but have hierarchy)
```suggestion
    helper_id = fields.Many2one(
```

base_tier_validation_helper/models/tier_definition_helper.py
```diff
@@ -0,0 +1,230 @@
...
+
+    @api.multi
+    def button_delete_tier(self):
+        self.line_ids.mapped('tier_definition_id').unlink()
+        self.write({'state': 'cancel'})
+
+
+class TierDefinitionHelperLine(models.Model):
+    _name = 'tier.definition.helper.line'
+    _description = 'Tier Definition Helper Line'
```
kittiu, line 135: Should it _order by approval_level?

base_tier_validation_helper/models/tier_definition_helper.py
```diff
@@ -0,0 +1,230 @@
...
+        index=True,
+        required=True,
+        ondelete='cascade'
+    )
+    approval_level = fields.Integer(
+        string='Level',
+        default=1,
+        required=True,
+    )
+    name = fields.Char(
```
kittiu, line 153: name should be requried.

base_tier_validation_helper/models/tier_definition_helper.py
```diff
@@ -0,0 +1,230 @@
...
+        string='Name',
+        default='New Tier Definition',
+    )
+    reviewer_id = fields.Many2one(
+        comodel_name="res.users",
```
kittiu: Double quote or single quote, please choose one.

base_tier_validation_helper/models/tier_definition_helper.py
```diff
@@ -0,0 +1,230 @@
...
+        string="Reviewer",
+    )
+    reviewer_group_id = fields.Many2one(
+        comodel_name="res.groups",
+        string="Reviewer group",
+    )
+    definition_domain = fields.Char(
```
kittiu: I think domain should be in header, not in line.

base_tier_validation_helper/models/tier_definition_helper.py
```diff
@@ -0,0 +1,230 @@
...
+            'review_type': self.parent_id.review_type,
+            'definition_type': self.parent_id.definition_type,
+            'notify_on_create': self.parent_id.notify_on_create,
+            'has_comment': self.parent_id.has_comment,
+            'approve_sequence': self.parent_id.approve_sequence,
+        }
+        # add name
+        if self.parent_id.auto_name:
+            tier_definition['name'] = self.parent_id.model_id.name +\
+                ' (' + str(self.approval_level) + ')'
```
kittiu: Please always do not use '+' for sting.
```
name = '%s (%s)' % (var1, var2)
```

kittiu: Anyway, I think auto_name is not needed (let use make name himself is ok).

base_tier_validation_helper/models/tier_definition_helper.py
```diff
@@ -0,0 +1,230 @@
...
+            'model_id': self.parent_id.model_id.id,
+            'sequence': 100 - self.approval_level,
+            'review_type': self.parent_id.review_type,
+            'definition_type': self.parent_id.definition_type,
+            'notify_on_create': self.parent_id.notify_on_create,
+            'has_comment': self.parent_id.has_comment,
+            'approve_sequence': self.parent_id.approve_sequence,
+        }
+        # add name
+        if self.parent_id.auto_name:
```
kittiu: multi or ensure_one() is needed.

base_tier_validation_helper/models/tier_definition_helper.py
```diff
@@ -0,0 +1,230 @@
...
+        if self.parent_id.definition_type == 'domain':
+            tier_definition['definition_domain'] = self.definition_domain
+        return tier_definition
+
+    @api.onchange('approval_level')
+    def _onchange_approval_level(self):
+        if self.tier_definition_id:
+            tier_definition = self.env['tier.definition'].search(
+                [('id', '=', self.tier_definition_id.id)], limit=1)
+            tier_definition.write({'sequence': 100 - self.approval_level})
```
kittiu: What number 100 is from? What if approval_level is > 100?
Or can we find max(approval_level) and make the number +1, etc...

base_tier_validation_helper/models/tier_definition_helper.py
```diff
@@ -0,0 +1,230 @@
...
+
+    @api.onchange('name')
+    def _onchange_name(self):
+        if self.tier_definition_id:
```
kittiu: What is this for? Do we still need it?

base_tier_validation_helper/models/tier_definition_helper.py
kittiu: We can use in view, create="0" and reduce state to only Draft -> Done.

base_tier_validation_helper/models/tier_definition_helper.py
kittiu: Let's use max(...) to find the right number?

base_tier_validation_helper/models/tier_definition_helper.py
kittiu: I think many of the onchange can be removed, if we only allowed 2 states, Draft -> Done. When set to draft, all lines will be deleted and restart all over again.
No need to edit line when it is active.

kittiu:
@ps-tubtim 
And because the intention of this module is to ease user (otherwise, user can use standard one). I think the "Domain" should be defined at the header level (then to cascade to all tiers).

base_tier_validation_helper/models/tier_definition_helper.py
ps-tubtim: Fixed

base_tier_validation_helper/models/tier_definition_helper.py
ps-tubtim: Fixed

base_tier_validation_helper/models/tier_definition_helper.py
ps-tubtim: Fixed

base_tier_validation_helper/models/tier_definition_helper.py
ps-tubtim: Fixed

base_tier_validation_helper/models/tier_definition_helper.py
ps-tubtim: Edited

base_tier_validation_helper/models/tier_definition_helper.py
ps-tubtim: Fixed

base_tier_validation_helper/models/tier_definition_helper.py
ps-tubtim: Fixed

base_tier_validation_helper/models/tier_definition_helper.py
ps-tubtim: Fixed

base_tier_validation_helper/models/tier_definition_helper.py
ps-tubtim: Edited

base_tier_validation_helper/models/tier_definition_helper.py
ps-tubtim: Fixed

base_tier_validation_helper/models/tier_definition_helper.py
ps-tubtim: Fixed

base_tier_validation_helper/models/tier_definition_helper.py
ps-tubtim: Fixed

base_tier_validation_helper/models/tier_definition_helper.py
ps-tubtim: Fixed

base_tier_validation_helper/models/tier_definition_helper.py
ps-tubtim: Fixed

base_tier_validation_helper/models/tier_definition_helper.py
ps-tubtim: Edited

base_tier_validation_helper/models/tier_definition_helper.py
ps-tubtim: Fixed

base_tier_validation_helper/models/tier_definition_helper.py
ps-tubtim: Fixed

base_tier_validation_helper/models/tier_definition_helper.py
ps-tubtim: Fixed

base_tier_validation_helper/models/tier_definition_helper.py
ps-tubtim: Fixed

ps-tubtim:
@kittiu @hveficent Thank you for review. Edited.

LoisRForgeFlow:
@kittiu Is this ready?

kittiu:
@LoisRForgeFlow we may have the real use case later on in v13 or v14. But for now, can we keep this pending (if no one request for it yet).

PR #98, opened 2019-10-17
[12.0][IMP] base_tier_validation

ps-tubtim:
- Add comment on validation
- Add approve by sequence

Selection_106

1. If you check **Comment**, you can add comment on reviews table by wizard comment after 'validate' or 'reject'.
Selection_097
Selection_098

2. If you check **Approve by sequence**, approval will order by sequence.

PR #108, opened 2019-11-26
[12.0][IMP] base_tier_validation

ps-tubtim:


ps-tubtim:
duplicate with https://github.com/OCA/server-ux/pull/98

PR #110, opened 2019-11-26
[12.0][IMP]base_tchnical_features. Assign automatically to admin not to root

AaronHForgeFlow:
There is no sense of root to be assigned to this group.


StefanRijnhart (approved):

pedrobaeza (approved):
Totally makes sense

PR #115, opened 2019-12-07
improved for dynamic date range

EdyKend:
@api.onchange('date_range_id')
    def onchange_date_range_id(self):
        """Handle date range change."""
        if self.date_range_id:
            self.date_from, self.date_to = self.date_range_id.get_value()


lmignon:
@EdyKend Thank you for this proposal. Can you split your PR into 2 distinct PR: One fort the enhancement  of the data_range addon and an other one with your new addon?
On the other side, can you functionally describe your proposed changes / enhancement?

Regards,

lmi

PR #122, opened 2020-01-09
[FIX] protect import

hbrunn:
all imports that are not in odoo's requirements must be wrapped in an exception handler for versions below 13


guewen (approved):

simahawk (approved):

yajo:
> all imports that are not in odoo's requirements must be wrapped in an exception handler for versions below 13

FTR, this requirement is for versions below 12, where https://github.com/odoo/odoo/pull/17790 landed.

hbrunn:
thanks, I mix up versions quite often lately. We need to migrate our 7-11 customers to finally get rid of some baggage...

PR #123, opened 2020-02-02
[13.0][ADD] approvals_tier_validation [WIP]

kittiu:
This new module combine the goods of module **Approvals from EE** and **Base Tier Validation**

**Approvals**

Good:
  * All kind of approval request starts and managed from one place.

Not so good:
  * But since it is a single approval.request record, the data fields option are quite limited.
  * Approval workflow options is too simple for real life. Only allow to choose who are approvers.

**Tier Validation**

Good:
  * The Tier Validation (base_tier_validation) is far more flexible in term of approval workflow, even more flexible with base_tier_validation_formula.
  * Since it is the workflow extension to any kind of document, so the data fields are not limited.

Not so good:
  * It require to develop a new addons for a new kind of document.

approvals_tier_validation

**Approvals Tier Validation**

This module combine the good of the two modules above.
* Use the concept of Approvals to have single place to start the approve request.
* Use Base Tier Validation for workflow approval process.
* Create any model object as data (i.e., purchase order) and link it to the approve request.
* No need to create new addon for new type of document approval, just create new Approval Type

Note:
* You need to be Approval Admin in order to configure the new type of approval.
* You need to create tier definition for tier.approval.request

---

TODO:
  - [X] Ref Document under pending approval must not allowed for modification
  - [x] When start approval process, the Ref Document's state can be changed as approval request state changes. And when approval completed, the process can be changed too. I.e., Approval start, PO should change state to pending approval, and if rejected, the PO can be cancelled too.
  - [x] Ability to start the workflow from the original document (i.e., create PO first, when start the approval request).
  - [ ] Ability to start the work flow from Approval Request, but manual link to the existing document, instead of always create anew.
  - [x] Easy way to configure tier definition from the approval type window.
  - [x] Ref Document, if under approval, should have some kind of visible Sign on the document.
  - [x] Fine tune readonly state in Approval Request -> make sure no fields is editable on pending approval
  - [ ] Sequence on each Approval Type
  - [ ] Overall Code Review
  - [ ] Demo module

---

@LoisRForgeFlow @etobella @ps-tubtim 

This module is not completed, but the concept already works. Please feel free to comment, especially whether it can fit with the real life scenario.
Also, thank you to the original base_tier_validation, I think it fit in this quite well :+1:


approvals_tier_validation/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2017-19 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Approvals Tier Validation",
+    "version": "13.0.1.0.0",
+    "category": "Tools",
+    "website": "https://github.com/OCA/server-ux",
+    "author": "Ecosoft, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": ["base_tier_validation"],
```
LoisRForgeFlow: You need to add a dependency with `mail` as well.

LoisRForgeFlow (commented):
I tested this locally. Conceptually sounds good, I have nothing against it.

A couple of comments to improve testing expereince:
* add demo data Michel admin to the manager group by default.
* Add an icon to the app, for web_responsive or web_enterprise.

I was not able to create "Tier approvals types" maybe there is some more dependencies missing? It just keep loading forever when you click on cleate:
image

approvals_tier_validation/__manifest__.py
kittiu: @LoisRForgeFlow 
Thanks, if so, I will continue for the second iteration. And you guys can review it more.

Thanks!

LoisRForgeFlow: Great, would love to!

kittiu:
@LoisRForgeFlow this PR is almost done. There is small test failed in base_tier_validation
https://travis-ci.org/github/OCA/server-ux/jobs/696528250#L945
This mean, I can't have the app extension of base_tier_validation in this repo? As I need to add this,
https://github.com/ecosoft-odoo/server-ux/blob/13.0-add-approvals_tier_validation/approvals_tier_validation/models/tier_definition.py

kittiu:
This will not be continued, but just to keep it for reference.

PR #128, opened 2020-02-13
[11.0][FIX] base_tier_validation: When accepting or rejecting a review followers should not be no…

### LINKED ISSUE #125: [13.0] Tier validation subtype of messages
```
def _notify_accepted_reviews(self):
def _notify_rejected_review(self):
```

both these methods use subtype="mt_comment" which I think is incorrect. 

In for example a purchase order workflow we don't want to send a message to the supplier of the PO (which also follower) if a PO is internally rejected or approved.

I suggest I create a PR to change the subtype to mt_note.

@LoisRForgeFlow

LoisRForgeFlow:
…tified. A customer or supplier can be a follower of the object and should not receive internal messages about the tier validation.

Backport of #126.

Fixes #125.

cc @CasVissers @HviorForgeFlow

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -162,7 +162,7 @@ class TierValidation(models.AbstractModel):
         if hasattr(self, 'message_post'):
             # Notify state change
             getattr(self, 'message_post')(
-                subtype='mt_comment',
+                subtype='mt_note',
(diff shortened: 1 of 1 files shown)


HviorForgeFlow (approved):

etobella (approved):
LGTM :+1:

PR #129, opened 2020-02-25
[WIP] 12.0 add navigate m2x

legalsylvain:
This module allow to configure new server Actions to navigate between models.

for exemple, if you want to have the list of the products, that has been sold in some sale orders, you can configure a new Server Actions like this : 

ir_actions_server_form

then you'll have a new option in the "Action" list : 

sale_order_tree

Once the button clicked, the result will be a list of products : 

product_product_tree

Technically, it adds a new selection key ``navigate`` in the field ``state`` of the model ``ir.actions.server``.
(thanks to @hbrunn for [having me discovered](https://github.com/OCA/server-ux/pull/97#discussion_r346850594) this model. ;-))


elvise:
wooow

pedrobaeza:
Shouldn't it be then `server_action_navigate` and be in OCA/server-backend?

legalsylvain:
Hi @pedrobaeza Thanks for review ! 

> Shouldn't it be then server_action_navigate

I was not very inspired for the name of the module. No problem to change. I'll do it when other more deep review will be done. Thanks for the proposal. 

> and be in OCA/server-backend?

Well, I see this module very similar to [``mass_editing``](https://github.com/OCA/server-ux/tree/12.0/mass_editing) for exemple that is in the repository ``server-ux``. (add extra action on any models). 
server-backend are much more technical IMO.

pedrobaeza:
For me they are different, as `mass_editing` generate an UX element for users to edit massively things. With this module, you are only providing a technical tool for adding a feature when creating server actions, but no visible UX thing. That's why I think it should be in OCA/server-backend.

hbrunn:
Odoo SA massively promotes server actions as UI for the advanced consultant, so I think it's debatable whether this belongs in -backend. But I tend to agree with @pedrobaeza

legalsylvain:
OK Then.

image

PR #116, opened 2019-12-11
[12.0][IMP] base_tier_validation_formula

ps-tubtim:
From issue: https://github.com/OCA/server-ux/issues/114

- Add **Domain & Formula** in `definition_type`.

PR #132, opened 2020-02-28
[13.0][IMP] base_tier_validation_formula

ps-tubtim:
Cherry-pick from https://github.com/OCA/server-ux/pull/116

### DIFF
--- a/base_tier_validation_formula/README.rst
+++ b/base_tier_validation_formula/README.rst
@@ -36,8 +36,11 @@ and the tier reviewers using python code.
 Usage
 =====
 
-To define the domain by python code choose the Formula option in the Definition field.
-To define the reviewers by python code choose Python Expression option in the Validated by field.
+To define the domain,
+* By python code choose the **Formula** option in the Definition field.
+* By both domain and python code, choose the **Domain & Formula** option in the Definition field.
+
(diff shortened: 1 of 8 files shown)


LoisRForgeFlow (approved):
LGTM :+1:

Thanks!

AdriaGForgeFlow (approved):
Code review :+1:

LoisRForgeFlow:
I cancelled the merge because I just realized one thing: one of the commits is full of conlicts (probably of a cherry-pick) and the other is just fixing them, plus they have the same commit msg. I think it is not justified to have both commits, can you just squash them together?

LoisRForgeFlow (changes requested):
until commits are squashed (the first one intrudoced git-operation-generated conflicts ans second just fixed them, this should have been done in the operation and resulting in one clean commit)

HviorForgeFlow:
@ps-tubtim it seems bb6d3db still having HEADs. A way to solve it would be to squash both commits into one called ```[IMP] base_tier_validation_formula: Allow use both domain and formula in tier definitions``` for example, personally, I don't like to use version number on imp or fix commits then other contributors can cherry-pick them without change commit name.

LoisRForgeFlow:
Thanks, ready to merge:

/ocabot merge minor

LoisRForgeFlow (approved):

Issue #142, opened 2020-03-19
[13.0]mass_editing domain in fields


ValentinVinagre:
Hi,
I don't understand why the 'field_id' field have 'read_only = false' in its domain. Could @legalsylvain explain me this change?

https://github.com/OCA/server-ux/blob/9a7a80210f2defdf8ebc3acb133300c8447c47e3/mass_editing/models/mass_editing_line.py#L26

Thanks a lot,

legalsylvain:
because if readonly is set to True, so the field should not be edited. So not be mass edited.

ValentinVinagre:
Hi @legalsylvain ,
but Odoo have a lot of 'readonly' fields and if we can't change the value of that..... it's an important limitation.

Don't you think so?

pedrobaeza:
Not only that... sometimes the technique is to put readonly the field, but readonly=False in the view (or according certain criteria), so you can't impose that limitation.

legalsylvain:
> Not only that... sometimes the technique is to put readonly the field, but readonly=False in the view (or according certain criteria), so you can't impose that limitation.

OK. never seen that implementation, but no problem.

@ValentinoDefelice feel free to propose a PR.

pedrobaeza:
> OK. never seen that implementation, but no problem.

Well, it's the usual way Odoo implements it:

https://github.com/odoo/odoo/blob/6dff3024d19c9dd44c71992469540e3f09dd3247/addons/sale/models/sale.py#L156

You have pong incorrect user.

legalsylvain:
@pedrobaeza : thanks for the link.

@ValentinVinagre : could you make a PR ?

ValentinVinagre:
I can try it, but tomorrow 👍

PR #144, opened 2020-03-20
[13.0][FIX] mass_editing: allow change readonly fields.

ValentinVinagre:
With this solution it is possible to select read-only fields again.
Fixes https://github.com/OCA/server-ux/issues/142

### DIFF
--- a/mass_editing/models/mass_editing_line.py
+++ b/mass_editing/models/mass_editing_line.py
@@ -23,7 +23,6 @@ class MassEditingLine(models.Model):
         string="Field",
         domain="["
         "('name', '!=', '__last_update'),"
-        "('readonly', '=', False),"
         "('ttype', 'not in', ['reference', 'function']),"
         "('model_id', '=', model_id)"
         "]",


yajo:
IMHO this should only be available for superusers.

HaraldPanten:
But @Yajo , you can limit mass editing access by permissions groups/users. Why should we limit this usability only to super users?

ValentinVinagre:
Hi @Yajo ,
In this case, as @HaraldPanten said, you can limit the access using "Allowed Groups" field.
If you want, I can do:

1. Change the default value for this field and set "base.group_erp_manager" by default.
or
2. Create a new mass editing group, and set this group by default in "Allowed Groups" field. Of course, admin user would be included by default to this new group.

Even though, I don't agree limit mass editing action only to superusers. But with one of these options we can solve your requirement.

What do you think @pedrobaeza @legalsylvain ?

Captura de pantalla 2020-03-20 a las 12 43 40

pedrobaeza:
I agree with current approach where you limit the generated server action to the groups you want.

HaraldPanten (approved):
Functional review 👍

legalsylvain (approved):
Thanks. Agree with the solution.

pedrobaeza (approved):

Issue #130, opened 2020-02-26
[12.0][base_tier_validation] Validation Date not show timezone


Saran440:
Testing with hr_expense_tier_validation module -> https://github.com/OCA/hr/tree/12.0/hr_expense_tier_validation

When I Validation or Reject, Field Validation Date will show datetime not timezone.
![Selection_008](https://user-images.githubusercontent.com/20896369/75320826-1c3e1b80-58a2-11ea-8035-bd87b91ce957.png)
and I setting admin timezone is Asia/Bangkok (+07 GTM). It should be 14:12:06.
![Selection_009](https://user-images.githubusercontent.com/20896369/75320926-51e30480-58a2-11ea-8adc-828ca60e66cf.png)

LoisRForgeFlow:
@Saran440 can you open a PR addressing this?

Saran440:
@LoisRForgeFlow Yes, I can but I don't know how to fix. Do you have idea how to fix?

brenoalef:
Is there an estimate for when this will be merged?

kittiu:
@LoisRForgeFlow @brenoalef can you guys help move this forward? I think this is a good fix can be deployed to all other version easily.

Note: I also have another fix for approve_sequence problem waiting, but better have minor bugs cleared first.

Thank you all!

PR #145, opened 2020-03-25
[12.0][FIX] base_tier_validation : convert datetime timezone

Saran440:
FIX bug from issue https://github.com/OCA/server-ux/issues/130

- Convert datetime field Validation Date to timezone.

### DIFF
--- a/base_tier_validation/models/res_users.py
+++ b/base_tier_validation/models/res_users.py
@@ -50,4 +50,8 @@ class Users(models.Model):
             r['display_status'] = dict(
                 review_obj.fields_get('status')['status']['selection']
             ).get(r.get('status'))
+            # Convert to datetime timezone
+            if r['reviewed_date']:
+                r['reviewed_date'] = \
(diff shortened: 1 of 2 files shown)


kittiu:
Function tested & code review :+1: 
I think this good / simple fix should be included in all version soon.

@Saran440 can you split to 2 commits
1. fix time zone display.
2. add test coverage.

kittiu (approved):

kittiu:
@Saran440 sorry, is it ok to fix timezone only in this PR? For the test script, can do in another PR as it is not related to this fix.

Thank you!

kittiu:
@LoisRForgeFlow can this be fast forward? Then we can apply it to all other versions.
Thanks!

ps-tubtim (approved):

LoisRForgeFlow (approved):
LGTM

Thanks and sorry for the delay reviewing this.

kittiu:
@Saran440 would be nice if you port this fix to all versions. :+1:

PR #148, opened 2020-03-29
[ADD] mass editing groups

ovnicraft:
we came ussing this modules by several years, author does not migrati it, so we keep him but migrate it to 12.0
TODO

* [ ] Add README

### DIFF
--- a/mass_editing_groups/README.rst
+++ b/mass_editing_groups/README.rst
@@ -1,0 +1,83 @@
+==================
+Mass Editing Group
+==================
+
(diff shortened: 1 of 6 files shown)


pedrobaeza:
For me this can be integrated in the main module. I always do this creating the action and navigating through the action for assigning groups, but it's true this is a shortcut.

ovnicraft:
@pedrobaeza sounds good, i can integrate it to main `mass_editing` let me check it.

ovnicraft:
Closed in favor #149

PR #153, opened 2020-04-09
[12.0][ADD] base_search_custom_field_filter: Add custom filters for fields via UI

pedrobaeza:
This module allows to define custom filters in the search views for an specific field belonging to the document or any other related document.

This nature makes the definition quite technical, but once done, it adds the element in the UI for regular user use.

Configuration
=============

* Go to *Settings > Technical > User Interface > Custom Field Filters*.
* Create a new record, and define following information:

  * The **Model** for which you are defining the filter. It will appear in all
    the search views of this model.
  * The label you want to see on the search line on the **Name** field. This
    field allows translations for proper UI in different languages.
  * The **Expression**, which is the field chain string with dot notation.
    Examples: `product_id`, `product_id.seller_ids.name`, `partner_id.lang`.
  * Optionally, you can fill **Position After** for indicating after which
    existing field (technical name) the filter will appear. If empty or not
    found, the filter will be added at the end.
* You can reorder records for determining sorting for multiple filters for the
  same model with the arrow handle in the left part.

Usage
=====

* Go to the menu entry for which you have defined the custom field filter.
* On the search bar, type anything.
* In the filter list, you will see the line for the element you have defined.

As demo data, a custom field filter is included for sample purposes.

ir_ui_custom_field_filter

Steps for trying this sample:

* Install `contacts` module.
* Go to *Contacts*.
* Type "english" and you'll find the filter "Language" at the end.

contact_search

cc @Tecnativa TT23306


base_search_custom_field_filter/models/base.py
```diff
@@ -0,0 +1,59 @@
...
+                node = arch.xpath(
+                    "//field[@name='%s']" % custom_filter.position_after)
+            if not node:
+                node = arch.xpath("//field[last()]")
+            if node:
+                elem = etree.Element('field', {
+                    'name': 'ir_ui_custom_filter_%s' % custom_filter.id,
+                    'string': custom_filter.name,
+                    'filter_domain': (
+                        "[('%s', '=', 'self')]" % custom_filter.expression),
```
carlosdauden: Remove apostrophes to don't search "self" string literally
```suggestion
                        "[('%s', '=', self)]" % custom_filter.expression),
```

carlosdauden (changes requested):

base_search_custom_field_filter/models/ir_ui_custom_field_filter.py
```diff
@@ -0,0 +1,42 @@
...
+    sequence = fields.Integer()
+    model_id = fields.Many2one(comodel_name="ir.model", required=True)
+    model_name = fields.Char(
+        related="model_id.model", store=True, readonly=True, index=True,
+        string="Model name",
+    )
+    name = fields.Char(required=True, translate=True)
+    expression = fields.Char(required=True)
+    position_after = fields.Char(
+        help="Optional field name for putting the filter after that one. "
```
simahawk, line 22:
```suggestion
        help="Optional field name for putting the filter after another. "
```

base_search_custom_field_filter/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+This module allows to define custom filters in the search views for an specific
+field belonging to the document or any other related document.
+
+This nature makes the definition quite technical, but once done, it adds the
```
simahawk, line 4: you can probably drop this last sentence

simahawk (approved):
looks nice, tnx :)

base_search_custom_field_filter/models/base.py
carlosdauden: Changed here: https://github.com/Tecnativa/server-ux/pull/1

base_search_custom_field_filter/models/ir_ui_custom_field_filter.py
pedrobaeza: Please read it again. The sentence is saying exactly the purpose.

base_search_custom_field_filter/readme/DESCRIPTION.rst
pedrobaeza: Why? I want to highlight that. Remember this is general "functional user" README, and a warning about this is always advisable.

base_search_custom_field_filter/models/base.py
pedrobaeza: Thanks, typo error while transcribing the result of the investigation.

base_search_custom_field_filter/models/base.py
```diff
@@ -0,0 +1,59 @@
...
+        """Inject field definition. Needed to overwrite this as well due to
+        subsequent call to this after `fields_view_get`.
+        """
+        res = super().fields_get(allfields=allfields, attributes=attributes)
+        custom_filters = self.env['ir.ui.custom.field.filter'].search(
+            [("model_name", "=", self._name)])
+        for custom_filter in custom_filters:
+            field = custom_filter._get_related_field()
+            res['ir_ui_custom_filter_%s' % custom_filter.id] = (
+                field.get_description(self.env))
```
sergio-teruel: You can not add a "non-existent" field in fields get, this crash write models due to mail_thread model...
https://github.com/odoo/odoo/blob/cc6eb096f61d84f35585005a9f891f62adbb28e8/addons/mail/models/mail_thread.py#L553
If you add this field in this method crash becouse not exists in record
https://github.com/odoo/odoo/blob/cc6eb096f61d84f35585005a9f891f62adbb28e8/addons/mail/models/mail_thread.py#L317

sergio-teruel (changes requested):

base_search_custom_field_filter/models/base.py
pedrobaeza: Thanks for checking. Please try now with the latest push.

sergio-teruel: Ok. Fixed

base_search_custom_field_filter/models/base.py
```diff
@@ -0,0 +1,67 @@
...
+                node = arch.xpath(
+                    "//field[@name='%s']" % custom_filter.position_after)
+            if not node:
+                node = arch.xpath("//field[last()]")
+            if node:
+                elem = etree.Element('field', {
+                    'name': 'ir_ui_custom_filter_%s' % custom_filter.id,
+                    'string': custom_filter.name,
+                    'filter_domain': (
+                        "[('%s', '=', self)]" % custom_filter.expression),
```
sergio-teruel: Why is equal?? Can not be a like operator?

pedrobaeza: Yeah, good question, probably it should be `ilike`. Let me check.

chienandalu (approved):
Tested on runbot. Really cool! :+1:

pedrobaeza:
@Tardo, can you please check 280311f55d60898a6a4092089fbbf2333e3debb5?

base_search_custom_field_filter/models/base.py
pedrobaeza: Fixed other way without side effects.

base_search_custom_field_filter/static/src/js/search_input.js
```diff
@@ -0,0 +1,22 @@
...
+        make_domain: function (name, operator, facetValue) {
+            name = this.attrs.custom_field_filter || name;
+            return this._super(name, operator, facetValue);
+        },
+    });
+
+    ManyToOneField.include({
+        make_domain: function (name, operator, facetValue) {
+            name = this.attrs.custom_field_filter || name;
+            return this._super(name, operator, facetValue);
```
Tardo: To avoid linter warnings....
```suggestion
            var name_n = this.attrs.custom_field_filter || name;
            return this._super(name_n, operator, facetValue);
```

base_search_custom_field_filter/static/src/js/search_input.js
```diff
@@ -0,0 +1,22 @@
...
+    "use strict";
+
+    var core = require('web.core');
+    var search_inputs = require('web.search_inputs');
+    var ManyToOneField = core.search_widgets_registry.map.many2one;
+
+    search_inputs.Field.include({
+        make_domain: function (name, operator, facetValue) {
+            name = this.attrs.custom_field_filter || name;
+            return this._super(name, operator, facetValue);
```
Tardo: To avoid linter warnings...
```suggestion
            var name_n = this.attrs.custom_field_filter || name;
            return this._super(name_n, operator, facetValue);
```

Tardo (approved):
Good work! only a few non-blocking comments

pedrobaeza:
@Tardo changes done

PR #159, opened 2020-05-18
[13.0][MIG]base search custom field filter

chusamo:


base_search_custom_field_filter/models/base.py
```diff
@@ -25,7 +26,7 @@ def _add_custom_filters(self, res, custom_filters):
                     {
                         "name": "ir_ui_custom_filter_%s" % custom_filter.id,
                         "string": custom_filter.name,
-                        'custom_field_filter': custom_filter.expression,
+                        "filter_domain": "[('%s', 'ilike', self)]" % custom_filter.expression,
```
chusamo: Doing this, we don't need js to process "custom_field_filter" attribute in xml

pedrobaeza: That's our first approach :wink: but it doesn't work correctly in several cases, like several records sharing the same search string, so please revert to the previous JS one.

chusamo: Done

HaraldPanten (approved):
Functional review. Seems to be OK 👍

ValentinVinagre (approved):
LGTM 👍

ivantodorovich (approved):

PR #152, opened 2020-04-09
12.0 mig base import security group

ajayjaviya:

PR #161, opened 2020-05-22
[FIX] base_import_security_group browser test

StefanRijnhart:
The browser wait code was removed in the migration to 12.0: https://github.com/OCA/server-ux/pull/152/commits/55af0ca4d4544c352e947b27dccaae6d7a75d57d#diff-42657ea13cced82fd0efda56ecf8fef0R32
Indeed tests did not run with that line because the modified API of the browser tests now requires the waiting code to return a boolean True instead of just a truthy value. Fact that the tests succeeded at the time of merging could be a timing issue but as of late the tests seem to fail consistently.

### DIFF
--- a/base_import_security_group/tests/test_base_import_security_group.py
+++ b/base_import_security_group/tests/test_base_import_security_group.py
@@ -28,8 +28,8 @@ class TestImportSecurityGroup(common.HttpCase):
         """ % ('!' if falsify else '')
         action = self.env.ref('base.action_partner_category_form').id
         link = '/web#action=%s' % action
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

pedrobaeza:
Fast tracking as straight forward

/ocabot merge nobump

StefanRijnhart:
Thanks for the merge @pedrobaeza. It still does not work reliable, main branch is currently red. I've opened #162 and I'll just have to restart tests a number of times to see if that one helps.

PR #162, opened 2020-05-23
[FIX] Give travis some more time to determine if user is allowed to import

StefanRijnhart:
Branch is still intermittently red after https://github.com/OCA/server-ux/pull/161. Could be because browser tests are slower on Travis.


pedrobaeza:
Is this correct then?

StefanRijnhart:
Tests have now run ~~twice~~ three times without failure (but then that was the case on the previous PR). Restarting a fourth time.

pedrobaeza:
Seems safe now, so one more test letting the bot to launch an extra Travis CI:

/ocabot merge nobump

PR #169, opened 2020-06-11
[12.0] Reject Tier Validation

### LINKED ISSUE #168: [12.0] Unable to reject record when review type is set to expression
I've only tested this in version 12.0 but it might also be an issues in other versions.

OCA modules required to reproduce this issue:
- `purchase-workflow/purchase_request`
- `purchase-workflow/purchase_request_tier_validation`
- `server-ux/base_tier_validation`
- `server-ux/base_tier_validation_formula`

Assuming you have two users named, `User1` and `User2`. Both users also need to have access to the `purchase_request` module.

### To reproduce:
- Create a tier definition with the following:
    - Referenced Model: Purchase Request
    - Validated By: Python Expression
    - Python Expression: `rec.env['res.users'].search([('name', '=', 'User2')])`
    - Definition: Domain
    - Definition Domain: Match all records
- Login with `User1` create a new Purchase Request and click on the Request Validation button
- Login with `User2`, navigate to the newly created purchase request and click on the Reject button.

### Expected:
The purchase request should be rejected

### What's happening:
Nothing. When the reviewer clicks on the `Reject` button nothing happens.

### Details:
At first I thought this might be a bug with the `purchase_request_tier_validation` module. However, after digging around, I've discovered that it's an issue with the rejection code in the `base_tier_validation` modules.

I'll be submitting a PR with my solution to this problem but I would like some feedback on my approach as I'm new to Odoo development.

The issue lies with the `_rejected_tier` method on the `tier.validation` model. When it filters the review records it only checks the user against the review record's `reviewer_id` and `reviewer_group_id` which are only set when the review type is `individual` or `group` (See [here](https://github.com/oca/server-ux/blob/12.0/base_tier_validation/models/tier_validation.py#L270)). It should instead be checking if the current user is in `reviewer_ids` like the validation method `_validate_tier` does (See [here](https://github.com/oca/server-ux/blob/12.0/base_tier_validation/models/tier_validation.py#L180)).

Taar:
- Fix a bug when a definition's review_type is set to expression, the
user(s) are unable to reject the record.

When rejecting a record, the reviews are filtered using the records
reviewer_id and reviewer_group_id which are only set when the review
type is set to individual or group. To support other review types,
the filter is being changed to check the current user against the
review_ids field instead.

Fixes #168 

I wasn't able to find a way to create a test for this inside of the `base_tier_validation` module but I was able to write a test for it in the `base_tier_validation_formula` modules. If there is a way to test this please let me know and I'll update this PR.

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -267,8 +267,7 @@ class TierValidation(models.AbstractModel):
         tier_reviews = tiers or self.review_ids
         user_reviews = tier_reviews.filtered(
             lambda r: r.status in ('pending', 'approved') and
-            (r.reviewer_id == self.env.user or
-             r.reviewer_group_id in self.env.user.groups_id))
+            (self.env.user in r.reviewer_ids))
         user_reviews.write({
             'status': 'rejected',
             'done_by': self.env.user.id,
--- a/base_tier_validation_formula/tests/test_tier_validation.py
+++ b/base_tier_validation_formula/tests/test_tier_validation.py
@@ -171,7 +171,7 @@ class TierTierValidation(common.SavepointCase):
             'rec.env["res.users"].browse([{user_id}])'
         ).format(user_id=user_id)
 
-        definition = self.tier_def_obj.create({
+        self.tier_def_obj.create({
             'name': 'Over 10',
             'model_id': self.tester_model.id,
             'review_type': 'expression',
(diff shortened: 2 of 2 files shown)


Saran440:
@Taar Hello,
In v13 will refactor code for fix this issue and other.
Could you backport PR https://github.com/OCA/server-ux/pull/198 to v12?

PR #170, opened 2020-06-22
[10.0][BCK] base_search_custom_field_filter

rousseldenis:


rousseldenis:
@Cedric-Pigeon @pedrobaeza 

I've added tests we can maybe forwardport

rousseldenis:
We maybe add in ROADMAP to unlink these filters if fields are removed (e.g.: during module uninstall)

pedrobaeza:
Well, the tests checking `load_views` and `fields_view_get` are not testing the real thing, so they are some kind useless IMO (out of the "unit test" concept). The main thing is to check that this is shown in the UI and behaves correctly (there is a JS part that is injecting properly the domain for not having side effects when having records with the same value in the field).

rousseldenis:
> Well, the tests checking `load_views` and `fields_view_get` are not testing the real thing, so they are some kind useless IMO (out of the "unit test" concept). The main thing is to check that this is shown in the UI and behaves correctly (there is a JS part that is injecting properly the domain for not having side effects when having records with the same value in the field).

Don't agree, these methods are called by UI (tested in real). And it's 100% unit test as testing the code we are overriding...

pedrobaeza:
Yeah, but this is not assuring that the module works. In fact, JS for v10 is a bit different and our trick may not work.

rousseldenis:
> Yeah, but this is not assuring that the module works. In fact, JS for v10 is a bit different and our trick may not work.

Tests were not present in v12 (from which I backport), so, adding these is not bad... If you want to do the js ones as I'm not used to. The module is working fine in UI

pedrobaeza:
Yeah, indeed, but seeing that 100% coverage can give you the false sensation of module fully working, which may be not the case.

rousseldenis:
> Yeah, indeed, but seeing that 100% coverage can give you the false sensation of module fully working, which may be not the case.

Let people test it and get feedback :smiley:

simahawk (approved):
It works. Tested on http://3441927-170-3f716e.runbot1.odoo-community.org/web

sbidoul (approved):

PR #171, opened 2020-06-23
[13.0][MIG] base_import_security_group: adapt to v13 from v12

kurogeek:
Migrate module base_import_security_group from code base version 12 to version 13. There is no change in the code of the migration. Just copy and past from version 12. However, as I tested, it seems to work in version 13 too.


pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

PR #174, opened 2020-06-30
[IMP] base_tier_validation: notification hook

jarroyomorales:
This PR adds a hook so everyone can customize the notification subtype they prefer. While I understand this commit https://github.com/OCA/server-ux/commit/b9dc12fc8496f53f58ae9c663d6210a55430af24 , we were used to the notifications and now we have lost this funtionality.

@LoisRForgeFlow @CasVissers @etobella

### DIFF
--- a/base_tier_validation/__manifest__.py
+++ b/base_tier_validation/__manifest__.py
@@ -13,10 +13,10 @@
     "application": False,
     "installable": True,
     "depends": [
-        "web",
-        "bus",
+        "mail"
     ],
     "data": [
+        "data/mail_data.xml",
         "security/ir.model.access.csv",
         "views/tier_definition_view.xml",
         "views/tier_review_view.xml",
--- a/base_tier_validation/data/mail_data.xml
+++ b/base_tier_validation/data/mail_data.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo noupdate="1">
+    <record id="mt_tier_validation_accepted" model="mail.message.subtype" forcecreate="1">
+        <field name="name">Tier Validation Accepted Notification</field>
+        <field name="default" eval="True"/>
+        <field name="internal" eval="True"/>
+        <field name="hidden" eval="True"/>
+    </record>
+
+    <record id="mt_tier_validation_rejected" model="mail.message.subtype" forcecreate="1">
+        <field name="name">Tier Validation Rejected Notification</field>
+        <field name="default" eval="True"/>
+        <field name="internal" eval="True"/>
+        <field name="hidden" eval="True"/>
+    </record>
+</odoo>
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -187,11 +187,17 @@ class TierValidation(models.AbstractModel):
             rec = self.env[review.model].browse(review.res_id)
             rec._notify_accepted_reviews()
 
+    def _get_accepted_notification_subtype(self):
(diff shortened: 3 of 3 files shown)


pedrobaeza:
I think a new subtype should be created for this and be internal and default.

etobella (approved):
It does not change the current behavior but allows changes. LGTM

etobella (changes requested):
I agree with @pedrobaeza proposal, much more cleaner, you can maintain the hook if you like

jarroyomorales:
@etobella @pedrobaeza Ok, I also agree its a good option. I will keep the hook anyways.

etobella (approved):

CasVissers-360ERP:
Wouldn't it be better to have an accepted and rejected subtype?

EDIT:
Like the change!

LoisRForgeFlow (approved):
LGTM :+1:

LoisRForgeFlow:
Let's hold on the merge until @CasVissers 's comment gets answered.

jarroyomorales:
@CasVissers @LoisRForgeFlow @etobella If no one is against it I can separate the subtypes, since at the end it doesnt do any harm...

CasVissers-360ERP:
@jarroyomorales 
I would prefer to separate the subtypes, makes it as flexible as possible without great complexity.

CasVissers-360ERP (approved):
👍

jarroyomorales:
Can this get merged?

pedrobaeza:
/ocabot merge minor

To be ported to v13

pedrobaeza:
Ouch, there was a hidden error that has been discovered due to https://github.com/OCA/maintainer-quality-tools/pull/672. We need to fix it first before being able to merge.

etobella (changes requested):
With this change, it is required to set mail as a required element on base_tier_validation. Install / upgrade might not work properly

pedrobaeza:
Indeed, `mail` dependency is not set in manifest.

LoisRForgeFlow:
Sorry, I didn't refresh the page and didn't see latest commits, I aborte merge operation.

LoisRForgeFlow:
Regarding the problem, we have been avoiding the mail dependency for a long time but I guess it is not possible anymore... anyway I cannot really imaging an Odoo instance without mail installed.

jarroyomorales:
@LoisRForgeFlow @pedrobaeza @etobella You are right about the `mail` dependency. I have just added it.

pedrobaeza:
@LoisRForgeFlow see my comment in https://github.com/OCA/server-ux/pull/174#issuecomment-652531099

LoisRForgeFlow:
Oh, I saw that PR merged and I assumed it was fixed.

jarroyomorales:
> Ouch, there was a hidden error that has been discovered due to [OCA/maintainer-quality-tools#672](https://github.com/OCA/maintainer-quality-tools/pull/672). We need to fix it first before being able to merge.

This was already merged before the first attempt of merging this, so I don't know where the error might be coming from...

pedrobaeza:
hehe, the linked PR is merged and what does is to enable the skipped tests that were before, which means that now some branches will become red because their tests were not executed, which is the case of this branch. It's not your fault, but fixing the other module test is mandatory for being able to merge this through the bot.

etobella:
But this has no sense. The travis of the PR should fail, not only on the bot merging :thinking: 
Also, I made the installation of the same PR locally and it works as expected

AdriaGForgeFlow (approved):
Code review :+1:

LoisRForgeFlow:
thanks god my fix (do nothing) worked! :rofl:

PR #175, opened 2020-07-01
Adding the main attachment to the exceptions

brenoalef:
Without the added exception, it wouldn't be possible to add messages/notes with attachments, since it would raise a validation exception informing that the operation is under validation.


etobella (approved):

JordiBForgeFlow (approved):

ivantodorovich (approved):

etobella:
This is the same issue we found on #174 It seems related to a change on OCA/maintainer-quality-tools#672. However, It is strange, as it does not fail when checking the PR, only on merge

PR #179, opened 2020-07-08
[12.0][IMP] base_tier_validation: Added new feature to notify reviewers by sequence

oreju:
This change makes it that the person receives a notification email that he/she needs to approve an item only when it's his/her turn to approve.

### DIFF
--- a/base_tier_validation/__manifest__.py
+++ b/base_tier_validation/__manifest__.py
@@ -3,7 +3,7 @@
 {
     "name": "Base Tier Validation",
     "summary": "Implement a validation process based on tiers.",
-    "version": "12.0.3.4.0",
+    "version": "12.0.3.4.1",
     "development_status": "Mature",
     "maintainers": ['LoisRForgeFlow'],
     "category": "Tools",
--- a/base_tier_validation/__manifest__.py
+++ b/base_tier_validation/__manifest__.py
@@ -3,7 +3,7 @@
 {
     "name": "Base Tier Validation",
     "summary": "Implement a validation process based on tiers.",
-    "version": "12.0.3.4.1",
+    "version": "12.0.3.4.2",
     "development_status": "Mature",
     "maintainers": ['LoisRForgeFlow'],
     "category": "Tools",
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -183,9 +183,23 @@ class TierValidation(models.AbstractModel):
             'done_by': self.env.user.id,
             'reviewed_date': fields.Datetime.now(),
         })
-        for review in user_reviews:
-            rec = self.env[review.model].browse(review.res_id)
-            rec._notify_accepted_reviews()
+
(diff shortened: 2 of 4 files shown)


base_tier_validation/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Base Tier Validation",
     "summary": "Implement a validation process based on tiers.",
-    "version": "12.0.3.4.0",
+    "version": "12.0.3.4.2",
```
etobella, line 6: Do not change the version if no migrations are required, This simplifies the cherry-picks

base_tier_validation/models/tier_validation.py
```diff
@@ -318,7 +365,14 @@ def request_validation(self):
                                 'requested_by': self.env.uid,
                             })
                     self._update_counter()
-        self._notify_review_requested(created_trs)
+
+        if created_trs and \
+            not created_trs[0].definition_id.notify_on_create and \
```
etobella, line 370: Why the [0]

base_tier_validation/models/tier_validation.py
```diff
@@ -297,6 +315,35 @@ def _notify_review_requested(self, tier_reviews):
...
+                    message_type='email',
+                    body=rec._notify_requested_review_body(),
+                    partner_ids=tier_review.reviewer_ids.mapped(
+                        "partner_id").ids)
+
+    def _notify_review_accepted_requested_by_sequence(self, tier_review):
+        if hasattr(self, 'message_post') and \
+                hasattr(self, 'message_subscribe'):
+            for rec in self:
+                getattr(rec, 'message_subscribe')(
```
etobella, line 336: If you have checked the `message_subscribe`, you can do rec.message_subscribe, no need of getattr

base_tier_validation/models/tier_validation.py
```diff
@@ -183,9 +183,23 @@ def _validate_tier(self, tiers=False):
...
             'reviewed_date': fields.Datetime.now(),
         })
-        for review in user_reviews:
-            rec = self.env[review.model].browse(review.res_id)
-            rec._notify_accepted_reviews()
+
+        user_pending_reviews = tier_reviews.\
+            filtered(lambda r: r.definition_id.notify_by_sequence
+                               and r.status == 'pending')
+        if user_pending_reviews:
```
etobella, line 190: What happens if some have notify_by_sequence and some not?

etobella (changes requested):

base_tier_validation/models/tier_validation.py
oreju: Sorry for late answer,
Yes, I wanted to do as you say. But most of the way in the code was using like that https://github.com/OCA/server-ux/blob/12.0/base_tier_validation/models/tier_validation.py#L263-L269 .

base_tier_validation/__manifest__.py
oreju: Ok, I will fix it. Thanks.

PR #181, opened 2020-07-13
[IMP] base_tier_validation: Make test more resilient with sales and purchase modules installed

sergio-teruel:
When run test with sales and purchases tier validation modules installed the base_tier_Validation test fail because the other modules add elements to validation model names list.

cc @Tecnativa

ping @carlosdauden ping @HviorForgeFlow

### DIFF
--- a/base_tier_validation/tests/test_tier_validation.py
+++ b/base_tier_validation/tests/test_tier_validation.py
@@ -155,7 +155,14 @@ class TierTierValidation(common.SavepointCase):
     def test_09_dummy_tier_definition(self):
         """Test tier.definition methods."""
         res = self.tier_def_obj._get_tier_validation_model_names()
-        self.assertEqual(res, [])
+        expected_tier_validation_model_name = []
+        module_id = self.env.ref("base.module_sale_tier_validation")
+        if module_id.state == "installed":
+            expected_tier_validation_model_name.append("sale.order")
+        module_id = self.env.ref("base.module_purchase_tier_validation")
+        if module_id.state == "installed":
+            expected_tier_validation_model_name.append("purchase.order")
+        self.assertSetEqual(set(res), set(expected_tier_validation_model_name))
 
     def test_10_systray_counter(self):
         # Create new test record
--- a/base_tier_validation/tests/test_tier_validation.py
+++ b/base_tier_validation/tests/test_tier_validation.py
@@ -152,22 +152,6 @@ class TierTierValidation(common.SavepointCase):
         )
         self.assertTrue(res)
 
-    def test_09_dummy_tier_definition(self):
-        """Test tier.definition methods."""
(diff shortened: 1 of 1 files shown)


base_tier_validation/tests/test_tier_validation.py
```diff
@@ -155,7 +155,14 @@ def test_08_search_validated(self):
     def test_09_dummy_tier_definition(self):
         """Test tier.definition methods."""
         res = self.tier_def_obj._get_tier_validation_model_names()
-        self.assertEqual(res, [])
+        expected_tier_validation_model_name = []
+        module_id = self.env.ref("base.module_sale_tier_validation")
```
carlosdauden: You can add raise_if_not_found=False and check if returns something or None
```suggestion
        is_installed = bool(self.env.ref("base.module_sale_tier_validation", raise_if_not_found=False))
```

carlosdauden (changes requested):

base_tier_validation/tests/test_tier_validation.py
```diff
@@ -155,7 +155,14 @@ def test_08_search_validated(self):
     def test_09_dummy_tier_definition(self):
         """Test tier.definition methods."""
         res = self.tier_def_obj._get_tier_validation_model_names()
-        self.assertEqual(res, [])
+        expected_tier_validation_model_name = []
+        is_installed = bool(self.env.ref("base.module_sale_tier_validation", raise_if_not_found=False))
```
sergio-teruel: You can have the module but in state not installed...

sergio-teruel:
@carlosdauden changes done

base_tier_validation/tests/test_tier_validation.py
carlosdauden: It is true

carlosdauden (approved):

etobella:
There are other OCA modules using tier_validation, like:
https://github.com/OCA/stock-logistics-warehouse/tree/12.0/stock_request_tier_validation

You can find them using this link:
https://github.com/search?l=Python&q=org%3AOCA+base_tier_validation&type=Code

pedrobaeza:
This test doesn't make any sense in a base module and seems to be put for "completing" coverage, but it's clearly a bad focused test and IMO it should be removed.

sergio-teruel:
For me ok to be removed

pedrobaeza:
The only sense to have that test is if:

1. You have a demo model where tier validation is applied.
2. The test checks that this model is included in the method.

If that can't be done, better to remove the test.

sergio-teruel:
So, do I remove the test?

pedrobaeza:
Yes if the other thing can't be done.

sergio-teruel:
Ok. Test removed

pedrobaeza (approved):

PR #193, opened 2020-08-13
[12.0][ADD] chained_swapper: new module

ernestotejeda:
Chained Swapper
===========

This module allows to swap the value of a field and propagate it in a chained way to linked records. Example: changing the delivery address in a confirmed sales order, it should be changed in its delivery orders as well.

It also allows to apply constraints for not allowing to do that change according rules, so the business logic is not broken. Example: Don't allow to change the delivery address if the delivery order is validated.

This module requires some technical knowledge for setting the chained swap and the constraint, as it's defined through technical names and Python code.

Peek 17-08-2020 10-02

cc @Tecnativa TT20613


pedrobaeza (approved):

ernestotejeda:
@pedrobaeza take a look at the last commit

pedrobaeza:
Great, let's involve a third person in the PR to be validated: @CarlosRoca13 may you?

CarlosRoca13 (approved):

HaraldPanten (approved):
Functional review. LGTM Tested on Runbot.

PR #194, opened 2020-08-24
[12.0][FIX] improve perf and solve compatibility with base_jsonify

sebastienbeau:
# Issue

Installing shopinvader (depend of base_jsonify) break if we install it after base_export_manager

# Details and solution

The inverse method on the name set the value of field1_id, field2_id...
As the name depend of the field1_id, field2_id we pass in the
_compute_name everytime we set one field and it's useless.

This generate the compatibility issue with base_jsonify and
shopinvader (and possibility other module) as the name computed
can be a dependency of other field.

For example (with base_jsonify) if I configure an ir.exports.line with
- name : "country_id/name"
- alias : country_id:countries/name

As the name is recomputed when we set the field1_id to "country_id" we
will have a name with the value "country_id" instead of
"country_id/name" and this do not match the alias and raise an error

Recomputing at the end of the inverse_name improve the perf and solve
the issue

@lmignon

### DIFF
--- a/base_export_manager/models/ir_exports_line.py
+++ b/base_export_manager/models/ir_exports_line.py
@@ -127,33 +127,32 @@ class IrExportsLine(models.Model):
     @api.multi
     def _inverse_name(self):
         """Get the fields from the name."""
-        for one in self:
-            # Field names can have up to only 4 indentation levels
-            parts = one.name.split("/")
-            if len(parts) > 4:
-                raise exceptions.ValidationError(
-                    _("It's not allowed to have more than 4 levels depth: "
-                      "%s") % one.name)
-            for num in range(1, 5):
-                if num > len(parts):
-                    # Empty subfield in this case
+        with self.env.norecompute():
+            for one in self:
+                # Field names can have up to only 4 indentation levels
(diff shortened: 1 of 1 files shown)


sebastienbeau:
@hparfr @rousseldenis @Yajo @pedrobaeza

pedrobaeza:
Is there a way to exercise the problem in a test?

lmignon (approved):
@sebastienbeau All this makes sense. Not sure the performance gain will boost the application but at least your changes will replace 4 potential calls to `_compute_name(self):` by a unique call. (regardless of the depth of the path expressed by the name), Moreover, IMO, it also ensure the atomicity the `compute_name`. This method is designed to work with a fully initialised record. Before your change, since all the `fieldx_id` was reset at once and thereafter assigned one by one means that the `compute_name` was not called with all the required information until all values have been assigned to the fieldx_id.

rousseldenis (approved):
Makes sense to me

rousseldenis:
@len-foss this can be in your interest

simahawk:
@sebastienbeau any test to update?

Issue #192, opened 2020-08-11
[RFC] Enhancement to base_tier_validation


kittiu:
We have the need to find a more flexible approval workflows engine. I want to try by extending this base_tier_validation first.

For example,

- [ ] Tier that auto by pass with out approval, but instead call server action - #207 
- [ ] Option to "Escalate" to 3rd person https://github.com/OCA/server-ux/pull/197
- [ ] ~~Option to Undo validation/rejection, when next step is not done~~ (not doing it)
- [ ] Persistent sub-state on document, for searching, reporting, etc. https://github.com/OCA/server-ux/pull/203
- [ ] ~~More report on approval status~~ (not doing it)
- [ ] ~~Audit trails, even if the flow is restarted~~ (not doing it)
- [ ] etc...

Sorry, it is still arbitrary ideas. But roughly, a more flexibility version of approval workflow.

I will starting doing it one by one, some as enhancement, and some as module as it deem appropriate.

cc @LoisRForgeFlow, @pedrobaeza 

---

Additional Improvements (added 14-Oct-2020)

- [ ] Replace notification system tray, with Odoo's Activity infrastrucgture.
- [ ] Revisit Base Tier Validation Helper (to group many tiers in a validation.type object for ease of management + more control)
- [ ] Auto start validation workflow, based on approving document status (i.e., Purchase Order is RFQ sent, start validation process)
- [ ] New reviewers selection option, that read HR Employee tables to get the eligible employee for validation condition (i.e., department, approval limit of each employee) 
- [ ] Tier Validation Substate enhancements, i.e., better looking UI of substate, no manual substate change, etc.
- [ ] Customised validation error messages - Business Messages rather than the system ones (explore base_exception)

nmoturi:
Hi Kitti,

other improvement list :

1. Customised validation error messages - Business Messages rather than the system ones
2. Also, can this be worked along side with base_exception OCA module..? Some of the functionality seems to be an overlap.

-Naran

kittiu:
Sure. But can you explain more. Especially on point 2.

kittiu:
@nmoturi I added the improvements in the list.

kittiu:
Among the above enhancement, I think the "Replace notification system tray, with Odoo's Activity infrastructure" is quite harsh and need more opinions.
Functionally, I think it make sense, as the tray came before activity, and once we have activity, it is more unified.

@LoisRForgeFlow do you have comment on this?

[...]

PR #196, opened 2020-08-27
[13.0][ADD] base_tier_validation_action [WIP]

kittiu:
Add option in definition to,

**1. Trigger server action after validate a tier**

image

**2. Auto validate tier**

image

* After validate the last step of PO, do "Confirm" the PO too.
* If user is on vacation and the amount is less than 1000, auto validate the PO.
* etc.


kittiu:
This PR is part of the RFC  - https://github.com/OCA/server-ux/issues/192

Per say, this is actually 2 separated feature, yet quite minimal. In my opinion, I think this one can should be part of `base_tier_validation` or otherwise, it should be split into 2 `base_tier_validation_action`, `base_tier_validation_auto`

Any comment for this WIP are welcome?

LoisRForgeFlow:
Is this still WIP?

kittiu:
Yes, it is still WIP. Would like to hear some more comment.
I want to propose this as ENH of base_tier_validation, I think this is rather simple and can be useful.

WDYT?

LoisRForgeFlow:
@kittiu Yes, I think this can be added in the core module. I'll be glad to review when it is ready :)

kittiu:
Moved to https://github.com/OCA/server-ux/pull/207

PR #197, opened 2020-08-29
[13.0][ENH] base_tier_validation, add escalation option

kittiu:
Add new feature: **Escalation**
- Allow Escalation in tier definition
- For that tier review, user has option to also Escalate
- Escalation means validate current tier and create extra tier for another reviewer

Pending issue:
- On escalated tier, when user make comment, it will set on the review. Because, to get the reviews, current logic needs definition first, I think it is not valid and require fixing too https://github.com/OCA/server-ux/pull/198

Depends on this fix:
- [X] https://github.com/OCA/server-ux/pull/198

Following is the example, we have 3 tiers approval by sequence for Purchase https://github.com/OCA/purchase-workflow/pull/967, where step 2 will allow user2 to Escalate to step 2.1. And only if step 2.1 is done, user 3 can proceed with step 3.


base_tier_validation/models/tier_review.py
```diff
@@ -20,22 +21,40 @@ class TierReview(models.Model):
...
+        readonly=False,
     )
     reviewer_ids = fields.Many2many(
         string="Reviewers",
         comodel_name="res.users",
         compute="_compute_reviewer_ids",
         store=True,
     )
-    sequence = fields.Integer(string="Tier")
+    sequence = fields.Float(string="Tier")
```
kittiu, line 52: kittiu: change to float because escalated tier will be insertion, i.e., 1.1, 2.1 etc.

base_tier_validation/models/tier_review.py
```diff
@@ -7,8 +7,9 @@
 class TierReview(models.Model):
     _name = "tier.review"
     _description = "Tier Review"
+    _order = "sequence"
 
-    name = fields.Char(related="definition_id.name", readonly=True)
+    name = fields.Char(compute="_compute_definition_data", store=True, readonly=False)
```
kittiu, line 12: kittiu: I need to use compuate store=True, readonly=False of v13, because the extra escalated tier will not have definition. But still need these fields value.

kittiu:
@pedrobaeza if you could share your opinion on Escalation feature that would be awesome. This feature is quite varied from user I ask, but here I try my best to make it simple and conform with existing.

pedrobaeza:
Functionally seeing the GIFs (thanks for them), then only thing I see weird is that the step 2 is logged as approved instead of a new status "Escalated".

kittiu:
Good point. I will add the new status "Escalated". Will be like this,

image

kittiu:
Ready for review.

HviorForgeFlow (commented):
Functional Review: :+1:
Code Review: I would rename escalate wizard to tier validation escalate wizard

kittiu:
@HviorForgeFlow done changed model name from `escalate.wizard` to `validation.escalate.wizard`

HviorForgeFlow:
I would prefer tier.validation.escalate.wizard sorry by the nitpicking

HviorForgeFlow (approved):

ps-tubtim (approved):
The code and functional test are LGTM.

kittiu:
> I would prefer tier.validation.escalate.wizard sorry by the nitpicking

Done :+1:

nmoturi:
Hi Kitti,

Is the base tier escalation enhancement includes the other enhancement related to server action & autovalidation..?

Thanks,
Naran

kittiu:
> Hi Kitti,
> 
> Is the base tier escalation enhancement includes the other enhancement related to server action & autovalidation..?
> 
> Thanks,
> Naran

Nopl. You need to test them separately for now. Until both get merged.

kittiu:
@HviorForgeFlow can this get merged? Thank you!

LoisRForgeFlow (commented):
This is starting to get more and more complex and I feel I don't have use cases for some of this features while this changes can impact existing installations. Would it be possible to extract this kind of extended features to a separate module? then we can add a section in the settings wizard to activate extension on tier validations. Also, this is particular feature (escalation) it is not really documented (not even a help string in the has_escalation field) to help users to understand how it is going to behave.

I think if we can extract features in smaller modules and improve functional documentation, we will improve UX a lot.

LoisRForgeFlow:
Also, if you separate in different modules it will be easier to mantain because the migrations will be more granular and the merges on changes faster because the impact will be limited to a module which is installed in less instances (you can even be maintainer of this subfeatures and control the release of changes).

kittiu:
@LoisRForgeFlow yes, sure. I think this one can. This is your module, I want you to have the final say for things like this :) Will notify you after it is done.

Note: the thing is I do have many things I wanted to make base_teir_validation works more flexible, but I also aware about the complexity added into it. Your guidance would be much appreciated. https://github.com/OCA/server-ux/issues/192

kittiu:
@LoisRForgeFlow https://github.com/OCA/server-ux/pull/220

Should we split this one too https://github.com/OCA/server-ux/pull/207 (to base_tier_validation_action)? I am ok, as it will be inline altogether.

LoisRForgeFlow:
@kittiu yes, if you don't mind I rather make base_tier_validation_action also separate.

Issue #195, opened 2020-08-27
[12,13][BUG] base_tier_validation, for case sequence approval


kittiu:
I have tested approval by sequence, for 3 steps approval. Where step 1 and 3 is by **user1**, and step 2 is for **user2**

I would expect that, it will run by sequence no matter what. But when user1 approve, it result like this.

![image](https://user-images.githubusercontent.com/1973598/91408024-5d24e600-e86d-11ea-86df-346d77560dbc.png)

I understand that it might seem like a feature, because without sequence approval, when a user approve, all steps by that users will get approved.

But I still think, it will be more intuitative in case approval by sequence, to go one step at a time only. @ps-tubtim , can you fix?

cc @LoisRForgeFlow WDYT?

kittiu:
Fixed here, https://github.com/OCA/server-ux/pull/198

nmoturi:
I tested the sequence flow, but for me it seems to be still a problem how the approval chain needed to cascade when seq is defined.

Here is the screenshot, I was expecting when a sequence is defined for the same validation the review seq's counter needs to be as per the seq flow. In the example shown below I was able to validate the higher review seq first and the rest later..

Not sure if this is a desired outcome if yes, whats the use of "Approve by sequence" option in the definition..
![image](https://user-images.githubusercontent.com/7597063/92319453-a807e000-f05b-11ea-80f6-f8e2dfa8cde9.png)

Thanks,
Naran

PR #198, opened 2020-08-29
[13.0][FIX] base_tier_validation, fix approve by sequence problem

kittiu:
- When use approve by sequence, still approve tier of the same reviewer https://github.com/OCA/server-ux/issues/195
- When using approve_sequence option in any tier.definition there can be inconsistencies in the systray notifications

### DIFF
--- a/base_tier_validation/models/res_users.py
+++ b/base_tier_validation/models/res_users.py
@@ -13,19 +13,22 @@ class Users(models.Model):
     def review_user_count(self):
         user_reviews = {}
         to_review_docs = {}
-        for review in self.env.user.review_ids.filtered(
-            lambda r: r.status == "pending"
-        ):
+        reviews = self.env["tier.review"].search(
+            [
+                ("status", "=", "pending"),
+                ("can_review", "=", True),
+                ("id", "in", self.env.user.review_ids.ids),
+            ]
+        )
+        for review in reviews:
             record = (
                 review.env[review.model]
                 .with_user(self.env.user)
                 .search([("id", "=", review.res_id)])
             )
-            if not record:
+            if not record or record.rejected or not record.can_review:
                 # Checking that the review is accessible with the permissions
-                continue
-            can_review = record.can_review
-            if not can_review:
+                # and to review condition is valid
                 continue
             if not user_reviews.get(review["model"]):
                 user_reviews[review.model] = {
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -29,7 +29,9 @@ class TierValidation(models.AbstractModel):
         compute="_compute_validated_rejected", search="_search_validated"
     )
     need_validation = fields.Boolean(compute="_compute_need_validation")
-    rejected = fields.Boolean(compute="_compute_validated_rejected")
+    rejected = fields.Boolean(
+        compute="_compute_validated_rejected", search="_search_rejected"
+    )
     reviewer_ids = fields.Many2many(
         string="Reviewers",
         comodel_name="res.users",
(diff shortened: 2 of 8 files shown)


kittiu:
@ps-tubtim can you help review this?

kittiu:
2 Problems I found before this fix.

1. Even without "approve_sequence", user1 reject step1, the document is "rejected", but user2 still see the document on tray (because step2 is pending)

2. With "approve_sequence", there is this problem #195, as well as the invalid tray counter even after that #195 was fixed. For example, with user1 on step1 and step3, and user2 on step2. When user1 approve, the tray for user1 should disappear because he is not the next approve, but it is not.

Now, those problems were fixed.

HviorForgeFlow (approved):
Function Review LGTM

base_tier_validation/models/res_users.py
```diff
@@ -13,19 +13,22 @@ class Users(models.Model):
...
-            lambda r: r.status == "pending"
-        ):
+        reviews = self.env["tier.review"].search(
+            [
+                ("id", "in", self.env.user.review_ids.ids),
+                ("status", "=", "pending"),
+                ("can_review", "=", True),
+            ]
+        )
+        for review in reviews:
```
ps-tubtim: Why you use this code instead 
`for review in self.env.user.review_ids.filtered(lambda r: r.status == "pending" and r.can_review):` 
?

kittiu: Because I think the performance will be better then using filtered()

ps-tubtim (approved):
Code LGTM

Saran440 (approved):
Functional Test :+1:

base_tier_validation/models/res_users.py
LoisRForgeFlow: I think the use of o2m fields has also some performance optimizations, have you compared both options?

base_tier_validation/models/tier_validation.py
```diff
@@ -42,30 +44,25 @@ class TierValidation(models.AbstractModel):
...
-        sequence = reviews.mapped("sequence")
-        sequence.sort()
-        my_sequence = sequence[0]
-        if self.review_ids.filtered(
-            lambda r: r.status != "approved" and r.sequence < my_sequence
-        ):
-            # A review has not been approved
-            return False
-        return True
+        # Include only my_reviews with accprove_sequence = True
```
LoisRForgeFlow: little typo here

base_tier_validation/models/res_users.py
```diff
@@ -13,19 +13,22 @@ class Users(models.Model):
...
             record = (
                 review.env[review.model]
                 .with_user(self.env.user)
                 .search([("id", "=", review.res_id)])
             )
-            if not record:
+            if not record or record.rejected or not record.can_review:
```
LoisRForgeFlow, line 29: is this something you found problmeatic? I mean being able to validate previously rejected reviews?

base_tier_validation/models/tier_validation.py
```diff
@@ -246,23 +243,27 @@ def _add_comment(self, validate_reject):
...
                 "default_res_model": self._name,
-                "default_definition_ids": definition_ids.ids,
+                "default_review_ids": reviews.ids,
                 "default_validate_reject": validate_reject,
             },
         }
 
     def validate_tier(self):
         self.ensure_one()
+        sequences = self._check_approve_sequence(self.env.user)
```
LoisRForgeFlow: maybe this is no longer "_check_approve_sequence" but something like "_get_sequences_to_approve", what do you think?

LoisRForgeFlow (commented):
Hi, this looks good, thanks for the fix! See some comments below

base_tier_validation/models/res_users.py
kittiu: No, but I think when time passed, the user.review_ids can have so much records. Let's say 100,000 records. The filter operation can take up time (it will go through record one by one).
But if we use search with `status` and `can_review`, it can reduce a lot of records already.

My assumption is filter() is slower than searching a field. Is it? What is your opinion to this, I am open.

base_tier_validation/models/res_users.py
kittiu: Yes, I found this problematic with the counter. Even when the document is rejected, but the review counter still increase because it count other reviews.

base_tier_validation/models/tier_validation.py
kittiu: OK, I agree. Fixed.

nmoturi:
I tested the sequence flow, but for me it seems to be still a problem how the approval chain needed to cascade when seq is defined.

Here is the screenshot, I was expecting when a sequence is defined for the same validation the review seq's counter needs to be as per the seq flow. In the example shown below I was able to validate the higher review seq first and the rest later..

image

Not sure if this is a desired outcome if yes, whats the use of "Approve by sequence" option in the definition..

Thanks,
Naran

kittiu:
You need to set approve_sequence = True for all defintion you want to respect sequence

nmoturi:
I did set the approve_sequence = True, here is the screenshot

image

kittiu:
> In the example shown below I was able to validate the higher review seq

As I see in the picture, you validate sequence 1 first, and then sequence 2, right? That's already correct.

Note:
* In the definition, higher number means the earlier the sequence in the review table. Your case, 20 -> 1, and 10 -> 2. And 1 come before 2 in validation step.
* I know, it sound contradict. But the original author of this module decided this, and I didn't change this concept. You may ask reason behind this from module maintainer.

nmoturi:
Thanks Kitti..Got it now..I haven't thought higher seq would be validated first, bubbling from bottom to top, bit unusual to Odoo seq design.

Would be nice to have the reviewer user or group displayed in the reviewers list, at least that way the user/requester knows who needs to be approved.

All good, thanks Kitti for clarification .

image

kittiu:
Sure, thank you also for your testing as well. :)
For the more information, I also agree. But let's do it on separated [IMP] PR. I try to only do the fixing here.

base_tier_validation/models/res_users.py
LoisRForgeFlow: Let's keep your change, intuitively it looks that it will be better performance-wise.

LoisRForgeFlow (approved):
Hi @kittiu I have tested this functionally and fixes exactly what it describes, so let's move forward. Thanks!

/ocabot merge patch

LoisRForgeFlow:
> Thanks Kitti..Got it now..I haven't thought higher seq would be validated first, bubbling from bottom to top, bit unusual to Odoo seq design.
> 
> Would be nice to have the reviewer user or group displayed in the reviewers list, at least that way the user/requester knows who needs to be approved.
> 
> All good, thanks Kitti for clarification .
> 
> image

Hi @nmoturi 

Answering your suggestion, there are some problems with it:

- There is also the posibility to specify reviewers using a formula, so information of group/reviewer here could be misleading.
- Groups can be too technical for some users.
- I'm open to a new computed field that pretty displays the reviewers but, what do you show? Sometimes if you show users if can be a very long list... 

Due to this, we simple covered this need with the "name" field, so in each use case you can write what you need users to see. E.g: "Order above 1000$ - needs review of Purchase manager", "Deparment leaders", "Sift leaders", "CEO and COO"...

PR #201, opened 2020-08-31
[12.0][FIX] chained_swapper: Avoid singleton error to update records

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/chained_swapper/models/chained_swapper.py
+++ b/chained_swapper/models/chained_swapper.py
@@ -87,9 +87,10 @@ class ChainedSwapper(models.Model):
         self.field_id = False
 
     def write(self, vals):
-        super().write(vals)
+        res = super().write(vals)
         if 'name' in vals:
-            self.ref_ir_act_window_id.name = self.name
+            self.mapped("ref_ir_act_window_id").write({"name": vals["name"]})
+        return res
 
     @api.multi
     def unlink(self):


chained_swapper/models/chained_swapper.py
```diff
@@ -89,7 +89,9 @@ def _onchange_model_id(self):
     def write(self, vals):
         super().write(vals)
```
pedrobaeza:
```suggestion
        res = super().write(vals)
```
and return at the end

chained_swapper/models/chained_swapper.py
```diff
@@ -89,7 +89,9 @@ def _onchange_model_id(self):
...
         if 'name' in vals:
-            self.ref_ir_act_window_id.name = self.name
+            for rec in self:
+                if self.ref_ir_act_window_id:
+                    self.ref_ir_act_window_id.name = self.name
```
pedrobaeza:
```suggestion
            self.filtered("ref_ir_act_window_id").write({"name": vals["name"]})
```

pedrobaeza (changes requested):

sergio-teruel:
@pedrobaeza Changes done!!

chained_swapper/models/chained_swapper.py
pedrobaeza: Please do this also (not critical in this case, but good habit for avoiding side effects)

chained_swapper/models/chained_swapper.py
```diff
@@ -89,7 +89,7 @@ def _onchange_model_id(self):
     def write(self, vals):
         super().write(vals)
         if 'name' in vals:
-            self.ref_ir_act_window_id.name = self.name
+            self.filtered("ref_ir_act_window_id").write({"name": vals["name"]})
```
carlosdauden:
```suggestion
            self.mapped("ref_ir_act_window_id").write({"name": vals["name"]})
```

pedrobaeza: Sorry for my mistake. It's:

```suggestion
            self.mapped("ref_ir_act_window_id").write({"name": vals["name"]})
```

carlosdauden (changes requested):

sergio-teruel:
done

pedrobaeza (approved):
Thanks for the changes

carlosdauden (approved):

HaraldPanten (approved):

PR #202, opened 2020-09-04
[12.0] mass_editing: fix label disappearing with many2many field

Florian-Ch:
Apply the same fix as for v11.0. Link to the original commit here : https://github.com/OCA/server-ux/commit/d899e8d4be3aaa3845d481c5e0967e4e5fca2ead

### DIFF
--- a/mass_editing/wizard/mass_editing_wizard.py
+++ b/mass_editing/wizard/mass_editing_wizard.py
@@ -29,14 +29,6 @@ class MassEditingWizard(models.TransientModel):
                 'colspan': '6',
                 'col': '6',
             })
-            etree.SubElement(xml_group, 'label', {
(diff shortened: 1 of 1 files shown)


simahawk:
@Florian-Ch you should preserve the ownership of the original commit by @Martronic-SA if possible. How have you ported it?

Florian-Ch:
I have forked the repo and add @Martronic-SA 's part of code to the 12.0 branch. @Martronic-SA has all credit for this.
If I can preserve the ownership of the commit I'll do this, I thought a link in description was the best idea

pedrobaeza:
You can use git cherry-pick git command for retrieving the exact commit (having fetched it first, that is to say that is in your local repo).

PR #199, opened 2020-08-30
[13.0][MIG] base_substate

kittiu:
Standard migration

PR #203, opened 2020-09-04
[13.0][ADD] base_tier_validation_substate

kittiu:
This is the glue module for `base_tier_validation` and `base_substate`.

In each Tier Definition, user has the option to set Approved Substate and Rejected Substate.
Based on validated (or rejected) tier, the substate will be assigned automatically to the document.

Dependency
- [ ] base_subtate https://github.com/OCA/server-ux/pull/199


nmoturi:
Hi Kitti,

I know this module is still in WIP, was trying to test through the current functionality I was unable to add the substate configuration. Is any use case or a config you could share for it to work.

Thanks,
Naran

kittiu:
Hi @nmoturi 

I am migrating base_substate for v13. But that also WIP as I am waiting for the v12 improvement proposal https://github.com/OCA/sale-workflow/pull/955#issuecomment-686425631

In summary, my plan is,
- base_substate v12 is improved
- base_substate v13 is done #199
- base_tier_validation_substate here will be done.

But I think the above do not stop you from testing it. Following is the module I think you will need for testing.

1. https://github.com/ecosoft-odoo/server-ux/tree/13.0-mig-base_substate
2. purchase_substate as attached (I use it to test locally) --> for substate to work with purchase
3. https://github.com/OCA/server-ux/tree/13.0/base_tier_validation
4. https://github.com/OCA/purchase-workflow/tree/13.0/purchase_tier_validation --> for tier validation to work with purchase
5. This module

[purchase_substate.zip](https://github.com/OCA/server-ux/files/5229305/purchase_substate.zip)

nmoturi:
Thanks @kittiu for the info. I will test through the process.

@nmoturi

nmoturi:
Hi Kitti,

I was testing the flow of sub-state functionality for PO. I was expecting the substate will active when the tier validation is approved. However, the PO starts with the default substate which is set in approved definition in the tier validation.

Somehow the substate functionality is not a smooth flow, unless the visibility of these substate only available for tier validation as per the definition instead of showing all the active substate in a default document.

image

We can discuss through if needed further explanation.

Thanks,
Naran

PR #207, opened 2020-09-15
[13.0][IMP] base_tier_validation, auto validate & action

kittiu:
Supersede -- https://github.com/OCA/server-ux/pull/196

This PR add 2 more option to definition:
1. Trigger server action after validate a tier
2. Auto validate tier

After validate the last step of PO, do "Confirm" the PO too.
If user is on vacation and the amount is less than 1000, auto validate the PO.
etc.

### DIFF
--- a/base_tier_validation/data/cron_data.xml
+++ b/base_tier_validation/data/cron_data.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <record id="ir_cron_auto_tier_validation" model="ir.cron">
+        <field name="name">Automatic Tier Validation</field>
+        <field name="interval_number">1</field>
+        <field name="interval_type">hours</field>
+        <field name="numbercall">-1</field>
+        <field
+            name="nextcall"
+            eval="(datetime.now() + timedelta(minutes=5)).strftime('%Y-%m-%d %H:%M:%S')"
+        />
+        <field name="doall" eval="False" />
+        <field name="model_id" ref="model_tier_definition" />
+        <field name="code">model._cron_auto_tier_validation()</field>
+        <field name="state">code</field>
+    </record>
+</odoo>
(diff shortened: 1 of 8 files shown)


kittiu:
@LoisRForgeFlow, ready for review.

kittiu:
@LoisRForgeFlow you may want to see this too. https://github.com/OCA/server-ux/pull/197

base_tier_validation/data/cron_data.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <record id="ir_cron_auto_tier_validation" model="ir.cron">
+        <field name="name">Automatic Tier Validation</field>
+        <field name="interval_number">1</field>
```
LoisRForgeFlow, line 5: This is a very high frequency, I would make it once an hour by default. You should not try to make something the closest to real-time as possible by using a very high frequency cron job, to do that is is better to use queue jobs.

LoisRForgeFlow (commented):
Maybe you can include new option in new tab or tabas in the form view of the tier definition. I think the view needs a little bit of reorganization, don't you think?

base_tier_validation/data/cron_data.xml
kittiu: Done

kittiu:
> Maybe you can include new option in new tab or tabas in the form view of the tier definition. I think the view needs a little bit of reorganization, don't you think?

Sure!

image

image

WDYT?

nmoturi:
@LoisRForgeFlow - Will this be merged soon..?

Thanks,
Naran

kittiu:
@LoisRForgeFlow @HviorForgeFlow can you help moving this on Code Sprint?

kittiu:
Closed as move to V14
https://github.com/OCA/server-ux/pull/228

PR #211, opened 2020-10-02
[12.0] Backport/forward port mass editing improvements from 13.0/10.0

simahawk:
From v13: refactor based on mass_operation_abstract https://github.com/OCA/server-ux/commit/8b7156422f8004d15bbdf53257ef8989c275ad4e

From this PR from v10 https://github.com/OCA/server-tools/pull/1856

* improve mig scripts
* ease override of field options

Preserved (although rewritten) improvement from v12 for security group https://github.com/OCA/server-ux/commit/620a58e5e57f6df3dff4cc0bcaa848b34a0393a9
This one should be fwd ported to v13 once this is over.

PR #213, opened 2020-10-07
[FIX] mass_editing migration V12 

legalsylvain:
avoid to run migration script, if mass_editing table exists, that case can occure if the V10 migration has been executed.

Introduced recently by #211 

CC : @simahawk , @lmignon, @Yajo, @HaraldPanten

### DIFF
--- a/mass_editing/migrations/12.0.2.0.1/post-migration.py
+++ b/mass_editing/migrations/12.0.2.0.1/post-migration.py
@@ -3,10 +3,11 @@
 # @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
 # Copyright 2017 ACSONE SA/NV (<http://acsone.eu>)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
 
 
 def migrate(cr, version):
-    if not version:
+    if not version or openupgrade.table_exists(cr, 'mass_editing'):
         return
     # Move field_ids to mass_editing_line
     cr.execute(
--- a/mass_editing/migrations/12.0.2.0.1/pre-migration.py
+++ b/mass_editing/migrations/12.0.2.0.1/pre-migration.py
@@ -7,7 +7,7 @@ from openupgradelib import openupgrade
 
 
 def migrate(cr, installed_version):
-    if not installed_version:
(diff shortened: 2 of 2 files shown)


simahawk (approved):
tnx @legalsylvain :)

simahawk:
@lmignon in this case we should use nobump IMO but we won't get a new release on pypi :thinking:

legalsylvain:
I missed an arg. do not merge now.

legalsylvain:
1) Now fixed.

2) 

> @lmignon in this case we should use nobump IMO but we won't get a new release on pypi thinking

Why use nobump ? bumbing in patch should be OK. (migration script will be executed if version is > folder name. don't you think ?

simahawk:
@legalsylvain hmm you are right. But then, if you use 2.0.0 you might have issues -> then you pin 2.0.1 -> you don't get any upgrade step.
Wouldn't be safer to rename the folder to 2.0.1? If you've already installed 2.0.0 -> no issue tnx to your fix. If not, you'll get the migration done even if you start from 2.0.1. Am I wrong?

lmignon (changes requested):
Thank you  for the fix @legalsylvain IMO we should remove the migrations script for 12.0.2.0.0 and create a new one for 12.0.2.0.1. In this way we are sure that the wrong migration script is no more used and we can bump the version to properly release this patch (no need to manually bump the version into the __manifest__)

legalsylvain:
> But then, if you use 2.0.0 you might have issues -> then you pin 2.0.1 -> you don't get any upgrade step.

if I use 2.0.0, 
Case A) 
- i have the issue, so the upgrade fail.
- when upgrading to 2.0.1, the upgrade will so pass.
-> OK

Case B) 
- i don't have issue.
- when upgrading to 2.0.1, the upgrade will so be skipped.
-> OK


So I think that it is useless to rename this folder. (Or maybe I misunderstood something). But no problem to rename the folder ! ;-)

lmignon:
@legalsylvain I prefer to make a proper release and link the change to the release.

StefanRijnhart (approved):
Thanks, I was hit by this as well during a nightly migration.

lmignon (approved):
Thank you @legalsylvain !

lmignon:
I propose to fast track this one once everything is :green_heart:

PR #214, opened 2020-10-07
[FIX] mass_editing: Add company field in wizard to avoid error for fields with attribute check_company

sergio-teruel:
cc @Tecnativa TT26036
In v13.0 any field can be have the attribute "check_company" that adds an extra domain in field view so we need the company_id field in mass editing wizard view.
As example see location_id from stock.location model.
The extra domain is added here:
https://github.com/odoo/odoo/blob/b9201ebdd65eaabab9257cf97c58410681783fa6/odoo/fields.py#L2350

ping @carlosdauden @Tardo

### DIFF
--- a/mass_editing/wizard/mass_editing_wizard.py
+++ b/mass_editing/wizard/mass_editing_wizard.py
@@ -3,13 +3,21 @@
 
 from lxml import etree
 
-from odoo import _, api, models
+from odoo import _, api, fields, models
 
 
 class MassEditingWizard(models.TransientModel):
     _name = "mass.editing.wizard"
     _inherit = "mass.operation.wizard.mixin"
     _description = "Wizard for mass edition"
+
+    # need for fields with attribute check_company
+    # See: odoo/fields.py#L2350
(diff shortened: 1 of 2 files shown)


victoralmau (approved):

pedrobaeza:
Not sure this is the best patch, as you can have several simultaneous companies viewed at the same time and do the mass editing for records of several companies, having an error. I think it's better to intercept fields_view_get/fields_get and remove the check_company attribute if exists.

Tardo (approved):

sergio-teruel:
@pedrobaeza Change the method to do it.

pedrobaeza:
Please check Travis.

pedrobaeza:
flake8 errors:
```
mass_editing/wizard/mass_editing_wizard.py:6:1: F401 'odoo.fields' imported but unused
mass_editing/wizard/mass_editing_wizard.py:163:20: F811 redefinition of unused 'fields' from line 6
```

HaraldPanten (approved):

PR #215, opened 2020-10-07
[12.0][MIG+FIX] mass_editing: Complete migration scripts

pedrobaeza:
- Set action_name on existing mass editing records
- Set proper context on window actions that come from mass editings.

@Tecnativa TT26041

### DIFF
--- a/mass_editing/migrations/12.0.2.0.2/post-migration.py
+++ b/mass_editing/migrations/12.0.2.0.2/post-migration.py
@@ -1,0 +1,22 @@
+# Copyright 2020 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    old_pattern = r"$$'mass_editing_object' : ([0-9]*)$$"
+    new_pattern = (
+        r"$$'mass_operation_mixin_id' : \1, "
+        r"'mass_operation_mixin_name' : 'mass.editing', $$"
+    )
+    openupgrade.logged_query(
+        env.cr,
+        r"""UPDATE ir_act_window
+        SET context = regexp_replace(context, %(old_pattern)s, %(new_pattern)s)
+        WHERE context ~ %(old_pattern)s
+        """ % {
+            'old_pattern': old_pattern,
+            'new_pattern': new_pattern,
+        }
+    )
--- a/mass_editing/migrations/12.0.2.0.2/pre-migration.py
+++ b/mass_editing/migrations/12.0.2.0.2/pre-migration.py
@@ -1,0 +1,11 @@
+# Copyright 2020 Vauxoo - Luis González
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if not openupgrade.column_exists(env.cr, 'mass_editing', 'action_name'):
+        openupgrade.copy_columns(
+            env.cr, {'mass_editing': [('name', 'action_name', None)]}
+        )


pedrobaeza:
@legalsylvain @lmignon @simahawk please review this that has screwed up all the existing mass editing records/actions with the change.

Another question: why setting widget=selection by default in many2one records? This is provoking 2 undesired changes over previous version:

- No possible to type for searching the value. This is very annoying for users and remove the possibility to write wildcards (`%`) as well.
- Selection widget applies a limit, so if you have more than 30 (I think) records, you can't select the rest of the values.

pedrobaeza:
You need to merge with at least `/ocabot merge patch`

pedrobaeza:
Ouch, it has launched the merge!, aborting...

lmignon (approved):
Thank you @pedrobaeza for the migration script.

simahawk:
@pedrobaeza tnx for the fix. Can we merge then?

> Another question: why setting widget=selection by default in many2one records? This is provoking 2 undesired changes over previous version:
> 
>     * No possible to type for searching the value. This is very annoying for users and remove the possibility to write wildcards (`%`) as well.
> 
>     * Selection widget applies a limit, so if you have more than 30 (I think) records, you can't select the rest of the values.

I'm with you: better to have the m2o widget IMO.

pedrobaeza:
I can include in this migration script the change of selection to empty (which is the normal m2o one).

pedrobaeza:
and also change that onchange behavior.

pedrobaeza:
@legalsylvain can you please comment about this?

mass_editing/migrations/12.0.2.0.2/post-migration.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2020 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        "UPDATE mass_editing SET action_name = name WHERE action_name IS NULL"
```
luisg123v: AFAIK this has to be done in a pre-migration, before the NOT NULL constraint is applied (as in https://github.com/OCA/server-ux/pull/221).

luisg123v (changes requested):

mass_editing/migrations/12.0.2.0.2/post-migration.py
pedrobaeza: That's not possible, as the column still doesn't exist. There's no problem though: next time you update, you will get the constraint.

luisg123v: You may copy it, as done in the mentioned PR.

mass_editing/migrations/12.0.2.0.2/post-migration.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2020 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
```
nhomar, line 3: This import will fail because this module does not depend of openupgradelib (for example in out servers such library is not installed).

Is this intentional, @pedrobaeza ?

luisg123v: It won't fail, because it's on [the requirements.txt](https://github.com/vauxoo/server-ux/blob/12.0/requirements.txt#L3)

lmignon: @luisg123v This dependecy should be declared into the __manifest__.py as external python dependency.
```python
 "external_dependencies": {"python": ["openupgradelib"]},
```

pedrobaeza: @nhomar first advice: install openupgradelib library everywhere. It's used in a lot of places, and not always for migration scripts, as it has interesting methods for easing certain operations. Second thing: it's the same bad failing due to missing library declared in the manifest than when executing the migration scripts. Doing this script without openupgradelib will lose featues (no logging) or uglier code, so it's reasonable to expect to have it in any deployment of OCA modules. As @luisg123v says, it's on requirements.txt of this repo, and you deploy using it.

pedrobaeza: We have got this debate several times, but if you will block this by this reason, I can add the `external_dependencies` key. I don't add it because someone installing the module for the first time doesn't need the library.

pedrobaeza: Note that openupgradelib "hidden dependency" was already introduced previously. See #213

lmignon: @pedrobaeza AFAIK, it's just a comment....  IMO it remains important to always declare external dependencies but I sometime also forget to do it or to pay attention to this point in my reviews.

pedrobaeza: But as said, new users don't need such dependency, so I prefer to keep it a it is now, but if you want to add it, I will do it.

mass_editing/migrations/12.0.2.0.2/post-migration.py
pedrobaeza: OK, copied in pre-migration as in your PR.

pedrobaeza:
@lmignon @nhomar what do you think about the widget `selection` for many2one commented here: https://github.com/OCA/server-ux/pull/215#issuecomment-704870051 ?

mass_editing/migrations/12.0.2.0.2/post-migration.py
lmignon: @pedrobaeza It's ok for me

pedrobaeza: OK, thanks.

lmignon:
@pedrobaeza I d'ont know why the widget has been changed... But if there is no good reason and introduce bad UX, you can restore the precious one.

mass_editing/migrations/12.0.2.0.2/post-migration.py
luisg123v: @pedrobaeza you seem to have deleted it instead. Could you check, please?

legalsylvain (approved):

mass_editing/migrations/12.0.2.0.2/post-migration.py
pedrobaeza: Sorry, I didn't do `git add` of the new file. You can check now.

pedrobaeza:
@legalsylvain can you give your opinion on the selection widget in many2one?

pedrobaeza:
@nhomar @luisg123v do you agree to merge this then?

moylop260 (approved):

pedrobaeza:
I proceed with the merge then:

/ocabot merge nobump

PR #218, opened 2020-10-10
[12.0][UPD] Brainbean Apps => CorporateHub

alexey-pelykh:


alexey-pelykh:
@pedrobaeza fast-track if possible

PR #219, opened 2020-10-12
[13.0][ADD] base_overdue_reminder

Saran440:
This module migrated / split from https://github.com/OCA/credit-control/pull/69 to base of other module in v13
Provide abstract models to manage overdue reminder.
overdue reminder to be applied on different models (invoices, expense advance, ...).
Selection_001

example:
--------

* for the expense advance overdue date and not clearing.
* We can also send mail or print letter when the overdue date.

**Note:** To be able to use this module in a new model you will need some
development. It is not useful for itself. You can see an example of implementation
in the 'hr_expense_advance_overdue_reminder' module. (https://github.com/OCA/hr-expense/pull/12).

cc @alexis-via


kittiu:
I support the idea of making this useful feature as base module for many other modules to extend.
@alexis-via at least, this can be done from v13 onward and make remark in readme? WDYT?

PR #220, opened 2020-10-16
[13.0][ADD] base_tier_validation_forward

kittiu:
Make a separated module from https://github.com/OCA/server-ux/pull/197


kittiu:
Closed as moved to work on v14

LoisRForgeFlow:
Thanks! could you check travis?

base_tier_validation_escalate/tests/test_tier_validation.py
```diff
@@ -0,0 +1,116 @@
+# Copyright 2018 ForgeFlow S.L.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo.tests import Form, common
+
+from .common import setup_test_model, teardown_test_model
```
LoisRForgeFlow: I would change this to `from odoo.addons.base_tier_validation.tests.common import setup_test_model, teardown_test_model` and remove the common file in this module.

kittiu: Thank you and fixed.

base_tier_validation_escalate/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "website": "https://github.com/OCA/server-ux",
+    "author": "Ecosoft, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": ["base_tier_validation"],
+    "data": [
+        "data/mail_data.xml",
+        "views/tier_definition_view.xml",
+        "wizard/escalate_wizard_view.xml",
+    ],
+    "development_status": "alpha",
```
LoisRForgeFlow: This should start with upper case: `Alpha`.

LoisRForgeFlow (commented):
Could you check last complains from pre-commit? (maybe you need to rebase and update it)

yostashiro:
@kittiu Thanks for the PR 😄 

Just some random comments. I was thinking that in real business situations, there could be more cases to 'delegate' the decisions than to 'escalate'. Wasn't sure the module is strictly intended for escalations, but thought the name could be reconsidered otherwise.

In any case, I thought it would be nice, for better security, to have an option of defining a group of users who you could escalate/delegate a validation to, although it may be outside of the scope of this module at this point.

kittiu:
> @kittiu Thanks for the PR 
> 
> Just some random comments. I was thinking that in real business situations, there could be more cases to 'delegate' the decisions than to 'escalate'. Wasn't sure the module is strictly intended for escalations, but thought the name could be reconsidered otherwise.
> 
> In any case, I thought it would be nice, for better security, to have an option of defining a group of users who you could escalate/delegate a validation to, although it may be outside of the scope of this module at this point.

Thank you Yoshi. In fact, at first I though of using "Forward", but some other member think it should named "Escalate". And yes, it allow to pick anyone, either higher or lower people in company structure.

I think it is about terminology that can communicate. From Biz people, is the word "Delegate" normally used?

As to restrict user for security, I think it also varied (I might want people in same dept than group). I will make sure it can be hooked easily then.

base_tier_validation_escalate/models/tier_review.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @api.depends(
+        "definition_id.name",
+        "definition_id.review_type",
+        "definition_id.reviewer_id",
+        "definition_id.reviewer_group_id",
+        "definition_id.has_comment",
+        "definition_id.approve_sequence",
+    )
+    def _compute_definition_data(self):
```
HviorForgeFlow: I see this is overwriting related behavior from Tier Validation Base module, @LoisRForgeFlow what do you think to move from related to compute + stored in base module? @kittiu why did you apply this fields behavior change.

kittiu: @HviorForgeFlow  So far i will need some persistant fields to create a new adhoc tier (escalate). But with the existing it does not allowed me to do (now everything tie to definition). Also dont like it too, but see no other way yet. WDYT?

LoisRForgeFlow: It seems not ideal indeed. However, if this is unavoidable (I would let @kittiu to decide it), I would only ask to add some comments in the code indicating that most of this is overriding base module and maybe also another comment in the readme highlighting that the data model is changed when installing this module.

kittiu: Thanks @LoisRForgeFlow also what i will try to do is to make sure model is back to normal when unistalled.
I shall go ahead.

yostashiro:
> In fact, at first I though of using "Forward", but some other member think it should named "Escalate".

@kittiu To me 'forward' actually sounds more generic, and therefore is better. Not a strong preference though.

kittiu:
Changed name to `base_tier_validation_forward` simpler is better :)

kittiu:
Can test with purchase_tier_validation_forward - https://github.com/OCA/purchase-workflow/pull/1011

base_tier_validation_escalate/models/tier_review.py
kittiu: Set model back to original.
https://github.com/ecosoft-odoo/server-ux/blob/13.0-add-base_tier_validation_escalate/base_tier_validation_forward/hooks.py

Note: the forwarded tier will be set back to status approved, and the sequence will be round down to its integer.

base_tier_validation_forward/hooks.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2020 Ecosoft Co., Ltd. (http://ecosoft.co.th)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+def uninstall_hook(cr, registry):
+    """ Revert table tier_review back to original before this module """
```
LoisRForgeFlow, line 6: :+1:

LoisRForgeFlow (approved):
LGTM

LoisRForgeFlow:
@yostashiro @HviorForgeFlow could you update your review?

kittiu:
cc @ps-tubtim @newtratip

HviorForgeFlow (approved):
Code review :+1:

PR #221, opened 2020-10-18
[FIX] mass_editing: Unsatisfied NOT NULL constraint on action name

luisg123v:
Since this module was refactored and it now depends on
`mass_operation_abstract`, a new required field `action_name` is
introduced, that might not be already filled. That causes the SQL
constraint to fail to be applied, showing the following message:

    odoo.modules.loading: Table 'mass_editing': column 'action_name': unable to set constraint NOT NULL
    column "action_name" contains null values

To solve the above, the new field is initialized with the same value as
name, in case it's not filled yet.

### DIFF
--- a/mass_editing/migrations/12.0.2.0.1/pre-migration.py
+++ b/mass_editing/migrations/12.0.2.0.1/pre-migration.py
@@ -11,3 +11,6 @@ def migrate(cr, installed_version):
         return
 
     openupgrade.rename_tables(cr, [('mass_object', 'mass_editing')])
+    if not openupgrade.column_exists(cr, 'mass_editing', 'action_name'):
+        copy_vals = {'mass_editing': [('name', 'action_name', None)]}
+        openupgrade.copy_columns(cr, copy_vals)


luisg123v:
Hi @ simahawk & @legalsylvain,

Could you review, please?

CC @moylop260

luisg123v:
Hi @simahawk & @legalsylvain,

Could you review, please?

CC @moylop260

legalsylvain (approved):
thanks.

pedrobaeza:
I proposed a week ago a fix for this and also did other thing. @legalsylvain I pinged you several times.

#215

simahawk (approved):
makes sense and seems a bit better than #215

pedrobaeza:
@simahawk (and the rest). This is not better, as:

- It's incomplete. It doesn't fix the populated action windows.
- It's in the wrong place. Those that have updated already to 12.0.2.0.1 won't have this patch.

moylop260:
IMHO https://github.com/OCA/server-ux/pull/215 is more complete

@luisg123v 
Could you check it, please?

luisg123v:
@JMPinyol-PlanetaTIC Initially https://github.com/OCA/server-ux/pull/215 didn't fix the issue addressed here, that was fixing instances that were not migrated yet due to this message. But now it does, so closing this.

luisg123v:
@moylop260  Initially https://github.com/OCA/server-ux/pull/215 didn't fix the issue addressed here, that was fixing instances that were not migrated yet due to this message. But now it does, so closing this.

PR #222, opened 2020-10-22
[14.0][FIX] date_range: typo in date_start field name

astirpe:
There's a typo ("`strint`" instead of "`string`") in the definition of field `date_start`.

This generated following error in runbot:

`2020-10-22 09:48:42,986 160 WARNING openerp_test odoo.fields: Field date.range.generator.date_start: unknown parameter 'strint', if this is an actual parameter you may want to override the method _valid_field_parameter on the relevant model in order to allow it`

### DIFF
--- a/date_range/wizard/date_range_generator.py
+++ b/date_range/wizard/date_range_generator.py
@@ -17,7 +17,7 @@ class DateRangeGenerator(models.TransientModel):
         return self.env.company
 
     name_prefix = fields.Char("Range name prefix", required=True)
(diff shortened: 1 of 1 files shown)


pedrobaeza:
But better to fix the typo and keep the current label instead of removing it.

astirpe:
What you call the "current label" if actually never applied and thus never translated.
Better to keep the current behavior so that the current translations are not affected.

pedrobaeza (approved):
You are right, Andrea!

lmignon (approved):

PR #223, opened 2020-10-22
[FIX] mass_operation_abstract : return True to allow call from xml-rpc

legalsylvain:
Trivial Patch.

add return True in two function to allow call from xml-rpc.


simahawk (approved):

simahawk:
to be fwd ported right?

legalsylvain:
Indeed.
Could you fast merge this one ?

PR #225, opened 2020-10-23
[14.0][MIG] mass operation abstract

michotm:
Migration of mass operation abstract to be able to migrate module mass_editing to 14.0


ivantodorovich:
Hi @michotm !

Thanks for your contribution.

There have been some discussions about the need of this module, and apparently the idea is to deprecate it in favour of `ir.actions.server`.

Here are the mentioned discussions:

- https://github.com/OCA/reporting-engine/pull/457#discussion_r527568225
- https://github.com/OCA/server-ux/pull/97#pullrequestreview-317654463

AFAIK this module is only depended by `mass_editing`, and based on those discussions it should be refactored and thus we can remove this module

ping @legalsylvain @hbrunn

CasVissers-360ERP:
Can be closed #255 is merged without the need of this module.

pedrobaeza:
Maybe it's used by other module than `mass_editing`, but Travis is red, so I don't think this one will flourish, so closing.

PR #226, opened 2020-10-26
[14.0][MIG] base_tier_validation

kittiu:
Standard Migration, ready for review.


LoisRForgeFlow:
@kittiu Thanks for the work. Could you include https://github.com/OCA/server-ux/pull/235 and https://github.com/OCA/server-ux/pull/233 commits here?

base_tier_validation/security/ir.model.access.csv
```diff
@@ -2,3 +2,4 @@ id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 access_tier_review,access.tier.review,model_tier_review,,1,1,1,1
 access_tier_definition_all,tier.definition.all,model_tier_definition,,1,0,0,0
 access_tier_definition_settings,tier.definition.settings,model_tier_definition,base.group_system,1,1,1,1
+access_comment_wizard,access.comment.wizard,model_comment_wizard,,1,1,1,1
```
LoisRForgeFlow, line 5: Interesting, do wizards require access rights records now in v14?

kittiu: Yes. like this,
https://github.com/odoo/odoo/blob/14.0/addons/account/security/ir.model.access.csv#L113

kittiu:
> @kittiu Thanks for the work. Could you include #235 and #233 commits here?

Done. Is the order of commits correct this way?

kittiu:
@LoisRForgeFlow I propose to include following to this PR. So, other following extension can be included in Extra Options tab.
https://github.com/OCA/server-ux/pull/207#issuecomment-699614958
WDYT?

LoisRForgeFlow:
I am doubtful on this, I am not really able to imagine very common usecases for those features, maybe if you provide an example of both, it would help me decide.

kittiu:
> I am doubtful on this, I am not really able to imagine very common usecases for those features, maybe if you provide an example of both, it would help me decide.

It is init from your coment below.

> Maybe you can include new option in new tab or tabas in the form view of the tier definition. I think the view needs a little bit of reorganization, don't you think?

So I try to propose the new tab so extension like "forward" can be there.

LoisRForgeFlow:
My doubts are about including the features to auto-complete and to execute server actions in the base module, if you were only talking about reorganizing the view and prepare the space for new settings, that's fine. Otherwise, if you want to include such feature in the base module I would need a better understanding.

kittiu:
Sorry to have confused you. Please ignore the server action in that PR, nothing to added in base module. I just want to add tab in view, like this.

image

Let's confuse no more, I better propose the UI reorganize in separated PR after this one merged.

LoisRForgeFlow:
Got it! No, I think it is perfect to take this occasion to do it! so, please, go ahead.

kittiu:
> Got it! No, I think it is perfect to take this occasion to do it! so, please, go ahead.

Tabs added.

LoisRForgeFlow (approved):
Tested locally, LGTM :+1:

etobella (approved):
LGTM Code Review :+1:

PR #228, opened 2020-10-29
[14.0][ADD] base_tier_validation_server_action

kittiu:
Add options to,

* Trigger server action after validate a tier
* Auto validate tier by cron job

After validate the last step of PO, do "Confirm" the PO too.
If user is on vacation and the amount is less than 1000, auto validate the PO.
etc.

Depends on:

- [x] base_tier_validation https://github.com/OCA/server-ux/pull/226


LoisRForgeFlow:
You can now rebase this, as the dependency is merged.

kittiu:
Rebased, ready for review.

base_tier_validation_server_action/tests/test_tier_validation.py
```diff
@@ -0,0 +1,177 @@
+# Copyright 2020 Ecosoft (http://ecosoft.co.th)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo.tests import common
+from odoo.tests.common import tagged
+
+from .common import setup_test_model, teardown_test_model
+from .tier_validation_tester import TierValidationTester
```
LoisRForgeFlow: Please do not duplicate this files, take them from `odoo.addons.base_tier_validation.tests...`

kittiu: Thank you @LoisRForgeFlow 
I can do for .common, but not yet for the test model, as I need more feature. (don't know how to inherit it, by code).
https://github.com/ecosoft-odoo/server-ux/blob/14.0-add-base_tier_validation_action/base_tier_validation_server_action/tests/tier_validation_tester.py#L22-L28

LoisRForgeFlow:
@kittiu Could you rebase this? I am afraid this might have been affected by the latest changes in base module tests.

base_tier_validation_server_action/tests/test_tier_validation.py
LoisRForgeFlow: I see, I don't know either, it is likely possible but it is also likely not worth it :smile:

LoisRForgeFlow (approved):
LGTM

PrapassornS (approved):
Functional Tested LGTM.

base_tier_validation_server_action/models/tier_review.py
```diff
@@ -0,0 +1,16 @@
...
+
+from odoo import api, models
+
+
+class TierReview(models.Model):
+    _inherit = "tier.review"
+
+    @api.constrains("status")
+    def _trigger_server_action(self):
+        for rec in self.filtered(lambda l: l.status == "approved"):
```
etobella, line 12: You are doing two loops. It is somehow better to use:
```
for rec in self:
    if rec.status != "approved":
        continue
    # continue here....
```

Non blocking as the improvement is not so big, but I think it is good to know :smile:

base_tier_validation_server_action/models/tier_definition.py
```diff
@@ -0,0 +1,61 @@
...
+        "- If > 1 user matched as reviewer, do not auto validate",
+    )
+    auto_validate_domain = fields.Char()
+
+    def _evaluate_review(self, review):
+        domain = [("id", "=", review.res_id)]
+        auto_domain = review.definition_id.auto_validate_domain
+        if auto_domain:
+            domain += literal_eval(auto_domain)
+        return self.env[review.model].search(domain)
```
etobella: You can use `filtered_domain`

base_tier_validation_server_action/tests/tier_validation_tester.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2020 Ecosoft (http://ecosoft.co.th)
```
etobella, line 1: I think you can use the same model generated on the test by `base_tier_validation`, so we can avoid this file

https://github.com/OCA/server-ux/blob/14.0/base_tier_validation/tests/tier_validation_tester.py

I don't know if it is possible

etobella (approved):
Just some nonblocking comments

base_tier_validation_server_action/models/tier_review.py
kittiu: For this one, I can't see how this is two loop. I just want the record with status changed to 'approved' first.
Not sure what is the difference here. Appreciate your explain (so I will know :)

etobella: The filter is making a for for each id: so, at the end the original code is doing something similar to the following code:
```
to_check = self.browse()
for rec in self:
    if rec.status == 'approved':
        to_check |= rec
for rec in to_check:
    # Your code
```

kittiu: OK I got Thank you Eric.
But may I keep this one as is for readability if you are ok. We expect just few records in self.

base_tier_validation_server_action/models/tier_definition.py
kittiu: This one, I can just make it a little better like this, if this is what you think?
```
    def _evaluate_review(self, review):
        doc = self.env[review.model].browse(review.res_id)
        domain = review.definition_id.auto_validate_domain or "[]"
        return doc.filtered_domain(literal_eval(domain))
```

base_tier_validation_server_action/tests/tier_validation_tester.py
kittiu: It need one more field `test_bool` as exception filed `_get_under_validation_exceptions()`. I am not sure we can inherit it further.

Or otherwise, we should have that field since `base_tier_validation`. WDYT?

etobella: If you need that field, keep the file here. I didn't know if it was required :wink:

base_tier_validation_server_action/models/tier_definition.py
etobella: great

base_tier_validation_server_action/models/tier_review.py
etobella: Good for me, it was just a comment and nonblocking :wink:

kittiu:
@etobella thank you for all your reviews, it was good as always. :+1:

LoisRForgeFlow:
Let's merge this, it has been reviewed and all comments attended, plus it is an alphan module.

/ocabot merge nobump

PR #229, opened 2020-10-30
[MIG] base_technical_features: Migration to 14.0

p-tombez:


base_technical_features/models/res_users.py
```diff
@@ -51,5 +51,5 @@ def _inverse_technical_features(self):
 
     def __init__(self, pool, cr):
         super().__init__(pool, cr)
-        self.SELF_READABLE_FIELDS += ["technical_features", "show_technical_features"]
-        self.SELF_WRITEABLE_FIELDS.append("technical_features")
+        type(self).SELF_READABLE_FIELDS += ["technical_features", "show_technical_features"]
+        type(self).SELF_WRITEABLE_FIELDS.append("technical_features")
```
guewen, line 58: maybe using properties could be clearer?

```
@property
def SELF_READABLE_FIELDS(self):
    return super().SELF_READABLE_FIELDS + ["technical_features", "show_technical_features"]

@property
def SELF_WRITEABLE_FIELDS(self):
    return super().SELF_WRITEABLE_FIELDS + ["technical_features"]

```

p-tombez: That would indeed be clearer that it overrides a property.
I'm just wondering, as this notation is used everywhere when overriding these exact same properties, would it be wise to use a different notation ?

guewen: > as this notation is used everywhere when overriding these exact same properties,

Oh, then it makes sense to use the same as the other places.
Thanks for the clarification!

guewen (approved):

tonygalmiche:
The screenshot in README is not the right one. Now it's like that : 
image

I just tested with runbot, but it didn't activate developer mode

pedrobaeza:
@p-tombez can you please check @tonygalmiche's comment about this not working?

pedrobaeza:
@tonygalmiche this module doesn't activate developer mode, but UI elements associated with technical group. Maybe you were expecting something different?

tonygalmiche (approved):
@pedrobaeza indeed, it is the README file that made me think of the developer mode
`In Odoo 9.0 and later, the debug mode grants every employee user access to the technical features`
So there is no problem. thank you for the answer
LGTM : Functional review only with runbot

pedrobaeza:
OK, if you think that it needs to be rephrased, just propose an alternative redaction.

pedrobaeza:
Let's merge it then as it's working:

/ocabot merge nobump

CasVissers-360ERP:
@pedrobaeza Merging seems stuck?

pedrobaeza:
It seems so:

/ocabot merge nobump

LoisRForgeFlow:
@p-tombez Pre-commit [is complaining](https://travis-ci.com/github/OCA/server-ux/jobs/443594351). Could you rebase and run pre-commit?

p-tombez:
```
-    "website": "https://github.com/oca/server-ux",
```
Really ? Well, it's fixed now

PR #230, opened 2020-11-02
[12.0][IMP] base_tier_validation: state check

yostashiro:
This commit splits the state condition checking
part into another method, allowing inheriting
modules to override the conditions as needed.

This can be useful when the state field of the
model is a compute field (e.g. stock.picking) and
therefore cannot be relied upon in the write
method.

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -149,8 +149,7 @@ class TierValidation(models.AbstractModel):
     @api.multi
     def write(self, vals):
         for rec in self:
-            if (getattr(rec, self._state_field) in self._state_from and
-                    vals.get(self._state_field) in self._state_to):
+            if rec._check_state_conditions(vals):
                 if rec.need_validation:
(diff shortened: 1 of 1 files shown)


TimLai125 (approved):
LGTM

base_tier_validation/models/tier_validation.py
```diff
@@ -172,6 +171,10 @@ def write(self, vals):
             self.mapped('review_ids').unlink()
         return super(TierValidation, self).write(vals)
 
+    def _check_state_conditions(self, vals):
```
kittiu, line 174:
```suggestion
    def _check_state_conditions(self, vals):
        self.ensure_one()
```
Just to make sure it is not called in multi.

kittiu:
I am ok with this proposal, more flexible and no harm.

kittiu (approved):

LoisRForgeFlow (approved):
Sounds reasonable. LGTM :+1:

PR #237, opened 2020-11-17
[12.0][MIG] mass_editing: Add apply domain option in lines

victoralmau:
Add apply domain option in lines to fix some many2one fields with default domain (`company_id` for example)

Please @pedrobaeza  review it.

@Tecnativa TT26786

### DIFF
--- a/mass_editing/models/mass_editing.py
+++ b/mass_editing/models/mass_editing.py
@@ -1,7 +1,7 @@
 # Copyright 2016 Serpent Consulting Services Pvt. Ltd. (support@serpentcs.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class MassEditing(models.Model):
@@ -14,3 +14,12 @@ class MassEditing(models.Model):
     line_ids = fields.One2many(
         comodel_name="mass.editing.line", inverse_name="mass_editing_id"
     )
+    apply_domain_in_lines = fields.Boolean(
+        string="Apply domain in lines",
+        compute="_compute_apply_domain_in_lines"
+    )
+
+    @api.depends("line_ids")
+    def _compute_apply_domain_in_lines(self):
+        for record in self:
+            record.apply_domain_in_lines = any(record.line_ids.mapped("apply_domain"))
--- a/mass_editing/models/mass_editing_line.py
+++ b/mass_editing/models/mass_editing_line.py
@@ -38,15 +38,19 @@ class MassEditingLine(models.Model):
         "'many2many_tags', 'selection'",
     )
 
+    apply_domain = fields.Boolean(
+        default=False,
+        string="Apply domain",
+        help="Apply default domain related to field"
+    )
+
     @api.depends("field_id")
     def _compute_widget_option(self):
         # this function propose selection, depending on the field
         for line in self.filtered("field_id"):
             field = line.field_id
             line.widget_option = False
-            if field.ttype == "many2one":
-                line.widget_option = "selection"
-            elif field.ttype == "many2many":
+            if field.ttype == "many2many":
                 line.widget_option = "many2many_tags"
             elif field.ttype == "Binary":
                 if "image" in field.name or "logo" in field.name:
--- a/mass_editing/models/mass_editing_line.py
+++ b/mass_editing/models/mass_editing_line.py
@@ -30,12 +30,11 @@ class MassEditingLine(models.Model):
 
     widget_option = fields.Char(
         string="Widget Option",
-        compute="_compute_widget_option",
         store=True,
         readonly=False,
         help="Add widget text that will be used"
         " to display the field in the wizard. Example :\n"
-        "'many2many_tags', 'selection'",
+        "'many2many_tags', 'selection'"
     )
 
     apply_domain = fields.Boolean(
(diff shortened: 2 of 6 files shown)


mass_editing/wizard/mass_editing_wizard.py
```diff
@@ -76,6 +76,9 @@ def fields_view_get(
             # Field part
             field = line.field_id
             field_info = fields_info[field.name]
+            if not line.apply_domain and 'domain' in field_info:
+                field_info['domain'] = ''
+
```
pedrobaeza: Don't add empty lines inside a method

pedrobaeza:
Please include the infamous widget selection removal. And I think that applying the domain, the fields in that domain should be automatically included in the view.

mass_editing/models/mass_editing_line.py
```diff
@@ -38,6 +38,12 @@ class MassEditingLine(models.Model):
         "'many2many_tags', 'selection'",
     )
 
+    apply_domain = fields.Boolean(
+        default=True,
```
pedrobaeza: Please, we don't want to fail by default

```suggestion
        default=False,
```

victoralmau:
> Please include the infamous widget selection removal. And I think that applying the domain, the fields in that domain should be automatically included in the view.

Added widget selection removal. 
IMO isn't necessary to auto-add field included in field domains because not exist apply domain field.

mass_editing/models/mass_editing_line.py
```diff
@@ -38,16 +38,8 @@ class MassEditingLine(models.Model):
...
 
-    @api.depends("field_id")
-    def _compute_widget_option(self):
-        # this function propose selection, depending on the field
-        for line in self.filtered("field_id"):
-            field = line.field_id
-            line.widget_option = False
-            if field.ttype == "many2one":
-                line.widget_option = "selection"
-            elif field.ttype == "many2many":
```
pedrobaeza, line 49: The many2many_tags and the image one are useful, but not the selection one

pedrobaeza:
I still see that fields of the domain must be added to the view if not already in it. Please check it.

mass_editing/models/mass_editing_line.py
pedrobaeza: I think this is bad conceived, as this shouldn't be a computed field, but a stored one with "defaults". And please do it in a separate commit, as they are different changesets from the domain one.

victoralmau:
> I still see that fields of the domain must be added to the view if not already in it. Please check it.

Ok, we auto-add the fields that are necessary for the domain when `apply_domain` is defined.

mass_editing/models/mass_editing_line.py
victoralmau: Try to define `default` in `widget_option` field but need to defined `field_id` previously, therefore it is now defined with an `@api.onchange`

mass_editing/readme/ROADMAP.rst
```diff
@@ -0,0 +1 @@
+- - Auto add fields that is used in domain related if apply_domain=True
```
pedrobaeza:
```suggestion
- Auto add fields that are used in related domains if apply_domain=True
```

mass_editing/models/mass_editing.py
```diff
@@ -14,3 +14,13 @@ class MassEditing(models.Model):
...
+    apply_domain_in_lines = fields.Boolean(
+        string="Apply domain in lines",
+        compute="_compute_apply_domain_in_lines"
+    )
+
+    @api.depends("line_ids")
+    def _compute_apply_domain_in_lines(self):
+        self.apply_domain_in_lines = False
+        if self.line_ids.filtered(lambda x: x.apply_domain):
+            self.apply_domain_in_lines = True
```
pedrobaeza: Not multi-record aware:
```suggestion
        for record in self:
            record.apply_domain_in_lines = any(record.line_ids.mapped("apply_domain"))
```

mass_editing/views/mass_editing_view.xml
```diff
@@ -30,6 +30,12 @@
         <field name="inherit_id" ref="mass_operation_abstract.view_mass_operation_mixin_form"/>
         <field name="mode">primary</field>
         <field name="arch" type="xml">
+            <field name="domain" position="after">
+                <field name="apply_domain_in_lines" invisible="1" />
+                <div colspan="2" class="text-warning" attrs="{'invisible': [('apply_domain_in_lines','=', False)]}">
+                    <b>WARNING</b>: Take into account that adding a field with a domain, and not including the fields of such domain in this operation defintion, will lead to an error when trying to perform it
```
pedrobaeza:
```suggestion
                    <b>WARNING</b>: Take into account that adding a field with a domain, and not including the fields of such domain in this operation definition, will lead to an error when trying to perform it. Make sure you include them.
```

mass_editing/wizard/mass_editing_wizard.py
```diff
@@ -76,6 +76,8 @@ def fields_view_get(
             # Field part
             field = line.field_id
             field_info = fields_info[field.name]
+            if not line.apply_domain and 'domain' in field_info:
+                field_info['domain'] = ''
```
pedrobaeza: Shouldn't this be  `pop field_info['domain']` or `field_info['domain'] = '[]'`?

pedrobaeza (changes requested):

pedrobaeza (approved):

HaraldPanten (approved):

pedrobaeza (approved):

pedrobaeza:
/ocabot merge minor

Please fw-port it to v13

PR #245, opened 2020-11-30
[14.0][MIG] base_tier_validation_forward

kittiu:
Standard Migration, ready for review.


LoisRForgeFlow:
Is this ready?

kittiu:
> Is this ready?

Yes

etobella (approved):
After looking on the migration code, no major changes.
Hard to test functionally as it needs a module that implements it.

Just a question, why not modifiying the Javascript Widget in order to avoid this issue and avoid glue modules?

kittiu:
> After looking on the migration code, no major changes.
> Hard to test functionally as it needs a module that implements it.
> 
> Just a question, why not modifiying the Javascript Widget in order to avoid this issue and avoid glue modules?

Thanks for the review. There is putchase_tier_validation_forward.
https://github.com/OCA/purchase-workflow/pull/103

I dont know JS. Please feel free to suggest, how it can help.

etobella:
Oh! I see! The problem is that it has not correctly been solved on `base_tier_validation`. You need to modify the view every time. I think that it should have been solved like in `edi` and modify the `fields_view_get` Now it is probably too late for 14.0, but It would be interesting to review it on migration to 15.0

https://github.com/tegin/edi/blob/13.0-imp-edi_account/edi/models/edi_exchange_consumer_mixin.py#L77

@LoisRForgeFlow WDYT?

LoisRForgeFlow:
@etobella I agree, that would be ideal but of course it is out a strict migration so I let @kittiu decide if he wants to put some time to do it.

kittiu:
@LoisRForgeFlow  think it make perfect sense to do, it will save many module development.

Now we have only 2 modules that need be removed. https://odoo-community.org/shop?version=16&search=tier_validation
* purchase_request_tier_validation
* purchase_tier_validation

Both are done by my team, so I have no problem to do a PR to remove it. Let me try couple of days. Will let you guys know how it goes.

Meanwhile, let's make sure no other xxx_tier_validation module gets merged.

etobella:
@kittiu It will probably be impossible to delete this modules, as we need to apply the mixin, but with this change, we will probably not need this glue modules for forwarding and other cases, not even modifying the view to add the buttons and so on.

If you need some help count with me :wink:

kittiu:
@etobella right, I think too far. I though of eliminating the Abstract model (mixin) completely and move all methods and fields (like as x_field...) :laughing: 

I saw some technique like register hooks, which allow us to inherit functions without _inherit.
https://github.com/ecosoft-odoo/server-ux/blob/13.0-add-approvals_tier_validation/approvals_tier_validation/models/tier_approval_category.py#L401

But can add function to a model on the fly, or inherit mixin on the fly without extra modules.  Do you think that is something possible with Odoo?

kittiu:
> But can add function to a model on the fly, or inherit mixin on the fly without extra modules. Do you think that is something possible with Odoo?

Well, I think that will be too big of changes and too difficult even it is possible. I think I will just fix this PR according to @etobella suggest.

Note: but still would like to know if such technique exists. Thanks!

kittiu:
@etobella https://github.com/ecosoft-odoo/server-ux/blob/14.0-mig-base_tier_validation_forward/base_tier_validation_forward/models/tier_validation.py#L13-L42

Thanks for your suggestion, it is a good one :smile: 
If all ok, I will port this back to v13 also.

---

Do you know why field and button is not invisible, despite field view get is so.

image

image

image

etobella:
Because you are not applying the modifiers. When adding fields you need to do https://github.com/OCA/edi/blob/13.0/edi/models/edi_exchange_consumer_mixin.py#L93

However i was talking to do this on base tier validation and use a qweb template. When installing forward you would modify the template and all will be passed. WDYT? I can make a small poc if you need 😄

kittiu:
Oh OK. You can make improvement on base_tier_validation. I will review that and extend it here. Is it ok?

kittiu:
I have tried by adding this template part,
```
<odoo>
    <template id="tier_validation_label_forward" inherit_id="base_tier_validation.tier_validation_label">
        <xpath expr="//button[@name='reject_tier']" position="after">
            <button name="forward_tier"
                string="Forward"
                attrs="{'invisible': [('can_forward', '=', False)]}"
                type="object"
                class="oe_inline oe_button btn-warning"
                icon="fa-arrow-circle-right" />
            <field name="can_forward" invisible="1" />
        </xpath>
    </template>
</odoo>
```

But as we are using this flag also, _tier_validation_manual_config = True, as of now, it means we need the manual template also. (only if it = False, the above template will work).

After trying, I think it is better to be in code only to server both cases. @etobella WDYT?

etobella:
What we should need to do is modifying purchase tier validation to remove the manual so we can proceed properly. WDYT? Also for other OCA modules existent on 14

kittiu:
> What we should need to do is modifying purchase tier validation to remove the manual so we can proceed properly. WDYT? Also for other OCA modules existent on 14

For simplicity, I agree, we shouldn't have `_tier_validation_manual_config option` in base_tier_validation. In v14, only 2 modules now, so it is not too late to fix them now.

* purchase_request_tier_validation
* purchase_tier_validation

etobella:
Well, i think we need to keep it false but force true on all oca tier validation modules

kittiu:
> Well, i think we need to keep it false but force true on all oca tier validation modules

Ok, this will do :+1: just understand what you try to say.

Fixed this one and I will go ahead and fix the 2 existing ones.

kittiu:
Almost ready, but a small test error. @etobella need your help :pray: 

With this `base_tier_validation_forward` module installed, the test in `base_tier_validation` will have this error,
```
2020-12-17 05:14:26,300 1 ERROR devel odoo.addons.base_tier_validation.tests.test_tier_validation: ERROR: TierTierValidation.test_view_automatic
  File "/opt/odoo/auto/addons/base_tier_validation/tests/test_tier_validation.py", line 516, in test_view_automatic
    with Form(self.test_record_2) as f:
  File "/opt/odoo/custom/src/odoo/odoo/tests/common.py", line 1626, in __init__
    self._process_fvg(recordp, fvg)
  File "/opt/odoo/custom/src/odoo/odoo/tests/common.py", line 1709, in _process_fvg
    if level and descr['type'] == 'one2many':
KeyError: 'type'
```

I try to track the problem, and found that this problem is because this line only, https://github.com/ecosoft-odoo/server-ux/blob/14.0-mig-base_tier_validation_forward/base_tier_validation_forward/templates/tier_validation_templates.xml#L7

And so, the only way I can do now is to remove tag 'post_install' from `base_tier_validation` https://github.com/OCA/server-ux/pull/245/commits/8350328dee9d6cc31f4f43995b3b5bfb5c556f6f

etobella:
It is necessary to remove the post_install or do something different. Let me explain it:

When using post_install, we are doing it at the end, but it is not recognizing (it should, but I don't know why it is failing....) a new field imported on the form. So, I think we need to change it. WDYT @LoisRForgeFlow ?

It would be interesting to use a similar test on the new repo in order to show that field is shown correctly.

LoisRForgeFlow:
@kittiu Maybe you can isolate the problematic test and only tag that one as post_install, this way all the other test will continue running unchanged.

With this final change I think we would be ready to go :)

kittiu:
> @kittiu Maybe you can isolate the problematic test and only tag that one as post_install, this way all the other test will continue running unchanged.

@LoisRForgeFlow not sure I understand your idea correctly. The problem test in `base_tier_validation`, you say tag 'post_install' in module , I though we should tag 'at_install' like this. https://github.com/OCA/server-ux/pull/245/commits/c85de130adf5ba43d52fef4e2e8b10508de5ce7e

LoisRForgeFlow:
>  I though we should tag 'at_install' like this. c85de13

@kittiu Yes, but that is affecting all tests in class `TierTierValidation` but only one (`test_view_automatic`) need to be run at the end. My proposal is that you change that in base_tier_validation, move test_view_automatic to a separate test class and apply the post_install only to that test class. This way all the other tests that will remain in `TierTierValidation` will continue being tested during  installation.

base_tier_validation/tests/test_tier_validation.py
```diff
@@ -9,7 +9,7 @@
 from odoo.tests.common import Form, tagged
 
 
-@tagged("post_install", "-at_install")
+@tagged("at_install")
 class TierTierValidation(common.SavepointCase):
```
etobella: So, from this class, you will do 3:
CommonTestTierValidation - Will only contain the setUpClass and the Teardown
TestTierValidation that is post_install,  inherits CommonTestTierValidation and executes all tests except the view
TestTierValidationView that is at_install,  inherits CommonTestTierValidation and executes only the test view

kittiu:
Thank you @LoisRForgeFlow @etobella for your comment. I always learnt somethings from you guys :+1:
Please see if the fix is good enough.

etobella (approved):
gr8!!
LGTM

LoisRForgeFlow (approved):
LGTM, thanks!

PR #233, opened 2020-11-10
[13.0][IMP] base_tier_validation: make module multi-company compatible.

LoisRForgeFlow:
Closed https://github.com/OCA/server-ux/issues/140.

@ForgeFlow

PR #247, opened 2020-12-11
[12.0] base_tier_validation  - Add security rule for multi-company environments

svalaeys:
In a multi-company environment, tier definitions are applied regardless of the company they belong to. Only when the tier definition itself filters on company is there a company-related process. 
This pull request adds security rules similar to elsewhere in Odoo : if a company is defined on the tier definition, then it will only be applied to target objects of that company.

### DIFF
--- a/base_tier_validation/__manifest__.py
+++ b/base_tier_validation/__manifest__.py
@@ -17,6 +17,7 @@
     ],
     "data": [
         "data/mail_data.xml",
+        "security/tier_validation_security.xml",
         "security/ir.model.access.csv",
         "views/tier_definition_view.xml",
         "views/tier_review_view.xml",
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow (approved):
LGTM

for the records the feature is already in v13 (https://github.com/OCA/server-ux/pull/233/files)

pedrobaeza:
Then the XML-ID should be the same for avoiding problems when migrating an instance from v12 to v13.

LoisRForgeFlow:
@svalaeys Could you update it as per Pedro comments?

marcelsavegnago (approved):

etobella:
Travis is failing @svalaeys  Can you check it please?

PR #248, opened 2020-12-12
[14.0][IMP] base_tier_validation: Remove unnecessary modification of views

etobella:
That is a the modification commented on #245 
@kittiu @LoisRForgeFlow


etobella:
This PR will make necessary to modify all existent Modules generating tier.validation models. We can modify the code here in order to only apply the change when necessary

LoisRForgeFlow:
@etobella and I just discossed and agreed on the following approach:

1. Create a new attribute `_tier_validation_manual_config`.
2. Set it by default to `True`in v14 to make migrating to the new approach optional in the tier validation submodules.
3. Add a notification in the roadmap of this change indicating that in v15 the default will be `False`.
4. During v15 migration change the attribute to `False`.

etobella:
Tests also have been improved in order to use odoo_test_helper FakeModule importer, that simplifies the loading of new modules :smile:

etobella:
As I was refactoring tests, I simplified the test en base_tier_validation_formula as it uses the same logic :wink:

etobella:
@LoisRForgeFlow I think all has been implemented, so, I think this is ready to review :wink:

base_tier_validation/models/tier_validation.py
```diff
@@ -372,3 +377,44 @@ def _update_counter(self):
...
+            doc = etree.XML(res["arch"])
+            for node in doc.xpath(self._tier_validation_buttons_xpath):
+                # By default, after the last button of the header
+                str_element = self.env["ir.qweb"]._render(
+                    "base_tier_validation.tier_validation_buttons"
+                )
+                new_node = etree.fromstring(str_element)
+                for new_element in new_node:
+                    node.addnext(new_element)
+            for node in doc.xpath("/form/sheet"):
```
kittiu, line 398: As we can set param _tier_validation_buttons_xpath, as start point of validation buttons,
Should we have param `_tier_validation_label_xpath` also? In case we want to add it to form without sheet.
WDYT?

kittiu: Otherwise, function tested and code reviewed, LGTM :+1:

kittiu (approved):

base_tier_validation/models/tier_validation.py
etobella: Well, I don't think so, as it has a specific behavior, showing a div before the sheet and showing a specific widget after. WDYT?

kittiu: I am ok, if it is intention. User still have option to do it manually.

LoisRForgeFlow (approved):
Tested

Thanks for taking the time! :+1:

PR #251, opened 2020-12-21
[FIX] base_tier_validation: backport fix from 13.0

### LINKED ISSUE #250: [12.0] base_tier_validation: test fails when purchase_tier_validation is in the same repo
This test https://github.com/OCA/server-ux/blob/30e3e035889fc9b7f461c4de70d9f26b78df9402/base_tier_validation/tests/test_tier_validation.py#L155-L160 fails in CI when purchase_tier_validation (or any of those xxx_tier_validation modules) is in the same repo.

yostashiro:
Fixes https://github.com/OCA/server-ux/issues/250.


pedrobaeza:
Fast-tracking this:

/ocabot merge patch

PR #253, opened 2020-12-27
[13.0][FIX] base_tier_validation_formula: Fix cache miss error on iteration records

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/base_tier_validation_formula/models/tier_review.py
+++ b/base_tier_validation_formula/models/tier_review.py
@@ -29,7 +29,10 @@ class TierReview(models.Model):
 
     @api.depends("definition_id.reviewer_expression", "review_type", "model", "res_id")
     def _compute_python_reviewer_ids(self):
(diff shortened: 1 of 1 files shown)


etobella (approved):
LGTM

victoralmau (approved):

pedrobaeza (changes requested):
I think this is not correct when no `definition_id.reviewer_expression` is defined, landing in this line: https://github.com/OCA/server-ux/pull/253/files#diff-e4429a6dfbf0c3b86048e58f796de3404a1b277811e7036f3502794460ff8303R50

You should assign them instead False

LoisRForgeFlow:
@pedrobaeza could you check if this is correct now?

pedrobaeza (approved):

LoisRForgeFlow:
@sergio-teruel could you forward port if needed?

PR #254, opened 2020-12-28
[FIX] Tier Definition should  order by sequence.

tanghulu0608:


kittiu:
Noramally it is ordered by id. Which i turn already order by sequence already.
Do you have case when you have problem currently?

tanghulu0608:
I think it is better to order by sequence. 
Cases
1. When all tier definitions are shown in a list(tree) view, we can see the validation steps one by one.
2. When developing a customize feature which needs to select a tier definition within a many2one field,It is easier to choose a tier definition when they are orderd by sequence.

kittiu:
I am ok with this change.

base_tier_validation/tests/test_tier_validation.py
```diff
@@ -218,7 +218,7 @@ def test_12_approve_sequence(self):
                 "reviewer_id": self.test_user_1.id,
                 "definition_domain": "[('test_field', '>', 1.0)]",
                 "approve_sequence": True,
-                "sequence": 30,
```
kittiu, line 221: In current implementation, higher sequence here will get lower sequence here.
Meaning, when you set sequence 30, 20, 10 here. The review table will be 1, 2, 3.

I think this is by design AFSIK.

tanghulu0608: In this pr，I have change to : higher sequence here will get higher sequence.

kittiu: This issue has been discussed over time, I don't remember the detail, may be we should check with the original author first. :sweat_smile: 
I think it will conflict with many existing installment and many test (not only this module but other module that depend on this one). While I also think it is more intuitive, we have to be careful.
cc @LoisRForgeFlow

tanghulu0608: I agree with you. 🙂
I have modify this module to use in my system, and it runs well.😊

kittiu: Great, so you can make some extra module on your own. :+1: 

Can you close this PR then. Thanks!

tanghulu0608: OK

PR #255, opened 2020-12-29
[14.0][MIG+REF] mass_editing : Migration + Refactor to rely on ir.actions.server

ivantodorovich:
Migration + Refactor of `mass_editing` to rely on `ir.actions.server` and to deprecate `mass_operation_abstract`.
By doing it we can remove a lot of duplicated code, reducing the module maintenance.

This is the result of the following discussions:
- https://github.com/OCA/server-ux/pull/97#pullrequestreview-317654463
- https://github.com/OCA/reporting-engine/pull/457#discussion_r527568225

**Main changes:**
- Deprecated `mass_operation_abstract`.
- Add module `server_action_domain`: isolate this feature in a standalone module, applicable on server actions. This module is optional. See PR https://github.com/OCA/server-ux/pull/257
- Removed "Mass Editing" menu. Mass Edit operations can be configured through Server Actions

**TODO:**
- [x] ~~Add migration scripts for `14.0.1.0.0`: migrate data from 13.0~~

ping @legalsylvain @hbrunn @YogeshMahera-SerpentCS @michotm


mass_editing/models/mass_editing_line.py
```diff
@@ -0,0 +1,58 @@
...
+    )
+    widget_option = fields.Char(
+        string="Widget Option",
+        help="Add widget text that will be used to display the field in the wizard.\n"
+        "Example: 'many2many_tags', 'selection', 'image'",
+        compute="_compute_widget_option",
+        readonly=False,
+        store=True,
+    )
+    apply_domain = fields.Boolean(
```
legalsylvain, line 47: Hi. Could you say the reason of this field. I dont understand why it exists. It is new ?

mass_editing/models/mass_editing_line.py
```diff
@@ -0,0 +1,58 @@
...
+        string="Apply domain",
+        help="Apply default domain related to field",
+    )
+
+    @api.depends("field_id")
+    def _compute_widget_option(self):
+        for rec in self:
+            if rec.field_id.ttype == "many2many":
+                rec.widget_option = "many2many_tags"
+            elif rec.field_id.ttype == "binary":
```
legalsylvain, line 66: Maybe a little optimistic, dont you think ? Or you could add test if name contains image or logo. You should handle 99% of the images. Dont you think ?

legalsylvain (commented):
A quick partial review. Thanks a lot for this huge refactoring !  2 questions inline.

mass_editing/migrations/14.0.1.0.0/pre-migrate.py
```diff
@@ -0,0 +1,83 @@
...
+def migrate_mass_editing(env):
+    """ Migrates mass.editing to ir.actions.server """
+    # Remove FK (mass_editing_id)
+    openupgrade.remove_tables_fks(env.cr, ["mass_editing_line"])
+    # Create ir.actions.server for each mass.editing
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE ir_act_server
+        ADD COLUMN __legacy_mass_editing_id__       int4,
```
legalsylvain, line 22: Openugrade has a function to rename column. You could use it i think.

mass_editing/migrations/14.0.1.0.0/post-migrate.py
```diff
@@ -0,0 +1,38 @@
...
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if openupgrade.table_exists(env.cr, "mass_editing"):
+        migrate_mass_editing(env)
+
+
+def migrate_mass_editing(env):
+    """ Migrates mass.editing to ir.actions.servere """
+    # Remove previous context actions
```
legalsylvain: Typo.

mass_editing/migrations/14.0.1.0.0/post-migrate.py
```diff
@@ -0,0 +1,38 @@
...
+        SELECT id FROM ir_act_server
+        WHERE state = 'mass_edit'
+        AND __legacy_ref_ir_act_window_id__ IS NOT NULL
+        """
+    )
+    server_action_ids = [r[0] for r in env.cr.fetchall()]
+    if server_action_ids:
+        env["ir.actions.server"].browse(server_action_ids).create_action()
+    # Cleanup
+    openupgrade.logged_query(
```
legalsylvain: Generaly such cleanup is done after, with database_cleanup module.

mass_editing/models/mass_editing_line.py
ivantodorovich: Thanks for the review @legalsylvain  :)

This was introduced by @victoralmau here: https://github.com/OCA/server-ux/pull/237
> Add apply domain option in lines to fix some many2one fields with default domain (company_id for example)

mass_editing/models/mass_editing_line.py
ivantodorovich: My bad. Somehow I missed that line when I rewrote this block :sweat_smile: Thanks!!
It's fixed now

mass_editing/migrations/14.0.1.0.0/post-migrate.py
ivantodorovich: Thanks! :)

mass_editing/migrations/14.0.1.0.0/post-migrate.py
ivantodorovich: Yes but these are temporary columns created by this script, so I thought it's better to remove them.
The do not belong to previous fields, so they have no use whatsoever

mass_editing/models/mass_editing_line.py
pedrobaeza: @legalsylvain this was for fixing something you introduce with your refactoring, and it's that if you apply the default field's domain, you need to add the fields used in the domain to the view, or you will get an error, so the apply domain is now optional. If applied, you need to add the needed fields. Clear example in v13: fields with `check_company=True`. They need to add always `company_id` field in the view, or you will get an error.

joao-p-marques (approved):
Code and functional review :+1: 

The refactor looks very good :ok_hand:

mass_editing/__manifest__.py
```diff
@@ -2,21 +2,21 @@
...
     "GRAP, "
+    "Iván Todorovich, "
     "Odoo Community Association (OCA)",
     "category": "Tools",
     "website": "https://github.com/OCA/server-ux",
     "license": "AGPL-3",
     "summary": "Mass Editing",
-    "uninstall_hook": "uninstall_hook",
-    "depends": ["mass_operation_abstract"],
+    "depends": ["server_action_domain"],
```
pedrobaeza: Can you avoid this dependency? We are removing one, but adding a new one, so the balance is similar. I would like to have this module again without any dependency.

joao-p-marques:
Could the functionality of server_action_domain be incorporated into 
this one?

mass_editing/__manifest__.py
joao-p-marques: Could the functionality of server_action_domain be incorporated into
this one?

mass_editing/migrations/14.0.1.0.0/post-migrate.py
```diff
@@ -0,0 +1,38 @@
...
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if openupgrade.table_exists(env.cr, "mass_editing"):
+        migrate_mass_editing(env)
+
+
+def migrate_mass_editing(env):
```
pedrobaeza, line 8: You need to put this method before the method that invokes it.

mass_editing/migrations/14.0.1.0.0/post-migrate.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+    )
+    server_action_ids = [r[0] for r in env.cr.fetchall()]
+    if server_action_ids:
+        env["ir.actions.server"].browse(server_action_ids).create_action()
+    # Cleanup
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE ir_act_server
```
pedrobaeza: Please remove this. Migration scripts mustn't remove anything, as it can be needed for dependent modules to do their migration. Imagine you add a custom module that depends on this one and adds a field on the model and needs to be propagated to the new one. Removing this, you are blocking such propagation as you are removing the link.

mass_editing/migrations/14.0.1.0.0/pre-migrate.py
pedrobaeza: Please use usual naming convention: `openupgrade.get_legacy_name("mass_editing_id")`

mass_editing/migrations/14.0.1.0.0/pre-migrate.py
```diff
@@ -0,0 +1,83 @@
...
+        ADD COLUMN __legacy_ref_ir_act_window_id__  int4
```
pedrobaeza: The same

mass_editing/migrations/14.0.1.0.0/pre-migrate.py
```diff
@@ -0,0 +1,83 @@
...
+        FROM mass_editing me
+        LEFT JOIN ir_model mo ON (me.model_id = mo.id)
+        """,
+    )
+    # Delete fields that no longer exist (ondelete didn't exist before)
+    openupgrade.logged_query(
+        env.cr,
+        """
+        DELETE FROM mass_editing_line
+        WHERE field_id IS NULL
```
pedrobaeza: Don't delete anything as explained before

mass_editing/migrations/14.0.1.0.0/pre-migrate.py
```diff
@@ -0,0 +1,83 @@
...
+        OR field_id NOT IN (SELECT id FROM ir_model_fields)
+        """,
+    )
+    # Migrate mass.editing.line
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE mass_editing_line
+        ADD COLUMN server_action_id int4
```
pedrobaeza: Do this on post-migration for avoiding to create the column on advance, or use `add_field` openupgrade method

mass_editing/models/mass_editing_line.py
```diff
@@ -1,51 +1,67 @@
...
+        "ir.model.fields",
         string="Field",
-        domain="["
-        "('name', '!=', '__last_update'),"
-        "('ttype', 'not in', ['reference', 'function']),"
-        "('model_id', '=', model_id)"
-        "]",
+        domain="""
+            [
+                ("name", "!=", "__last_update"),
```
pedrobaeza: You should include here any of the "magic fields". There's a variable in Odoo core (in upper case) that contains the list of all of them

mass_editing/models/mass_editing_line.py
```diff
@@ -1,51 +1,67 @@
...
+                ("ttype", "not in", ["reference", "function"]),
+                ("model_id", "=", model_id),
+            ]
+        """,
+        ondelete="cascade",
```
pedrobaeza, line 39: This shouldn't be cascade, but `restrict` IMO

pedrobaeza (changes requested):
Apart from the inline comments, `mass_editing` mustn't need to depend on server_action_domain. Both should be independent, although complementary. You can advise to install it if domain are needed for mass actions.

pedrobaeza:
Please also squash commits accordingly and I would say to let this module for the mass_editing migration and put the other module in a subsequent PR.

mass_editing/migrations/14.0.1.0.0/pre-migrate.py
ivantodorovich: If I don't do this I get this error:

```
odoo.sql_db: bad query: ALTER TABLE "mass_editing_line" ADD FOREIGN KEY ("field_id") REFERENCES "ir_model_fields"("id") ON DELETE cascade
ERROR: insert or update on table "mass_editing_line" violates foreign key constraint "mass_editing_line_field_id_fkey"
DETAIL:  Key (field_id)=(15266) is not present in table "ir_model_fields".
```

I think, as some fields have been removed from previous odoo version (normally), some mass editing lines are left in this inconsistent state. But if the field have been removed.. I think it's ok to remove the mass.editing.line, don't you think?

ivantodorovich:
Thanks @pedrobaeza for you review!
I'm working on it :)

mass_editing/models/mass_editing_line.py
ivantodorovich: Wouldn't that prevent module updates that remove fields, for instance?
What's the point on preventing a field removal just for a mass editing rule?

mass_editing/models/mass_editing_line.py
ivantodorovich: Actually they're not the same.
I think the developer intention was to ignore only this field: https://github.com/odoo/odoo/blob/14.0/odoo/models.py#L380
Magic fields like `create_date`, `write_date`, `create_uid`, `write_uid` can be mass editted, I don't see any reason not to..

I did do this: https://github.com/OCA/server-ux/pull/255/commits/5b63806003e8215e780a2a171e736b8fe59403fa, because `id` shouldn't be mass editted either

mass_editing/migrations/14.0.1.0.0/pre-migrate.py
ivantodorovich: done

mass_editing/migrations/14.0.1.0.0/pre-migrate.py
ivantodorovich: done

mass_editing/__manifest__.py
ivantodorovich: Done! Dependency removed, `server_action_domain` module is here: https://github.com/OCA/server-ux/pull/257

mass_editing/migrations/14.0.1.0.0/post-migrate.py
ivantodorovich: done!

mass_editing/migrations/14.0.1.0.0/post-migrate.py
ivantodorovich: Done!

ivantodorovich:
Do not merge yet. I still need to squash commits
I didn't do it yet so that it's easier for your to review

mass_editing/migrations/14.0.1.0.0/pre-migrate.py
pedrobaeza: The error you are mentioning doesn't relates to this code. Removing null lines is not the same as having a FK constraint. What you mention can happen, but maybe the solution is to keep `ondelete="cascade"` on the corresponding field.

mass_editing/models/mass_editing_line.py
pedrobaeza: That's not true: `create_date` and so on are automatically removed from any create/write dict.

mass_editing/models/mass_editing_line.py
pedrobaeza: The point can be to not allow that unless you remove explicitly the line for being aware, but thinking twice and with the other comment I have put, we can let this way for avoiding harder problems.

pedrobaeza:
Please comment back when finished

mass_editing/migrations/14.0.1.0.0/pre-migrate.py
ivantodorovich: I'm not sure I understand. `ondelete="cascade"` didn't exist before, I introduced it here.

After migrating the database, some `mass.editing.line` records are pointing to non-existing fields (because they have been removed by the corresponding modules)

Honestly I don't know why the missing field id (15266) is still there, the default ondelete value being `ondelete="set null"`, it should've set it to null..  however this is what we've received from upgrade.odoo.com

In any case you get the point. After the migration you're left with mass.editing.lines that will either:
- point to missing field ids (but let's assume that's out of scope, and consider this a bug in upgrade.odoo.com)
- field_id will be NULL

What's the point in keeping these lines? They fields don't exist anymore and they will only prevent the "field required" database constraint from being created.

I'll remove it if you like, I just think others might have the same issue without these lines and having them is consistent with the new `ondelete="cascade"` attribute

mass_editing/models/mass_editing_line.py
ivantodorovich: Done!

mass_editing/migrations/14.0.1.0.0/pre-migrate.py
pedrobaeza: OK, I get your point as well and it's OK to keep such code. OpenUpgrade doesn't remove such field entries.

ivantodorovich:
All review suggestions resolved then :)
This is ready

pedrobaeza:
It's OK, but I still think you should squash that 4 commits into one, as they don't add too much value to have them split.

ivantodorovich:
> It's OK, but I still think you should squash that 4 commits into one, as they don't add too much value to have them split.

Done!

pedrobaeza (approved):

legalsylvain:
Thanks !

PR #256, opened 2020-12-31
[14.0][FIX] base_technical_features: user setting in hidden group

StefanRijnhart:
The company_id field on the users form is hidden by default now: https://github.com/odoo/odoo/blob/14.0/odoo/addons/base/views/res_users_views.xml#L458-L459


HaraldPanten (approved):

cubells (approved):
:+1: 

LGTM

Code review and test

pedrobaeza (approved):

Issue #260, opened 2021-01-19
[12.0]  Mass editing does not work correctly with one2many fields


azucenatrey:
ERRORS DETECTED:

- When you select the "Set" option from the mass editing wizard and leave the one2many field empty with the aim that the empty value is assigned to all the selected records, that is, that their values are emptied, that assignment is not applied in the selected records. The value of the field is not modified.

- When you select the "Remove" option from the mass editing wizard, leaving the one2many field empty so that all records associated with said field are deleted, this deletion does not apply. The value of the field is not removed.

CAUSE:

Is being assigned False value in the writing of the one2many field instead of the empty assignment for this type of fields: [(6, 0, [])].

SOLUTION:

Detect if the field is of type one2many, and if so, assign [(6, 0, [])] instead of False.

I have a PR ready that solves it. PR proposed here: https://github.com/OCA/server-ux/pull/261

PR #261, opened 2021-01-19
[12.0] Mass editing does not work correctly with one2many fields

azucenatrey:
ERRORS DETECTED (open in issue https://github.com/OCA/server-ux/issues/260):

When you select the "Set" option from the mass editing wizard and leave the one2many field empty with the aim that the empty value is assigned to all the selected records, that is, that their values are emptied, that assignment is not applied in the selected records. The value of the field is not modified.

When you select the "Remove" option from the mass editing wizard, leaving the one2many field empty so that all records associated with said field are deleted, this deletion does not apply. The value of the field is not removed.

CAUSE:

Is being assigned False value in the writing of the one2many field instead of the empty assignment for this type of fields: [(6, 0, [])].

SOLUTION:

Detect if the field is of type one2many, and if so, assign [(6, 0, [])] instead of False.

### DIFF
--- a/mass_editing/wizard/mass_editing_wizard.py
+++ b/mass_editing/wizard/mass_editing_wizard.py
@@ -101,10 +101,17 @@ class MassEditingWizard(models.TransientModel):
                 if key.startswith("selection_"):
                     split_key = key.split("__", 1)[1]
                     if val == "set":
-                        values.update({split_key: vals.get(split_key, False)})
-
+                        if TargetModel._fields[split_key].type == 'one2many':
+                            values.update({
+                                split_key: vals.get(split_key, [(6, 0, [])])})
+                        else:
+                            values.update({
+                                split_key: vals.get(split_key, False)})
                     elif val == "remove":
-                        values.update({split_key: False})
+                        if TargetModel._fields[split_key].type == 'one2many':
+                            values.update({split_key: [(6, 0, [])]})
+                        else:
+                            values.update({split_key: False})
 
                         # If field to remove is translatable,
                         # its translations have to be removed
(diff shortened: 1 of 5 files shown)


rlizana (approved):
LGTM :+1:

ivantodorovich (commented):
I think the change should be done here: https://github.com/OCA/server-ux/blob/6d605d05cc0f11a46ae9e8110bc7a4747d24be81/mass_editing/wizard/mass_editing_wizard.py#L19 instead, so that it uses `remove_m2m` for one2many fields too, or at least using a similar approach

azucenatrey:
@ivantodorovich you are right, I have moved the field type selector functionality to "_prepare_fields" function.

mass_editing/wizard/mass_editing_wizard.py
```diff
@@ -22,6 +22,11 @@ def _prepare_fields(self, line, field, field_info):
                 ("remove_m2m", _("Remove")),
                 ("add", _("Add")),
             ]
+        elif field.ttype == "one2many":
+            selection = [
```
cubells, line 26:
```suggestion
            selection = [("set_o2m", _("Set")), ("remove_o2m", _("Remove"))]
```

Maybe this way coverage doesn't decrease.

cubells (changes requested):

mass_editing/wizard/mass_editing_wizard.py
azucenatrey: Thanks @cubells , I have applied the change that you have suggested but Travis gives an error, I don't know why. I have left it as it was before although it still shows:

    mass_editing/wizard/mass_editing_wizard.py#L26
    Added line #L26 was not covered by tests

cubells (approved):
:+1: 

LGTM

Code review and test

PR #263, opened 2021-01-24
[13.0][add][base_hide_menu] This module allows to explicitly indicate which menus you would like to hide to all users.

JordiBForgeFlow:
@davidbforgeflow @mateugforgeflow can you review?

### DIFF
--- a/base_hide_menu/__init__.py
+++ b/base_hide_menu/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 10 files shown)


pedrobaeza:
Why not using https://github.com/OCA/server-ux/tree/13.0/base_menu_visibility_restriction?

JordiBForgeFlow:
@pedrobaeza you're right. Better use that one instead cc @AdriaGForgeFlow

Issue #265, opened 2021-01-28
[14.0][BUG] base_tier_validation -> button template not work on all model.


kittiu:
Based on this,
https://github.com/OCA/server-ux/blob/14.0/base_tier_validation/templates/tier_validation_templates.xml#L8

The button will be invisible on certain status `('state','not in',['draft','sent','to approve'])`, which is ok for PR/PO.

But not OK for other model which may have different states.

@kittiu

kittiu:
Actually states should come from `_state_from` to be correct.

PR #266, opened 2021-01-28
[14.0][FIX] base_tier_validation, template only work on some model

kittiu:
https://github.com/OCA/server-ux/issues/265


etobella:
@kittiu I was thinking of another approach that might be easier to define (no new field and so on).
https://github.com/tegin/server-ux/tree/14.0-fix-tier
WDYT? You can get my commit or I can propose a PR if you like it :wink:

kittiu:
> @kittiu I was thinking of another approach that might be easier to define (no new field and so on).
> https://github.com/tegin/server-ux/tree/14.0-fix-tier
> WDYT? You can get my commit or I can propose a PR if you like it

Awesome! I also want to do somthing like that, but lack of technique that you have :+1: 
You can propose your PR please, I will approve. :)

I will close this one.

PR #269, opened 2021-02-01
[14.0][FIX] base_tier_validation : Error "The operation is under validation"

newtratip:
I'm migrating account_move_tier_validation https://github.com/OCA/account-invoicing/pull/853 and I tested `validate` by reviewers. It's not allow and throw error "The operation is under validation" always.
Peek 2021-02-01 20-44

I check on log and found some vals (not state) pass in this function
Selection_103

which this line check with state in vals and it will be True
Selection_104

This PR is fix for if state not in vals will not throw error "The operation is under validation"

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -184,6 +184,7 @@ class TierValidation(models.AbstractModel):
             if (
                 rec.review_ids
                 and getattr(rec, self._state_field) in self._state_from
+                and vals.get(self._state_field)
                 and not vals.get(self._state_field)
                 in (self._state_to + [self._cancel_state])
                 and not self._check_allow_write_under_validation(vals)


newtratip:
@kittiu Can you review this ?

kittiu:
Have you check this method. You may need to add field that allowed changes during validation.

`def _get_under_validation_exceptions(self):`

newtratip:
> Have you check this method. You may need to add field that allowed changes during validation.
> 
> `def _get_under_validation_exceptions(self):`

@kittiu , Do you mean put access_token in `_get_under_validation_exceptions` function ?

I don't know why access_token pass in this function ?

I tested with `sale_tier_validation` module, It has a same problem.

Selection_158

Please suggest me, How should I do ?

kittiu:
I think, this will fix your problem. https://github.com/OCA/server-ux/pull/283

newtratip:
Thank @kittiu, This issue is fixed. I will close the PR.

PR #270, opened 2021-02-02
[14.0][FIX] base_tier_validation_forward inverse value problem

kittiu:
Problem with related field and caching. Following problem can occur.

Given this module `base_tier_validation_forward` is installed. As this module overwrite the filed `name` in tier.review, as readonly=False. In some situation, the name of record to process tier validation will be used as the name of tier.review (because readonly=False, and it was related field before).

In any case, readonly=True is good as protection.


kittiu:
Sorry, wrong commits history.

PR #273, opened 2021-02-05
[14.0][FIX] base_tier_validation, improve accuracy of systray on case approve_sequence

kittiu:
For case approve by sequence. At currently status,

* On the record, the button showing correctly on the next document to validate (thanks to rec.can_review)
* The systray count is already correct.
* But when click on systray to open the document list, still not taken sequence into account, and so it list also the document need approve in the future sequence. (i.e., systray says 2 docs, but it may open 3 docs)

This PR enhance that, to make sure that we use rec.can_review to help determine the correct document list.

Note: this fix should be applied to version 12.0 and 13.0 where we have approve by sequence feature.

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -43,7 +43,9 @@ class TierValidation(models.AbstractModel):
         compute="_compute_reviewer_ids",
         search="_search_reviewer_ids",
     )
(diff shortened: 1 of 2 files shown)


kittiu:
cc @LoisRForgeFlow

LoisRForgeFlow (approved):
Indeed this is more consistent.

etobella (approved):

PR #275, opened 2021-02-09
[14.0][IMP] base_tier_validation, add 'Todo by' column in review table

kittiu:
This PR add new column that allow users to see who is next in approve line.

image

### DIFF
--- a/base_tier_validation/models/tier_review.py
+++ b/base_tier_validation/models/tier_review.py
@@ -1,7 +1,7 @@
 # Copyright 2017-19 ForgeFlow S.L. (https://www.forgeflow.com)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
 
 
 class TierReview(models.Model):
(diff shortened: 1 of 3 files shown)


etobella (approved):
LGTM

base_tier_validation/models/tier_review.py
```diff
@@ -77,5 +78,10 @@ def _compute_reviewer_ids(self):
         for rec in self:
             rec.reviewer_ids = rec._get_reviewers()
 
+    @api.depends("reviewer_ids")
+    def _compute_todo_by(self):
+        for rec in self:
+            rec.todo_by = ", ".join(rec.reviewer_ids.mapped("display_name"))
```
LoisRForgeFlow: I'm not sure I like this, if you make tiers by groups, the list can be massive. Some ideas:
1. You could show the group name if it is a tier by group.
2. You can truncate the list if too long. E.g: "James Jamison, Daniel Davidson, Jacks Jackson (and 12 more .).

kittiu: Agree with option 2, I will revise it a bit for that. Thanks for your review.

kittiu: image

kittiu: Done! if it is a group, show group name, if it is a list show abbreviation list.

etobella: :+1:

LoisRForgeFlow: That is great, thanks!

LoisRForgeFlow (approved):
LGTM

PR #279, opened 2021-02-17
[13.0] [MIG] base_import_security_group

Murtaza-OSI:


Chandresh-OSI:
@StefanRijnhart Do you have any idea about the test case failing issue in newer version?

StefanRijnhart:
Oh this test is terribly unreliable. @hbrunn came up with https://github.com/OCA/server-ux/pull/290 for this one, and even that one fails occasionally. @rconjour recently suggested to make this into a QUnit test. I have no experience with that. What do you think?

hbrunn:
I actually never managed to reproduce the issue in my draft PRs I kept rebuilding, so for the time being I think we're better off with this version for sure

StefanRijnhart:
@hbrunn for the record, I've had our project with #290 fail twice on me recently. I do think that is less than the current implementation, but it's not ideal.

hbrunn:
@StefanRijnhart it's hard to debug this because we can't ssh into an existing build that's not debug, but I think I'll need the screenshots taken during the test to see what's going on there. Maybe I'll add a script that pushes them as artifacts somewhere.

base_import_security_group/security/base_import_security_group_security.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+
+    <!-- GROUPS -->
+    <record id="group_import_csv" model="res.groups">
+        <field name="name">Import CSV/Excel files</field>
+        <field name="category_id" ref="base.module_category_hidden" />
```
chienandalu, line 7: Remove so it's shown as a checkbox in the other permissions group
```suggestion
```

chienandalu (commented):
My colleague @victoralmau is trying to fix tests here https://github.com/OCA/server-ux/pull/336 Let's see if we can join efforts to unblock this :slightly_smiling_face:

StefanRijnhart:
Tests were rewritten successfully on the duplicate PR, so I think we should go for that one: https://github.com/OCA/server-ux/pull/336. Thanks nevertheless for proposing the port!

pedrobaeza:
Superseded by #336

PR #283, opened 2021-02-18
[14.0][FIX] base_tier_validation, access_token as exception field

kittiu:
Based on, https://github.com/OCA/sale-workflow/pull/1386#issuecomment-781371729


newtratip (approved):

LoisRForgeFlow (approved):
This a generic enough field to be included here, so LGTM :+1:

PR #285, opened 2021-02-22
[13.0] [IMP] base_technical_features: use has_group instead of loading all users

ThomasBinsfeld:
On databases with a large number of users, loading all users of base.group_no_one can be slow.
Moreover, the 'technical_features' and 'show_technical_features' are always computed for a single record


etobella (approved):
LGTM

pedrobaeza (approved):

pedrobaeza:
/ocabot merge patch

Can you fw-port it to v14?

PR #287, opened 2021-02-23
[13.0][FIX] base_tier_validation, improve accuracy of systray on case approve_sequence

kittiu:
Back port fix from v14 https://github.com/OCA/server-ux/pull/273


LoisRForgeFlow (approved):
:+1:

PR #290, opened 2021-02-23
[FIX] use more robust ready indicator

hbrunn:
this should avoid timing issues during tests


StefanRijnhart:
To have this fixed would be wonderful. Why does this work better?

hbrunn:
jQuery returns an empty jQuery collection, but it's still an object, which when cast to boolean is true. In JS, also the empty object is still an object:

```javascript
> Boolean({})
true
```

StefanRijnhart (approved):
Great, thanks!

StefanRijnhart:
Superseded by https://github.com/OCA/server-ux/pull/341

PR #293, opened 2021-02-25
[14.0] [mig] chained_swapper

schhatbar-initos:


dsolanki-initos (approved):
LGTM

schhatbar-initos:
@ernestotejeda Can you please review this PR?

ernestotejeda:
When I try to access contacts I get this error
image

I think it has to do with this module, because if I uninstall it, I no longer have the error, please check it.

ernestotejeda:
maybe It will have to do with the demo data, where an ir.actions.act_window associated with res.partner is created.
what do you think?

schhatbar-initos:
> maybe It will have to do with the demo data, where an ir.actions.act_window associated with res.partner is created.
> what do you think?

Yes! There was something wrong with context. I've changed the code for the same.

Thank You!

ernestotejeda (approved):

chained_swapper/tests/test_chained_swapper.py
```diff
@@ -2,6 +2,7 @@
 # Copyright 2020 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+
```
pedrobaeza: Why adding this extra line?

chained_swapper/tests/test_chained_swapper.py
```diff
@@ -85,12 +95,32 @@ def test_change_constrained_partner_language(self):
...
+        chained_many2one = (
+            self.env["chained.swapper.wizard"]
+            .with_context(
+                active_model="product.product",
+                chained_swapper_id=self.chained_swapper_inherits.id,
+            )
+            .create({"product_tmpl_id": 1})
+        )
+        chained_many2one.fields_view_get()
+        not_chained_swapper_id = self.env["chained.swapper.wizard"].with_context(
```
pedrobaeza: Why the `_id` suffix? It's not an integer.

chained_swapper/tests/test_chained_swapper.py
```diff
@@ -85,12 +95,32 @@ def test_change_constrained_partner_language(self):
...
             active_model="res.partner",
             active_id=self.partner_parent.id,
             active_ids=[self.partner_parent.id],
-            chained_swapper_id=self.chained_swapper.id,
-        ).create({"lang": "es_ES"})
+        )
+        not_chained_swapper_id.fields_view_get()
```
pedrobaeza: Why do you need to do this?

schhatbar-initos: I added this to coverage the code. As code patch was failing, so as suggested in this PR some lines were not covered by tests. I added them to make it work the condition.

pedrobaeza: But that's a dirty trick. You are not doing any testing at all, just tinkering the coverage report, and what is more important, you are not proving that the real thing won't explode, si better to remove it.

chained_swapper/tests/test_chained_swapper.py
```diff
@@ -0,0 +1,189 @@
...
+            self.env["chained.swapper.wizard"]
+            .with_context(
+                active_model="res.partner",
+                active_id=self.partner_parent.id,
+                active_ids=[self.partner_parent.id],
+                chained_swapper_id=self.chained_swapper.id,
+            )
+            .create({"lang": "es_ES"})
+        )
+        chained.fields_view_get()
```
pedrobaeza: This is still calling the same method without testing nothing real

chained_swapper/tests/test_chained_swapper.py
```diff
@@ -0,0 +1,184 @@
...
+        self.assertEqual(self.partner_child_2.lang, "es_ES")
+        chained_many2one = (
+            self.env["chained.swapper.wizard"]
+            .with_context(
+                active_model="product.product",
+                chained_swapper_id=self.chained_swapper_inherits.id,
+            )
+            .create({"product_tmpl_id": 1})
+        )
+        chained_many2one.fields_view_get()
```
pedrobaeza: There's still call to this method + code that doesn't test the real purpose of the module.

schhatbar-initos: In this test case, I should not use fields_view_get or use it in separate testing?

pedrobaeza: You are not adding anything calling it, as this is an internal method called by action controller in the web client, so you are not testing anything real even if you call it manually and test the return result.

schhatbar-initos: So should I consider fields_view_get method in test cases or not?

pedrobaeza: I don't think so, at least this way. The only way to do it "real" is to create a QUnit test that simulates your navigation in Odoo web client.

fshah-initos:
Can somone please review the PR?

fshah-initos:
It would be great if this PR is merged.

chained_swapper/__manifest__.py
```diff
@@ -8,7 +8,7 @@
     "category": "Tools",
     "website": "https://github.com/OCA/server-ux",
     "license": "AGPL-3",
-    "depends": ["base"],
+    "depends": ["product"],
```
pedrobaeza: This dependency change is incorrect

pedrobaeza (changes requested):

fshah-initos:
I have adjusted the dependency change.

pedrobaeza:
Why not doing such tests in an existing object instead of changing dependency or now removing them?

pedrobaeza (approved):

PR #267, opened 2021-01-28
[14.0][FIX] tier_validation: Fix views from template as state was not used

etobella:
Continues #266 and fixes #265 

@LoisRForgeFlow @kittiu

PR #294, opened 2021-02-26
[13.0][BACKPORT] base_tier_validation: Remove unnecessary modification of views

LoisRForgeFlow:
Backport of https://github.com/OCA/server-ux/pull/248 and https://github.com/OCA/server-ux/pull/267.

cc @etobella 

@ForgeFlow


etobella (approved):
It seems an easy backport :smile:

kittiu:
Functional test OK !

But from the test error, you probably need to rebase. Your branch still not have this file -> https://github.com/OCA/server-ux/blob/13.0/base_tier_validation/tests/common.py

kittiu (approved):

LoisRForgeFlow:
> Functional test OK !
> 
> But from the test error, you probably need to rebase. Your branch still not have this file -> https://github.com/OCA/server-ux/blob/13.0/base_tier_validation/tests/common.py

It is removed in this PR, as the approach changes. In any case, It is true that I need to adapt other modules tests too, will do!

PR #297, opened 2021-02-26
[14.0][ADD] base_cancel_confirm

kittiu:
This module is a base module that make it easy to extend cancellation dialog, with cancel reason (none / optional / required)

Selection_044

For example, by base on this module, following code (and nothing else) is enough to make `sale_cancel_confirm`, 

```
class SaleOrder(models.Model):
    _name = "sale.order"
    _inherit = ["sale.order", "base.cancel.confirm"]

    _has_cancel_reason = "optional"  # ["no", "optional", "required"]

    def action_cancel(self):
        if not self.filtered("cancel_confirm"):
            return self.open_cancel_confirm_wizard()
        return super().action_cancel()
```

**Note:** We use `sale_cancel_reason` before, but we have requirement to make cancel confirm/reason for almost 10 kind of documents. Writing xxx_cancel_reason one by one is not a good idea.

TODO:

- [x] Unit test


kittiu:
You can see implementation example as followings,

* `sale_cancel_confirm` https://github.com/OCA/sale-workflow/pull/1490
* `purchase_cancel_confirm` https://github.com/OCA/purchase-workflow/pull/1125
* `purchase_request_cancel_confirm` https://github.com/OCA/purchase-workflow/pull/1126
* `account_move_cancel_confirm` https://github.com/OCA/account-invoicing/pull/872

cc @ps-tubtim @SurachetS @Saran440

rousseldenis (approved):
Code review. Looks great

ps-tubtim (approved):

atchuthan:
@kittiu Why did we move the reasons as text? instead of many2one.
We could have used the new object and add model as many2many such that we can show the reasons based on the model it is being triggered.  Also, set active field to disable if needed in future.

Is there any technical difficulty due to which this change has been made.  Please let us know

cc: @theangryangel

PR #298, opened 2021-02-27
[MIG] mass_operation_abstract: Migration to 14.0

dzungtran89:


mass_operation_abstract/wizard/mass_operation_wizard_mixin.py
```diff
@@ -44,24 +44,26 @@ def default_get(self, fields):
         operation_description_danger = False
 
         # Compute items quantity
-        active_ids = self.env.context.get("active_ids")
+        active_ids = self._context.get("active_ids")
```
simahawk: never use a private attr if you have a public one (which is more likely to stay). In any case these changes are not required.

mass_operation_abstract/models/mass_operation_mixin.py
```diff
@@ -15,7 +15,7 @@ class MassOperationMixin(models.AbstractModel):
 
     # To Overwrite Section (Optional)
     def _prepare_action_name(self):
-        return _("Mass Operation (%s)" % (self.name))
+        return _("Mass Operation ({})".format(self.name))
```
simahawk: better to keep the old version

simahawk (changes requested):
LG overall

mass_operation_abstract/wizard/mass_operation_wizard_mixin.py
dzungtran89: thx. I'll update it

simahawk:
sorry I missed to link this PR https://github.com/OCA/server-ux/pull/348 could you have a look and see if we can integrate it here? Otherwise we can open another PR later.

dzungtran89:
hi @simahawk , I ported the update of #348 to this PR, could you pleave have a look?

simahawk:
> hi @simahawk , I ported the update of #348 to this PR, could you pleave have a look?

hmm I don't see the commit :thinking:

dzungtran89:
ah, I actually had appended the update to the migration commit. I just created another commit then

simahawk:
@dzungtran89 you are losing autorship on that commit. Can't you simply cherry-pick? Or use format patch from scratch?

dzungtran89:
Good day @simahawk, thx for the advice, the PR has been updated

simahawk (approved):

mmequignon (approved):

PR #299, opened 2021-03-01
[14.0][mig] default_multi_user

schhatbar-initos:


default_multi_user/tests/test_default_multi_user.py
```diff
@@ -3,58 +3,59 @@
...
 
 
-class TestDefaultMultiUser(common.SavepointCase):
-    @classmethod
-    def setUpClass(cls):
-        super().setUpClass()
-        cls.env = cls.env(
-            context=dict(cls.env.context, tracking_disable=True, no_reset_password=True)
+@tagged("post_install", "-at_install")
+class TestDefaultMultiUser(TransactionCase):
```
LoisRForgeFlow: Any spacial reason to change to TransactionCase?

schhatbar-initos: @LoisRForgeFlow I've changed it to SavepointCase again. No need to convert it into TransactionCase so that all the test cases can use the same data without having to recreate test data and also use of setUpClass() is correct to create DB test data once.
Thanks!

dsolanki-initos (approved):
LGTM

LoisRForgeFlow (approved):
LG, thanks!

Issue #306, opened 2021-03-09
[BUG] base_tier_validation, change in tier.definition's reviewer should not change reviews in tier.review


kittiu:
Currently, the reviewers in tier.reviews of the ongoing documents will be updated (without notice), if the related tier.definition's reviewers is updated.

Steps to error:

1. Create a tier definition "Tier 1", assign "User 1" as reviewer.
2. Start Validation in a new document, now next reviewer is "User 1"
3. Back to tier definition "Tier 1", change reviewer to "User 2".
4. The document now change reviewer to "User 2" ????

I think this is not correct, the only time the tier.review table get updated, **is when Restart Validation only**. Otherwise, it result in a mess for the on going documents.

@LoisRForgeFlow @etobella do you think this is a valid bug (or a feature)?

JordiBForgeFlow:
The reviewer_ids in the tier validation will need to be set as a computed, stored table then, only to be recomputed when you set/reset thr validation

kittiu:
> The reviewer_ids in the tier validation will need to be set as a computed, stored table then, only to be recomputed when you set/reset the validation

If so, I confirm this as a BUG, as it is changed right after edit the Definition.

I will propose a PR to fix it.
Thanks!

PR #308, opened 2021-03-10
[14.0][FIX] base_tier_validation, change definition should not affect reviews

kittiu:
Based on https://github.com/OCA/server-ux/issues/306

I tried to use related field and stored=True, also not fix. Computed store=True, seem to be the only option.


kittiu:
Hi @JordiBForgeFlow, are you ok with this change?

PR #318, opened 2021-03-23
Fix avoid_quick_create with multiprocess

SimoRubi:
Adding https://github.com/OCA/server-tools/pull/1454.
I think it has been lost when moving the module across repositories, or maybe has been merged after that because the commit simply does not exist in https://github.com/OCA/server-ux/commits/12.0/base_optional_quick_create but exists in https://github.com/OCA/server-tools/commits/10.0/base_optional_quick_create

### DIFF
--- a/base_optional_quick_create/models/ir_model.py
+++ b/base_optional_quick_create/models/ir_model.py
@@ -50,4 +50,6 @@ class IrModel(models.Model):
     def write(self, vals):
         res = super().write(vals)
         self._patch_quick_create()
+        if 'avoid_quick_create' in vals:
+            self.pool.signal_registry_change()
         return res
--- a/base_optional_quick_create/models/ir_model.py
+++ b/base_optional_quick_create/models/ir_model.py
@@ -50,6 +50,7 @@ class IrModel(models.Model):
     def write(self, vals):
         res = super().write(vals)
         self._patch_quick_create()
-        if 'avoid_quick_create' in vals:
-            self.pool.signal_registry_change()
+        if "avoid_quick_create" in vals:
+            self.pool.registry_invalidated = True
+            self.pool.signal_changes()
         return res


etobella (approved):
LGTM

base_optional_quick_create/models/ir_model.py
```diff
@@ -50,4 +50,7 @@ def create(self, vals_list):
     def write(self, vals):
         res = super().write(vals)
         self._patch_quick_create()
+        if "avoid_quick_create" in vals:
+            self.pool.registry_invalidated = True
+            self.pool.signal_changes()
```
StefanRijnhart, line 55: Isn't it enough to set `registry_invalidated`? I think any `write` call will be wrapped in `manage_changes` here: https://github.com/odoo/odoo/blob/13.0/odoo/service/model.py#L40, so that `signal_changes` is called here: https://github.com/odoo/odoo/blob/13.0/odoo/modules/registry.py#L507, and that method will indeed undertake action when `registry_invalidated` is set (here: https://github.com/odoo/odoo/blob/13.0/odoo/modules/registry.py#L475).

StefanRijnhart: Oops links are to 13.0 but same code can be found in Odoo 12.0, just on different line numbers.

SimoRubi: Hi, thanks for having a look at this!
Unfortunately I'm looking at all your links after some time so the line numbers are quite different.
Anyway I looked into it and just invalidating the registry is enough for triggering a signal to other workers because the response computation is all wrapped in a web request context manager (https://github.com/odoo/odoo/blob/716f36b172c4e129f2ba049ae33faa7d42fc500e/odoo/http.py#L1467-L1492) and it signals for changes on exit (https://github.com/odoo/odoo/blob/716f36b172c4e129f2ba049ae33faa7d42fc500e/odoo/http.py#L287).

This signaling is very far from the code where we need the signaling to be done (this PR's code) and I'd rather keep the signaling here too because if the signaling is removed from https://github.com/odoo/odoo/blob/716f36b172c4e129f2ba049ae33faa7d42fc500e/odoo/http.py#L287 then it would be very hard to find out why this module stopped working.
Also, I think having the signaling in here is more explicit on the fact the we need the signaling to be done after this `write`, given the condition `"avoid_quick_create" in vals`.

TL;DR: you are right, but I'd rather keep it like this.

SimoRubi:
@OCA/server-environment-maintainers (I hope it's the correct team) is it possible to merge this or what do I have to change in order to have this merged?
Thanks

base_optional_quick_create/models/ir_model.py
StefanRijnhart: OK, thanks for giving it a thought.

StefanRijnhart:
Can you squash your commits? I'm thinking a rebase might be good to get a fresh set of CI tests.

StefanRijnhart (approved):

SimoRubi:
> Can you squash your commits? I'm thinking a rebase might be good to get a fresh set of CI tests.

I have done the rebase, let's see how the CI behaves.

About the commits: I kept them separated because the first one is the clean cherry-pick of https://github.com/OCA/server-tools/pull/1454, then I had to change something in order to make it work in `12.0` and I did this in a separate commit in order not to mix the authorship.
Actually I never know how to correctly manage this situation, any advice? I think making my changes in another author's commit is not correct but also a _fixup_ commit is not very nice to have in a _master_ branch.

SimoRubi:
@OCA/server-environment-maintainers can someone merge or what do I have to change in order to have this merged?
Thanks

PR #322, opened 2021-04-02
[12.0][IMP] base_tier_validation: display reviewer in Reviews Table

astirpe:
In case the review is still pending and the reviewer is a specific user, display the reviewer name.

### DIFF
--- a/base_tier_validation/models/tier_review.py
+++ b/base_tier_validation/models/tier_review.py
@@ -107,8 +107,8 @@ class TierReview(models.Model):
                 )
                 num_users = len(rec.reviewer_ids)
                 if num_users > num_show:
-                    todo_by = "{} (and {} more)"\
-                        .format(todo_by, num_users - num_show)
+                    todo_by = _("%s (and %s more)")\
+                        % (todo_by, num_users - num_show, )
             rec.todo_by = todo_by
 
     def _get_reviewers(self):
--- a/base_tier_validation/models/tier_review.py
+++ b/base_tier_validation/models/tier_review.py
@@ -34,7 +34,11 @@ class TierReview(models.Model):
         compute="_compute_reviewer_ids", store=True,
     )
     sequence = fields.Integer(string="Tier")
-    todo_by = fields.Char(compute="_compute_todo_by", store=True)
+    todo_by = fields.Char(
+        compute="_compute_todo_by",
+        compute_sudo=True,
+        store=True,
+    )
     done_by = fields.Many2one(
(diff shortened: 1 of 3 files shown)


kittiu:
Not sure this is something you look for, https://github.com/OCA/server-ux/pull/275

astirpe:
@kittiu yes, indeed! I will revert my commit and backport your solution. Thanks!

astirpe:
@kittiu would you have a look at https://github.com/OCA/server-ux/pull/322/commits/e752f9bf4c8c95602ef6e5a9741e1cd8a8636a3c? It seems fixing the failure in the tests and I'm wondering: shouldn't it apply also to V14?

kittiu:
> @kittiu would you have a look at [e752f9b](https://github.com/OCA/server-ux/commit/e752f9bf4c8c95602ef6e5a9741e1cd8a8636a3c)? It seems fixing the failure in the tests and I'm wondering: shouldn't it apply also to V14?

Yeah, it seem valid. Can you pick it to v13, v14 as well _/l\_

astirpe:
@kittiu I updated my commit. According to the [V12 ORM documentation](https://www.odoo.com/documentation/12.0/reference/orm.html), I used the `compute_sudo` to bypass the access right check instead of using the `sudo()`. Please notice that this fix seems not necessary for V13 and V14. This because from V13 the `compute_sudo` is the default for stored fields, see: [V13 ORM documentation](https://www.odoo.com/documentation/13.0/reference/orm.html).

HaraldPanten (approved):

base_tier_validation/models/tier_review.py
```diff
@@ -89,5 +94,22 @@ def _compute_reviewer_ids(self):
...
+        for rec in self:
+            if rec.reviewer_group_id:
+                todo_by = _("Group %s") % rec.reviewer_group_id.name
+            else:
+                todo_by = ", ".join(
+                    rec.reviewer_ids[:num_show].mapped("display_name")
+                )
+                num_users = len(rec.reviewer_ids)
+                if num_users > num_show:
+                    todo_by = "{} (and {} more)"\
```
StefanRijnhart: This should be translatable (and then you should use %s substitution because of https://github.com/OCA/pylint-odoo/issues/302)

astirpe: Done. Thanks!

StefanRijnhart (approved):
Thanks for the update @astirpe and thanks for the initial PR @kittiu!

PR #323, opened 2021-04-06
[13.0][REF] document_quick_access_folder_auto_classification:

etobella:
With this change, we are allowed to use different kinds of storages and not adding glue modules: This could avoid 
https://github.com/OCA/server-ux/pull/300

Still pending migration scripts


etobella:
@schhatbar-initos @LoisRForgeFlow

etobella:
Travis is failing due to OCA/edi#362

etobella:
Now is :green_apple: WDYT @LoisRForgeFlow @schhatbar-initos ?

etobella:
Migration scripts added, I think now it is ready :smile:

document_quick_access_folder_auto_classification/models/document_quick_access_rule.py
```diff
@@ -16,132 +12,44 @@
...
+        new_limit = limit
+        for backend in backends:
+            new_limit = self._cron_folder_auto_classification(backend, new_limit)
+
+    def _cron_folder_auto_classification_file(self, backend, file_data):
+        exchange_record = backend.create_record(
+            "document_quick_access",
+            {
+                "edi_exchange_state": "input_received",
+                "exchange_file": backend.storage_id.get(file_data, binary=False),
```
etobella, line 32: One issue I found with this is that we cannot use the same folder we were using, as storage is expecting the folder to be inside the odoo path dir.

albariera (approved):

olgamarcocb (approved):
:+1:

simahawk:
@etobella just stumbled upon this PR. I don' t know this module at all but I wonder if it's worth to pull in the whole EDI machinery if you only need ext storage capabilities. Couldn't you use `storage_backend` only and provide a cron to process documents?
Just asking! :wink:

etobella:
Well, at the end I want to keep a track of all received files, if something happened in order to retry them and so on. So, at the end, it is really similar to standard `edi` but with a complete different approach. I have been using it on production and everything works fine for me.

LoisRForgeFlow (approved):
Finally, I tested migrating this on an existing installation, all looks good :+1:

PR #324, opened 2021-04-07
[13.0][UPD] base_tier_validation: add notification.

acsonefho:
- Add possibility (into tier definition) to notify the asker of review when target is Accepted;
- Add possibility (into tier definition) to notify the asker of review when target is Rejected.

**Bonus:**
Extract params of `message_post(...)` to let others modules to add/edit/remove parameters.


kittiu:
@acsonefho thanks for this enhancement. I agree we should have this feature. Can you add test coverage to complete this?
Thanks!

PR #325, opened 2021-04-07
[12.0][IMP] base_tier_validation: add details to error message

astirpe:
This would be helpful to understand the reasons of why an error occurs.

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -184,10 +184,20 @@ class TierValidation(models.AbstractModel):
                     self._state_from and not vals.get(self._state_field) in
                     (self._state_to + [self._cancel_state]) and not
                     self._check_allow_write_under_validation(vals)):
-                raise ValidationError(_("The operation is under validation."))
+                err_msg = _("The operation is under validation.")
+                err_msg = self._add_details_to_error_message(err_msg, vals)
+                raise ValidationError(err_msg)
         if vals.get(self._state_field) in self._state_from:
             self.mapped('review_ids').unlink()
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow:
@astirpe What is the aim of this? I don't see much value on showing technical names of fields to users. If you need this information, maybe is better if you debug your use case to see why is not working as expected, don't you think?

EDIT: Another option is that the message is only extended when the debug mode is active. WDYT? What concerns me is that this is an informative error that happens often (it is the main point of it, to block modification of records under validation), so for me it is not an application error where you show a technical trace.

astirpe:
@LoisRForgeFlow I agree with you but we have an issue: https://github.com/OCA/account-invoicing/pull/871#issuecomment-791228153 and this kind of error could potentially occur with every module that extends the validation of an invoice. I didn't have a better solution for spotting which field is being written. Notice that the users who are entering a vendor bill are usually not so technical.

LoisRForgeFlow:
> Another option is that the message is only extended when the debug mode is active. WDYT?

what about this? or you can use the display string of the fields?

kittiu:
To me, I also don't see much use. The current message already fulfilled that nothing can be modified.

astirpe:
@LoisRForgeFlow I now display the list of fields when user belongs to group "Technical Features" and I display the field string name instead of the technical name. Thanks!

LoisRForgeFlow:
I have reviewd, I still see the a technical name:

image

astirpe:
@LoisRForgeFlow  You are right, was my mistake! Should be working now.

LoisRForgeFlow (approved):
Thanks!

PR #326, opened 2021-04-08
[FIX] date_range: typo in field keyword argument

len-foss:
The dumbest fix you'll see this day.
Still, this is annoying as it generates a WARNING dbname odoo.fields: Field date.range.generator.date_start: unknown parameter 'strint', if this is an actual parameter you may want to override the method _valid_field_parameter on the relevant model in order to allow it


pedrobaeza (approved):
As this is changing translations, I think is better to remove the entire argument.

len-foss:
@pedrobaeza oh you're entirely right, thanks for the catch!

pedrobaeza (approved):

pedrobaeza:
Fast-tracking this as obvious:

/ocabot merge patch

Issue #328, opened 2021-04-15
[14.0] Date range - Filter not working correctly


CasVissers-360ERP:
See video:
https://drive.google.com/file/d/1S2B-dz8vLypn2B-aZUHI9AMp7xQJCDv5/view

When applying a filter the filter is converted to a different date range.

@astirpe @len-foss

len-foss:
Hey @CasVissers-360ERP  :-)
You may have meant to ping someone else, I never even used this module.
Anyway, I solved it at https://github.com/OCA/server-ux/pull/329, if you want to apply locally you can just add a `parseInt`.

PR #329, opened 2021-04-15
[FIX] date_range: make the range selector work

len-foss:
Because ev.target.value is a string and the range ids are integers,
comparison was always false, so the first range values were always used.

Fixes #329


len-foss:
Hello @fuentes010 
Could you review this? I don't talk JS so there may be changes to make, but I think this solves the issue https://github.com/OCA/server-ux/issues/328.

fuentes73 (approved):

CasVissers-360ERP:
@len-foss 
Thanks a lot!

PR #334, opened 2021-04-25
[14.0][FIX] base_cancel_confirm, access window action error

kittiu:
This is a small fix, can be FFW.


StefanRijnhart (approved):
Thanks! I was not entirely aware myself that permissions on the action models were tightened in Odoo 14.0

PR #335, opened 2021-04-29
[12.0][IMP] date_range: add a search mixin class

StefanRijnhart:
To configure a model to use the Many2one style search field, make the model
inherit from `date.range.search.mixin`:

```
    class AccountMove(models.Model):
        _name = "account.move"
        _inherit = ["account.move", "date.range.search.mixin"]
```

This will make a `Period` field show up in the search view:

period

By default, the mixin works on the `date` field. If you want the mixin to work
on a field with a different name, you can set a property on your model:

```
   _date_range_search_field = "invoice_date"
```

By default, the mixin works on the `date` field. If you want the mixin to work
on a field with a different name, you can set a property on your model:

.. code-block::

   _date_range_search_field = "invoice_date

I added this `code` field to the `date.range.type` model to make this type selection less muddled than it would have been by `name` which is translatable and seems more volatile to me due to its cosmetic nature.


date_range/README.rst
```diff
@@ -28,6 +28,9 @@ Date Range
 This module lets you define global date ranges that can be used to filter
 your values in tree views.
 
+It also provides a mixin model for developers that extends the model's search
+view so that date ranges can be search as any relational field.
```
hbrunn:
```suggestion
view so that date ranges can be searched as any relational field.
```

date_range/models/date_range_search_mixin.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+        groups = root.xpath("/search/group")
+        if groups:
+            groups[0].addprevious(separator)
+            groups[0].addprevious(field)
+        else:
+            search = root.xpath("/search")
+            search[0].append(separator)
+            search[0].append(field)
+        result['arch'] = etree.tostring(root, encoding='unicode')
```
hbrunn: shouldn't this be utf8? or actually nothing?

date_range/models/date_range_search_mixin.py
```diff
@@ -0,0 +1,88 @@
...
+
+    @api.model
+    def fields_view_get(
+            self, view_id=None, view_type='form', toolbar=False,
+            submenu=False):
+        """Inject the dummy Many2one field in the search view"""
+        result = super().fields_view_get(
+            view_id=None, view_type=view_type, toolbar=toolbar)
+        if view_type != "search":
+            return result
```
hbrunn: just to be sure do nothing too if the field already exists in the view?

date_range/models/date_range_search_mixin.py
```diff
@@ -0,0 +1,88 @@
...
+            search[0].append(separator)
+            search[0].append(field)
+        result['arch'] = etree.tostring(root, encoding='unicode')
+        result["fields"].update(self.fields_get("date_range_search_id"))
+        return result
+
+    date_range_search_id = fields.Many2one(
+        comodel_name="date.range",
+        string="Filter by period (technical field)",
+        compute="compute_date_range_search_id",
```
hbrunn:
```suggestion
        compute="_compute_date_range_search_id",
```

and below

date_range/models/date_range_search_mixin.py
```diff
@@ -0,0 +1,88 @@
...
+        groups = root.xpath("/search/group")
+        if groups:
+            groups[0].addprevious(separator)
+            groups[0].addprevious(field)
+        else:
+            search = root.xpath("/search")
+            search[0].append(separator)
+            search[0].append(field)
+        result['arch'] = etree.tostring(root, encoding='unicode')
+        result["fields"].update(self.fields_get("date_range_search_id"))
```
hbrunn: I *think* you can set searchable to false here in order to suppress the field in extended search without impeding normal search

date_range/tests/common.py
```diff
@@ -0,0 +1,19 @@
...
+    env.registry.init_models(
+        env.cr, [model_cls._name for model_cls in model_clses],
+        dict(env.context, update_custom_fields=True)
+    )
+
+
+def teardown_test_model(env, model_clses):
+    for model_cls in model_clses:
+        del env.registry.models[model_cls._name]
+    env.registry.setup_models(env.cr)
```
hbrunn: doesn't seem to work as well as i thought: https://travis-ci.com/github/OCA/server-ux/jobs/501959634#L986

hbrunn (changes requested):
I think an extra module that allows to configure a set of fields on date range types and does this dynamically based on this code would be nice, but that's a thing for another future PR

date_range/models/date_range_search_mixin.py
StefanRijnhart: Copied mindlessly from Odoo's `ir.ui.view`, so should be as good as anything.

date_range/tests/common.py
StefanRijnhart: Committed a potential fix already.

date_range/models/date_range_search_mixin.py
StefanRijnhart: OK, done

date_range/models/date_range_search_mixin.py
StefanRijnhart: What would be the benefit of that? Keeping GUI consistency in mind.

date_range/models/date_range_search_mixin.py
StefanRijnhart: Those dang guidelines and conventions! Taken.

date_range/tests/common.py
StefanRijnhart: Had a lot of potential, now working on a more effective fix.

date_range/tests/models.py
```diff
@@ -0,0 +1,10 @@
+from odoo import fields, models
+
+
+class TestDateRangeSearchMixin(models.Model):
```
hbrunn, line 4:
```suggestion
class TestDateRangeSearchMixin(models.TransientModel):
    _auto = False
```

hbrunn: hmmno, the `_auto` defeats the purpose, but TransientModel is helpful

date_range/models/date_range_search_mixin.py
hbrunn: then I find the (Technical Field) part superfluous

StefanRijnhart: Oh does that show up there. The technical field label seems useful to me when creating exports.

date_range/tests/models.py
StefanRijnhart: Fixed it by cleaning out _inherit_children even more aggressively.

date_range/models/date_range_search_mixin.py
StefanRijnhart: Fixed by overwriting the label in `load_views` (where, I'm sure, we could also reset searchable if we still want to)

date_range/tests/common.py
```diff
@@ -0,0 +1,28 @@
...
+        del env.registry.models[model_cls._name]
+        inherits = ["base"] + list(model_cls._inherits) + (
+            model_cls._inherit if isinstance(model_cls._inherit, list)
+            else [model_cls._inherit])
+        for inherited_cls in inherits:
+            if model_cls._name in env.registry.models[
+                    inherited_cls]._inherit_children:
+                env.registry.models[
+                    inherited_cls]._inherit_children.remove(
+                        model_cls._name)
```
hbrunn: why not just iterate through all models then?

StefanRijnhart: Isn't this the more educated solution?

hbrunn: as it works, don't change it, no

hbrunn (approved):

hbrunn:
maybe you can improve coverage

StefanRijnhart:
> maybe you can improve coverage

Hit the target now.

HaraldPanten (approved):

date_range/tests/test_date_range_search_mixin.py
```diff
@@ -0,0 +1,174 @@
...
+from odoo.tests import SavepointCase
+from .common import setup_test_model, teardown_test_model
+from .models import TestDateRangeSearchMixin
+
+
+class TestDateRangeearchMixin(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        setup_test_model(cls.env, [TestDateRangeSearchMixin])
```
lmignon: @StefanRijnhart You could use [odoo-test-helper](https://pypi.org/project/odoo-test-helper/) to build your test model.

StefanRijnhart: Looks good! How do I deal with the additional library dependency? It would be awkward to add a new dependency to the module manifest in such an old release, but if I only import it in tests such that it does not affect runtime code, will it be picked up by pre-commit hook for inclusion in requirements.txt?

lmignon: You just need to put the dependency into the `test-requirements.txt` file at the root of the repo. No need to add the dependency into the manifest. (ex https://github.com/OCA/search-engine/blob/14.0/test-requirements.txt)

StefanRijnhart: Perfect, will do!

hbrunn:
@StefanRijnhart is this ready for review (or actually merge) now?

StefanRijnhart:
No, I want to add one more feature.

StefanRijnhart:
I added a function to actually assign a date range when reading the field, from a specific date range type that is selected by a new `code` field I'm adding to the `date.range.type` data model. Our users asked for a meaningful value of the field so that they could create pivot tables after exporting the records to Excel. I hope that's alright. Better rereview @HaraldPanten and @hbrunn.
Separate commit for easier review, but I can squash after approval.

date_range/models/date_range_search_mixin.py
```diff
@@ -20,9 +25,35 @@ class DateRangeSearchMixin(models.AbstractModel):
...
+                    JOIN date_range AS dr
+                        ON {} BETWEEN dr.date_start AND dr.date_end
+                    WHERE dr.type_id = %s AND tbl.id IN %s
+                    """).format(
+                        sql.Identifier(self._table),
+                        sql.Identifier(self._date_range_search_field),
+                    )
+                self.env.cr.execute(
+                    query,
+                    (dr_type.id, tuple(self.ids or [0])),
```
hbrunn: this can be more intelligible with [AsIs](https://www.psycopg.org/docs/extensions.html#psycopg2.extensions.AsIs)

date_range/models/date_range_search_mixin.py
```diff
@@ -20,9 +25,35 @@ class DateRangeSearchMixin(models.AbstractModel):
...
+                )
+                record2range = dict(self.env.cr.fetchall())
         for record in self:
-            record.date_range_search_id = False
+            record.date_range_search_id = record2range.get(record.id, False)
```
hbrunn: what's the difference between `None` and `False` here?

date_range/models/date_range_search_mixin.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+class DateRangeSearchMixin(models.AbstractModel):
+    _name = "date.range.search.mixin"
+    _description = "Mixin class to add a Many2one style period search field"
+    # Override this attribute to apply the mixin to another field
+    _date_range_search_field = "date"
+    # Set this attribute to assign a date range of the type with this code.
+    # This has to be a date range type without company set
+    _date_range_assign_type_code = None
```
hbrunn: for making this perfectly general, can't we assume `company_id` on the record and allow to fine tune the field name the way we do it here?

hbrunn (commented):
I'm not sure if this should go to the general module and not to the customer specific part

StefanRijnhart:
> I'm not sure if this should go to the general module and not to the customer specific part

Agreed. Let's hear from @lmignon for instance.

date_range/models/date_range_search_mixin.py
StefanRijnhart: What do you mean exactly?

date_range/models/date_range_search_mixin.py
StefanRijnhart: Oh if None works I'll do that.

date_range/models/date_range_search_mixin.py
StefanRijnhart: What do you mean? I think having a company on the date range type is a problem because the attribute by necessity lacks a company context.

hbrunn: but you can join the company field on the table to the range's company_id in the statement below, right?

date_range/models/date_range_search_mixin.py
hbrunn:
```python
query,
(dr_type.id, tuple(self.ids or [0]), AsIs(table), etc),
```

StefanRijnhart: If you mean, use `AsIs` instead of `sql.SQL`, I believe `AsIs` does not really help against SQL injection so I think `sql.SQL` should be used as much as possible for cases like these.

date_range/models/date_range_search_mixin.py
StefanRijnhart: I'm just thinking that setting a company_id on date ranges by default was not the best design to begin with, so I am hesitant to go along with it. Or is that my lack of imagination?

date_range/models/date_range_search_mixin.py
hbrunn: you don't need to protect against sql injection from developers using your code

date_range/models/date_range_search_mixin.py
hbrunn: I'd think of a join condition like `{company_id_field} = res_company.id or {company_id_field} is null`

date_range/models/date_range_search_mixin.py
StefanRijnhart: Like I said, it's good practice.

date_range/models/date_range_search_mixin.py
StefanRijnhart: Yes, I will add that if the feature is accepted.

dreispt:
Is this ready to merge?
@lmignon do you want to add any comment?

StefanRijnhart:
To reiterate my question to @lmignon: do you think the functionality of the second commit (assign a date range to entries of a model based on some date field, mainly for grouping outside of Odoo using exported data) should be implemented in this module, or rather as a customization for this particular customer.
If no, I will remove the second commit, if yes, I will work on implementing multicompany support as suggested by @hbrunn to make this feature more generically useful.

lmignon:
@StefanRijnhart Thank you for all this work. The functionality you put add into you second commit is not clear for me when reading the code. It could be at least put into an other PR so we can merge the first part of your changes.

StefanRijnhart:
@lmignon thanks, I removed the second commit.

lmignon (approved):

lmignon:
@hbrunn Is-it OK for you?

PR #158, opened 2020-05-14
WIP [13.0][MIG] base_user_locale

Tonow-c2c:
WIP :hammer: 

- [x] check what should I do change to tackle this comment
- [x] check https://github.com/OCA/server-ux/pull/158/files#diff-4b4975f36e2ac07b8f87ba79597fd23bR21-R23 about https://github.com/odoo/odoo/pull/36532
- [ ] test it

PR #337, opened 2021-04-29
[13.0][MIG] base_user_locale

MiquelRForgeFlow:
Supersedes https://github.com/OCA/server-ux/pull/158.

Tests are a bit refactored due to https://github.com/odoo/odoo/pull/38357.


LoisRForgeFlow (approved):
functional review in local, LGTM :+1:

base_user_locale/controllers/web_client.py
```diff
@@ -0,0 +1,41 @@
...
+
+from odoo.http import request, route
+
+from odoo.addons.web.controllers.main import WebClient, ensure_db
+
+
+class WebClient(WebClient):
+    def get_user_lang_parameters(self, user):
+        res = {}
+        date_format = user.date_format or user.company_id.date_format
```
LoisRForgeFlow: all these `user.company_id` should be `user.env.company`, isn't it?

MiquelRForgeFlow: right

AaronHForgeFlow (approved):
Tested on runbot and working fine. Nevertheless, it would be great a configuration section in the readme. Currently I have to go to the code in order to know how to configure things.

On the roadmap perhaps we can include a help section so the users can configure their own settings on their own.

LoisRForgeFlow:
Readme updated! thanks for your review :)

/ocabot merge nobump

PR #338, opened 2021-05-04
[13.0][IMP] base_tier_validation: comment on rejection

JordiBForgeFlow:
Add the possibility to request feedback after validation only when a tier
validation has been rejected. The tier.definition now has a boolean
'comment_on_rejection'.

image

@ForgeFlow @GenisPForgeFlow

### DIFF
--- a/base_tier_validation/models/tier_definition.py
+++ b/base_tier_validation/models/tier_definition.py
@@ -57,11 +57,15 @@ class TierDefinition(models.Model):
         help="If set, all possible reviewers will be notified by email when "
         "this definition is triggered.",
     )
-    has_comment = fields.Boolean(string="Comment", default=False)
     comment_option = fields.Selection(
-        [("all", "Both Validation & Rejection"),
-         ("validate", "Validation Only"),
-         ("reject", "Rejection Only")], string="Comment only on rejection", default='all',
+        [
+            ("none", "No Comment"),
+            ("all", "Both Validation & Rejection"),
+            ("validate", "Validation Only"),
+            ("reject", "Rejection Only"),
+        ],
+        string="Comment after review",
+        default="none",
     )
     approve_sequence = fields.Boolean(
         string="Approve by sequence",
--- a/base_tier_validation/models/tier_definition.py
+++ b/base_tier_validation/models/tier_definition.py
@@ -58,8 +58,10 @@ class TierDefinition(models.Model):
         "this definition is triggered.",
     )
     has_comment = fields.Boolean(string="Comment", default=False)
-    comment_on_rejection = fields.Boolean(
-        string="Comment only on rejection", default=False
+    comment_option = fields.Selection(
+        [("all", "Both Validation & Rejection"),
+         ("validate", "Validation Only"),
+         ("reject", "Rejection Only")], string="Comment only on rejection", default='all',
     )
     approve_sequence = fields.Boolean(
         string="Approve by sequence",
--- a/base_tier_validation/models/tier_definition.py
+++ b/base_tier_validation/models/tier_definition.py
@@ -58,6 +58,9 @@ class TierDefinition(models.Model):
         "this definition is triggered.",
     )
     has_comment = fields.Boolean(string="Comment", default=False)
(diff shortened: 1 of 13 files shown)


JordiBForgeFlow:
@kittiu @Saran440 what do you think about this one?

kittiu:
> @kittiu @Saran440 what do you think about this one?

This is feature we also need, thank you!

By the way, IMHO, the better is to be setting True to what we need (so it can add things in more symmetry), i.e., https://github.com/OCA/server-ux/pull/324/files

But that would need changes on fields, CON is the need of migration script.
* has_comment_on_validation
* has_comment_on_rejection

Another option though, is to change `comment_on_rejection` to a selection `comment_option` field, i.e.,
```
[("all", "Both Validation & Rejection"),
 ("validate", "Validation Only"),
 ("reject", "Rejection Only")]
```

My 2 cents.

GenisPForgeFlow (approved):

etobella:
We could remove also the has_comment field.
It will not have comment if comment_options is False. WDYT?

base_tier_validation/models/tier_definition.py
```diff
@@ -58,8 +58,10 @@ def _get_tier_validation_model_names(self):
         "this definition is triggered.",
     )
     has_comment = fields.Boolean(string="Comment", default=False)
-    comment_on_rejection = fields.Boolean(
-        string="Comment only on rejection", default=False
+    comment_option = fields.Selection(
+        [("all", "Both Validation & Rejection"),
+         ("validate", "Validation Only"),
+         ("reject", "Rejection Only")], string="Comment only on rejection", default='all',
```
etobella: Why don't you make default=False and remove the has_comment field (or make it computed and storable :smile:) Remember to add  a migration script in this case :smile:

LoisRForgeFlow: :+1: The default behavior should be no comment as it was before.

kittiu: Should we remove the has_comment field? It seem no longer needed.

kittiu: Also, this change has affect the `base_tier_validation_forward`'s as it is on top of this. I am ready to apply the change after this one is merged.

But how can we pass travis, i.e., should I remove the test from `base_tier_validation_forward` at the moment? Any idea?

JordiBForgeFlow:
@kittiu @etobella thanks for your feedback. I have removed the attribute 'has_comment', and we now only have 'comment_option'. Added also migration scripts, and I included an extra commit that fixes the base_tier_validation_forward.

etobella:
@JordiBForgeFlow some has_comment is missing on base_tier_validation_forward, also, pre-commit is not happy

JordiBForgeFlow:
@kittiu I am not very confortable with the way the base_tier_validation_forward works by setting a value from a wizard to a computed, stored field. Any time this field si recomputed (perhaps in a migration) the data will be erased. Perhaps the values from the wizard such as https://github.com/OCA/server-ux/blob/13.0/base_tier_validation_forward/wizard/forward_wizard.py#L38 should be stored in the tier.review on a separate field, and then the computed field uses them if they are not null.

kittiu:
> @kittiu I am not very confortable with the way the base_tier_validation_forward works by setting a value from a wizard to a computed, stored field. Any time this field si recomputed (perhaps in a migration) the data will be erased. Perhaps the values from the wizard such as https://github.com/OCA/server-ux/blob/13.0/base_tier_validation_forward/wizard/forward_wizard.py#L38 should be stored in the tier.review on a separate field, and then the computed field uses them if they are not null.

@JordiBForgeFlow since we need to create tier.review on the fly (without definition_id), should we have a boolean field like `manual = True` or `forward = True` to tell function `_compute_definition_data` not to recompute?

**Note:** or even no need for the new field and just use `definition_id = False` to not recompute, i..e,

```
    def _compute_definition_data(self):
        for rec in self.filtered("definition_id"):   # <--
            rec.name = rec.definition_id.name
            rec.review_type = rec.definition_id.review_type
            ....
```

WDYT?

PR #339, opened 2021-05-04
[FIX] base_revision: auto_search and no_destroy are no longer valid attributes

dreispt:


dreispt:
Ping!

kittiu (approved):
LGTM, thank you for your review.

PR #343, opened 2021-05-13
[13.0][IMP] base_tier_validation: Remove tier review when model state goes to cancel

sergio-teruel:
If I have a review for a sale order and user cancel it for ever the review is pendig for ever.

cc @Tecnativa TT29643
ping @chienandalu @CarlosRoca13

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -211,7 +211,7 @@ class TierValidation(models.AbstractModel):
                 and not self._check_allow_write_under_validation(vals)
             ):
                 raise ValidationError(_("The operation is under validation."))
-        if vals.get(self._state_field) in self._state_from:
+        if vals.get(self._state_field) in (self._state_from + [self._cancel_state]):
             self.mapped("review_ids").unlink()
         return super(TierValidation, self).write(vals)
 


base_tier_validation/models/tier_validation.py
```diff
@@ -211,7 +211,7 @@ def write(self, vals):
                 and not self._check_allow_write_under_validation(vals)
             ):
                 raise ValidationError(_("The operation is under validation."))
-        if vals.get(self._state_field) in self._state_from:
+        if vals.get(self._state_field) in [self._state_from, self._cancel_state]:
```
pedrobaeza: First team is a list, so you have to merge both things:
```suggestion
        if vals.get(self._state_field) in (self._state_from + [self._cancel_state]):
```

sergio-teruel:
@pedrobaeza Changes done!!

pedrobaeza (approved):

chienandalu (approved):

kittiu:
@sergio-teruel we have got another bug based on this change. https://github.com/OCA/server-ux/pull/447
May I know why it is not OK having the review table showing on Cancelled document (I rather think it is good as it is the evidence of rejection).

Note: we are trying to fix the problem on top of this fix, but it seem harder than we think given the review was removed.

rousseldenis:
> @sergio-teruel we have got another bug based on this change. #447 May I know why it is not OK having the review table showing on Cancelled document (I rather think it is good as it is the evidence of rejection).
> 
> Note: we are trying to fix the problem on top of this fix, but it seem harder than we think given the review was removed.

I would say it would be nice to have a test for this change. @sergio-teruel

PR #346, opened 2021-05-17
[13.0][FIX] base_tier_validation: invalid review_user_count()

acsonefho:
In some case, when the validation process is not really well processed (user click on standard 'cancel' button instead of 'reject', link between the record and reviewers still present and it's counted into the total of pending review instead of ignored them.

**How to reproduce:**
Just create a validation process on `purchase.order`.
Instead of validate the PO, click on the cancel button.
The user still have 1 PO to review (the review that we just cancel). If you click on the button, you'll be redirect to an empty PO tree view as the PO has been cancelled before.

**The fix:**
Now with the fix, PO (or others records) without validation needed are ignored

### DIFF
--- a/base_tier_validation/models/res_users.py
+++ b/base_tier_validation/models/res_users.py
@@ -26,7 +26,12 @@ class Users(models.Model):
                 .with_user(self.env.user)
                 .search([("id", "=", review.res_id)])
             )
-            if not record or record.rejected or not record.can_review:
+            if (
+                not record
+                or record.rejected
+                or not record.can_review
+                or not record.need_validation
+            ):
                 # Checking that the review is accessible with the permissions
                 # and to review condition is valid
                 continue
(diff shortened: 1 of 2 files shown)


Cedric-Pigeon (approved):
Code review

LoisRForgeFlow:
I think this other proposal was trying to fix the same issue: https://github.com/OCA/server-ux/pull/343. Unfortunately, it got merged before I was able to review it. What do you think of that other proposal? is it good for you to remove reviews in cancel state or you want/need to keep the reviews?

StefanRijnhart:
I've implemented similar functionality in 2017 and we made the design choice to unlink reviews on cancelling there.

acsonefho:
I confirm that PR #343 fix the issue. But a migration script is missing to fix existing data.
There is an example of a migration script to do on `purchase.order`
```
from odoo import SUPERUSER_ID, api
import logging

_logger = logging.getLogger(__name__)



def migrate(cr, version):
    if not version:
        return
    _logger.info("Start migration script of purchase validation")
    with api.Environment.manage():
        env = api.Environment(cr, SUPERUSER_ID, {})
        domain = [('state', '=', 'cancel'), ('review_ids', '!=', False)]
        purchases = env['purchase.order'].search(domain)
        _logger.info("These purchase orders will be re-updated to the cancel "
                     "state because they still have reviewers. IDs: %s",
                     tuple(purchases.ids))
        purchases.write({'state': 'cancel'})
    _logger.info("Migration script finished")
```

PR #347, opened 2021-05-19
[12.0][IMP] date_range: fully automatic generation from date range types

StefanRijnhart:
Configure defaults for the generation wizard on the date range type. Also add configuration for autogenerating ranges from cron. Introduce an optional, complex expression to generate date range names like `2021Q3`.

image

image


date_range/models/date_range_type.py
```diff
@@ -42,3 +71,42 @@ def _check_company_id(self):
...
+        """Wipe the prefix if an expression is entered.
+
+        The reverse is not implemented because we don't want to wipe the
+        users' painstakingly crafted expressions by accident.
+        """
+        if self.name_expr and self.name_prefix:
+            self.name_prefix = False
+
+    def test_name_expr(self):
+        """Button on the form to preview newly generated names"""
```
lmignon: @StefanRijnhart I would be nice to have a computed field `name_preview` displayed next to the `name_expr` to always display the preview without having to push a button. Moreover, if the expression is wrong, the UserError will prevent to store the value at save time. If you make this new field `store=True` you'll ensure that this implicit validation will always be applied even if `name_expr` is modified into the code.

date_range/models/date_range_type.py
```diff
@@ -42,3 +71,42 @@ def _check_company_id(self):
...
+                [("autogeneration_count", "!=", False),
+                 ("autogeneration_unit", "!=", False)]):
+            try:
+                with self.env.cr.savepoint():
+                    form = Form(self.env["date.range.generator"])
+                    form.type_id = dr_type
+                    wizard = form.save()
+                    wizard.action_apply()
+            except (AssertionError, UserError, ValidationError):
+                pass
```
lmignon, line 128: We should at least log an exception...

date_range/models/date_range_type.py
```diff
@@ -42,3 +71,42 @@ def _check_company_id(self):
...
+
+    @api.model
+    def autogenerate_ranges(self):
+        """Generate ranges for types with autogeneration settings"""
+        for dr_type in self.search(
+                [("autogeneration_count", "!=", False),
+                 ("autogeneration_unit", "!=", False)]):
+            try:
+                with self.env.cr.savepoint():
+                    form = Form(self.env["date.range.generator"])
```
lmignon, line 117: PLZ a Form is not business logic... we must call the wizard model...

date_range/models/date_range_type.py
StefanRijnhart: I considered that, but those errors are usually related to no new periods needed to be created.

date_range/wizard/date_range_generator.py
```diff
@@ -26,37 +35,162 @@ def _default_company(self):
...
+            elif name_prefix:
+                names.append(name_prefix + index)
+            else:
+                raise ValidationError(
+                    _("Please set a prefix or an expression to generate "
+                      "the range names."))
+        return names
+
+    def test_name_expr(self):
+        """Button on the form to preview the generated names"""
```
lmignon: same comment as https://github.com/OCA/server-ux/pull/347/files#r635903610

date_range/models/date_range_type.py
StefanRijnhart: I was shown this trick by @pedrobaeza. This code shows exactly how elegant it can work in business logic.

pedrobaeza: Yes, but it seems @lmignon is totally against it. I still prefer to use it although we lose a bit of performance, which is not critical here.

lmignon: Yes I'm totally against it since this breaks with a fundamental principle in programming: business logic should not depend on presentation logic.

pedrobaeza: That's not really true in Odoo, as onchange affects business logic.

jgrandguillaume:
Hi here,

Just to say I fully support this nice idea :) Thank you !

date_range/models/date_range_type.py
StefanRijnhart: The integration of onchange with compute methods in core Odoo very much reflects this. Isn't `Form` a good tool to use in the mean time until we've all switched over?

date_range/wizard/date_range_generator.py
StefanRijnhart: In this case, the button returns all names to generate, with the accurate zero fill on the index. I think that unlike the case on the date range type, we lose value with your suggestion here.

StefanRijnhart: Oh maybe I can just run that code in the compute method.

date_range/models/date_range_type.py
lmignon: In both cases we've the same number of lines...

```python

with self.env.cr.savepoint():
    wizard = self.env["date.range.generator"].create({"type_id": dr_type.id})
    wizard._onchange_type_id()
    wizard.action_apply()
``` 

IMO the code is even more readable.

pedrobaeza: But if you have extension modules that adds other onchanges, this is not working. Not only that, the code you propose don't trigger chained onchanges.

date_range/models/date_range_type.py
lmignon: Can we just ignore the specific error "No ranges to generate with these settings" ? You could create a typed error for this one only that inherits from UserError to ease to ignore it. Or you could pass a paramter to the action_apply method to ignore this specific error... What do you think about?

StefanRijnhart: Yes, alright.

StefanRijnhart: Improved the error handling and ignored 'No ranges to generate' message in batch mode. Thank you for the suggestion!

date_range/models/date_range_type.py
StefanRijnhart: Done now. Screenshots updated in top comment. Thanks for suggesting!

date_range/wizard/date_range_generator.py
StefanRijnhart: Done

date_range/models/date_range_type.py
StefanRijnhart: Given that there is no consensus, can we agree to disagree, and then keep the current version based on my preference as the contributing developer. Every day I feel the pain of disagreeing entirely with some of the choices made by other developers in the OCA or with some of the OCA conventions, but that's the price of collaboration.

lmignon: Here, extension is not a problem. If we want we can add extension hook, or we could also use [onhange_helper](https://github.com/OCA/server-tools/tree/12.0/onchange_helper) . As OCA, we want to promote the best way of doing things. We should also highlight what's wrong into Odoo and educate people how to write code to follow the best practices. I remain against the use of code designed to be used in tests and emulate the UI to provide business logic in my own business logic. Your proposed functional enhancement is great but I stay convinced that these 3 lines of code should be changed.

pedrobaeza: I support Stefan's PoV.

StefanRijnhart: We can only promote the best way when we agree on what that is. If (or while) we have different ideas about that, then in cases like this it is arbitrary which way is used for the moment.

lmignon: Do you agree that the use of presentation logic to provide business logic is a is nonsense? I'll not block the merge of this PR nor approve it. Can you at least add a comment with a TODO to specify that this code must be rewritten in odoo >= 13.0 to directly call the wizard and replace onchange methods into the wizard by computed fields declared with read_only=False?

StefanRijnhart: @lmignon I'm afraid that is exactly what we disagree on. I, like many other Odoo developers have ignored onchange functionality way for too long, introducing subtle bugs and inconsistencies between front end and backend. The `Form` class provided by Odoo solves all of this in a way that is working really well at least in this particular case.
I thought that stored, read-write computed fields in Odoo 14.0 would replace that, and I gave it a try in the 14.0 port of this PR (#349). I have successfully replaced the onchange functionality that I introduce in this PR, but I can't seem to drop the use of `Form`, because if I try to create my wizard using `self.env["date.range.generator"].create({"type_id": dr_type.id})`, Postgresql complaints about integrity errors related to non-null columns in the wizard's table before Odoo executes the compute methods that populates the fields. Do you know how to fix that?

lmignon: @StefanRijnhart You should remove `required=True` on a computed field. AFAIK, it's not possible to declare a compute field as required.

StefanRijnhart: But then we are sacrificing the data model in order to avoid what you call using presentation logic. That can't be right, can it?

lmignon: @StefanRijnhart When you use a wizard from the business)logic side you don't need to create a real record into the DB. You should change of paradigm and use an in memory record `wiz = self.env["date.range.generator"].new({"type_id": dr_type.id})`. By using an in memory record you avoid useless SQL queries and sql constrains are not applied. As an Odoo developer we are conditioned to always think and design our code so that it can be extended via the exclusive mechanism of inheritance. As this inheritance mechanism only applies to recordsets, we end up creating temporary records in DB to execute logic when there is no need to persist this data. All this logic could be executed by a simple python object initialized with the necessary data but in this case, we lose the ability to allow others to extend this logic... By using in memory record, we mix the two worlds....

StefanRijnhart: @lmignon perfect, thank you for this guidance. You are absolutely right, I had forgotten about `new()` and it works very well with the computed fields in the version 14.0 of this PR https://github.com/OCA/server-ux/pull/349/files#diff-fff6c0ce2e913bda32693527150b8380694e82e36dc7c3f900f7c50dbf951d6eR139-R144.
I don't know if there is any reason why that would not work in 12 as well so I'll try and refactor this PR likewise.

StefanRijnhart: Looks like the compute methods are not triggered when working with `NewId`s in Odoo 12, so that would mean keeping Form in this version.

pedrobaeza (approved):

pedrobaeza:
This is still interesting to be added. What is missing?

lmignon (approved):

lmignon:
I'm still not for the use of `Form`  into the business logic. I would prefer to use the same mechanism as it's done for sale.order.line int odoo core to fill missing values not provided at creation if the code is called from server side. https://github.com/odoo/odoo/blob/15.0/addons/sale/models/sale_order_line.py#L146

PR #348, opened 2021-05-26
[13.0][IMP] mass_editing: improvement to apply update

acsonefho:
**Improvements:**
- mass_operation_abstract: let the possibility to use the `active_domain` (instead of always `active_ids`);
- mass_edition unit test: disable tracking (minor improvement);
- mass_edition code: split the update of target recordset into a new function to have an inherit point;
- mass_edition code: split the create to avoid mix values of the wizard and values of target records.

**Why using active_domain?**
In some case, when you select all items (the dedicated select-box item), Odoo fill `active_ids` with only the 80 items displayed into the tree view (even if you have thousands of records). And sometimes it's not possible to display all of them because there is too much records.

### DIFF
--- a/mass_editing/tests/test_mass_editing.py
+++ b/mass_editing/tests/test_mass_editing.py
@@ -5,6 +5,7 @@
 # import ast
 
 from odoo.modules import registry
+from odoo.osv import expression
 from odoo.tests import common
 
 from ..hooks import uninstall_hook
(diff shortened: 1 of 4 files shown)


mass_editing/tests/test_mass_editing.py
```diff
@@ -16,7 +16,7 @@ class TestMassEditing(common.SavepointCase):
 
     def setUp(self):
```
simahawk: as you are here... can you move it to SavepointCase and do this in setUpClass?

Cedric-Pigeon (approved):

mass_editing/wizard/mass_editing_wizard.py
```diff
@@ -3,14 +3,19 @@
...
-from odoo import _, api, models
+from odoo import _, api, fields, models
 
 
 class MassEditingWizard(models.TransientModel):
     _name = "mass.editing.wizard"
     _inherit = "mass.operation.wizard.mixin"
     _description = "Wizard for mass edition"
 
+    use_active_domain = fields.Boolean(
```
StefanRijnhart, line 14: This is the technical term, which might not make a lot of sense to the users. Maybe you could reword using something like 'search results'?

StefanRijnhart:
A test for the active domain functionality would be good, and would probably make codecov happy as well.

simahawk (changes requested):
can you check @StefanRijnhart comment as well?

simahawk (approved):

simahawk:
/ocabot merge minor

will you port it to v14?

simahawk:
> /ocabot merge minor
> 
> will you port it to v14?

nevermind, I linked this to v14 mig https://github.com/OCA/server-ux/pull/298 -> can you review it?

dzungtran89:
ok @simahawk , I'll port the update to v14

PR #342, opened 2021-05-12
[14.0] [MIG] base_import_security_group: Migrated the module in v14.

Chandresh-OSI:

PR #351, opened 2021-05-31
[14.0][MIG] base_import_security_group

StefanRijnhart:
Continued from https://github.com/OCA/server-ux/pull/342 by @Chandresh-SerpentCS

Additional changes:

* Patch existing class instead of replacing it, for better compatibility with other modules inheriting from the same class.
* To circumvent problems executing async calls in a non-widget class, pass the user's import permission through the web session.
* Adapt tests to consult he `Favorites` dropdown in the 14.0 GUI.
* Rebased according to OCA conventions.
* In the tour, don't target the `Users` menu which is the default menu of the `Settings` menu when no other module is installed, and in that case clicking the menu in the tour is a void action that makes the tests fail. Target the `Companies` menu item instead.


StefanRijnhart:
~~Darn. I missed that Odoo skips the javascript tests, which fail here. Having a look.~~ Tests were not adapted to the situation in 14.0. To further improve inheratibility of the `ImportMenu` class I refactored the javascript so that it is patched instead of replaced by this module.

base_import_security_group/static/src/js/import.js
```diff
@@ -0,0 +1,24 @@
...
+odoo.define("base_import_security_group.group_import", function (require) {
+    "use strict";
+
+    const ImportMenu = require("base_import.ImportMenu");
+    const shouldBeDisplayed_orig = ImportMenu.shouldBeDisplayed;
+
+    ImportMenu.shouldBeDisplayed = function (env) {
+        return (
+            shouldBeDisplayed_orig(env) &&
+            this.env.session.base_import_security_group__allow_import === 1
```
PieterPaulussen: also valid apparently: `env.session...` instead of `this.env`
But I do not really know what difference it would make.

PieterPaulussen (approved):

base_import_security_group/static/src/js/import.js
StefanRijnhart: You're right! `this.env` seems to be a given in owl classes but if we get handed an `env` we better use that one.

dsolanki-initos (approved):
LGTM. Code review

dsolanki-initos:
It would be great if the maintainers could merge it

PR #354, opened 2021-06-08
[12.0] FIX date_range Expected singleton while using advanced filter

eLBati:
Steps:

 - Create a new DB
 - Install date_range and account
 - Go to customer invoices and add an advanced filter on "invoice date" selecting "in month"

Get

```
  server-ux/date_range/models/date_range.py", line 91, in get_domain
    self.ensure_one()
  odoo/odoo/models.py", line 4768, in ensure_one
    raise ValueError("Expected singleton: %s" % self)
```

### DIFF
--- a/date_range/static/src/js/date_range.js
+++ b/date_range/static/src/js/date_range.js
@@ -100,19 +100,22 @@ class
         on_range_selected: function (e) {
             var self = this;
             self.domain = '';
-            framework.blockUI();
-            return rpc.query({
-                args: [this.get_value()],
-                kwargs: {
-                    field_name: this.field.name,
-                },
-                model: 'date.range',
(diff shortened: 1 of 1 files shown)


mrcast (approved):
Functional test ok!

StefanRijnhart:
Did you consider fixing the bug in the javascript instead? It seems to me that it might be better to prevent the call here: https://github.com/OCA/server-ux/blob/12.0/date_range/static/src/js/date_range.js#L105 when `this.get_value()` does not return a valid id.

eLBati:
@StefanRijnhart it seems better, thanks. Modified

StefanRijnhart (approved):
Perfect, thanks!

PR #358, opened 2021-06-27
[14.0][IMP] base_tier_validation_server_action, use write to trigger action

kittiu:
By using write() instead of @api.depends, it is better to accept context value for some situation.

This can be FFW.

### DIFF
--- a/base_tier_validation_server_action/models/tier_review.py
+++ b/base_tier_validation_server_action/models/tier_review.py
@@ -1,20 +1,24 @@
 # Copyright 2020 Ecosoft (http://ecosoft.co.th)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import api, models
+from odoo import models
 
 
 class TierReview(models.Model):
     _inherit = "tier.review"
 
-    @api.constrains("status")
-    def _trigger_server_action(self):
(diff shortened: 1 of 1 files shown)


newtratip (approved):
LGTM :+1:

base_tier_validation_server_action/models/tier_review.py
```diff
@@ -1,20 +1,24 @@
...
-                ctx = {"active_model": rec.model, "active_id": rec.res_id}
-                server_action.with_context(ctx).run()
+    def write(self, vals):
+        res = super().write(vals)
+        if vals.get("status") in ["approved", "rejected"]:
+            for rec in self:
+                server_action = False
+                if rec.status == "approved":
+                    server_action = rec.definition_id.server_action_id
+                if rec.status == "rejected":
```
Saran440, line 17:
```suggestion
                elif rec.status == "rejected":
```
if status is approved, it can't be rejected.

Saran440 (approved):
Code Reivewed, However it's LGTM :+1:

base_tier_validation_server_action/models/tier_review.py
kittiu: I think the result is the same, status can never be both either.
Note: I am trying to not use `else` and `elif` also in the code, the code will be cleaner.

PR #359, opened 2021-07-03
[14.0][IMP] base_tier_validation, use sudo to post message

kittiu:
We face problem for case when user has only readonly access to document, he still can't validate because after that it will post message to chatter (write access).

This PR resolve it by sudo()

@JordiBForgeFlow @LoisRForgeFlow do you think this can be FFW?


newtratip (approved):
LGTM :+1:

LoisRForgeFlow (approved):
Seems fair

PR #360, opened 2021-07-04
[14.0][ADD] sequence_prefix_suffix_increment

kittiu:
This module increment number part of prefix / suffix.

For example, to run sequence in Buddhist year by increase +543, then PO2021/0001 can become PO2564/0001

image

@pedrobaeza WDYT about this? should be in server-ux or server-tools ?


pedrobaeza:
It seems a bit odd. Shouldn't this be instead `base_sequence_buddhist_year` that adds a new placeholder called `buddhist_year` and `range_buddhist_year` that gets such year?

kittiu:
> It seems a bit odd. Shouldn't this be instead `base_sequence_buddhist_year` that adds a new placeholder called `buddhist_year` and `range_buddhist_year` that gets such year?

Buddhist year is one thing, I use that example to be easy to understand. In fact, on current project we need this,

* Says, fiscal year start 10/01/2021 - 09/31/2022
* By using range_year, on 05/01/2021 prefix is PO2020, and on 10/01/2021 prefix is PO2021.
* But in fact, 10/01/2021 - 09/31/2022, we consider it FY2022, and prefix should be PO2022. So I need to +1 in this case.

Even I make for range_buddhist_year, also, I still need to +1.

kittiu:
@pedrobaeza another thinking, not sure it is possible. Can we make `range_end_year` which use ending range date instead of starting range date. Then, we can use `range_end_buddhist_year` also.

pedrobaeza:
> @pedrobaeza another thinking, not sure it is possible. Can we make `range_end_year` which use ending range date instead of starting range date. Then, we can use `range_end_buddhist_year` also.

I think this is just what the placeholder does without `range` part.

pedrobaeza:
I remember @legalsylvain did something similar.

kittiu:
> I think this is just what the placeholder does without `range` part.

Not exactly. It just not care about fiscal year (range) anymore.

kittiu:
@pedrobaeza I make this `sequence_date_range` (I saw many sequence_xxx here, or should it be base_xxx). https://github.com/OCA/server-ux/pull/361

With the same method, I can have `sequence_buddhist_year`

WDYT?

pedrobaeza:
OK

legalsylvain:
> I remember @legalsylvain did something similar.

regarding end date range, I did (and deployed) that odoo improvement : 
https://github.com/odoo/odoo/pull/61384

However, this PR is sleeping.

I also worked on that module https://github.com/OCA/account-financial-tools/pull/1005

hope it could help.

regards.

kittiu:
> > I remember @legalsylvain did something similar.
> 
> regarding end date range, I did (and deployed) that odoo improvement :
> [odoo/odoo#61384](https://github.com/odoo/odoo/pull/61384)
> 
> However, this PR is sleeping.
> 
> I also worked on that module [OCA/account-financial-tools#1005](https://github.com/OCA/account-financial-tools/pull/1005)
> 
> hope it could help.
> 
> regards.

Thank you very much. I am surprised how #361 is similar to yours. I will go ahead with that one then :).

PR #365, opened 2021-07-09
[12.0][FIX] date_range: Add comment to missing dependency for odoo_test_helper

joao-p-marques:
@Tecnativa

ping @pedrobaeza @victoralmau

### DIFF
--- a/date_range/__manifest__.py
+++ b/date_range/__manifest__.py
@@ -12,6 +12,7 @@
     "depends": [
         "web",
     ],
+    # odoo_test_helper is needed for the tests
     "data": [
         "security/ir.model.access.csv",
         "security/date_range_security.xml",


date_range/README.rst
```diff
@@ -41,6 +41,8 @@ Installation
 
 The addon use the daterange method from postgres. This method is supported as of postgresql 9.2
 
+The `odoo_test_helper` Python library is needed for the tests.
```
pedrobaeza: README is a user documentation, so this shouldn't be here.

joao-p-marques: Thanks, I reverted the change :+1:

pedrobaeza (approved):

victoralmau (approved):

PR #367, opened 2021-07-15
[14.0][ENH] Add Notification message on restart tier

newtratip:
I use sudo to post message follow this PR https://github.com/OCA/server-ux/pull/359

Pls review @kittiu

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -251,6 +251,9 @@ class TierValidation(models.AbstractModel):
     def _get_rejected_notification_subtype(self):
         return "base_tier_validation.mt_tier_validation_rejected"
 
+    def _get_restarted_notification_subtype(self):
+        return "base_tier_validation.mt_tier_validation_restarted"
+
     def _notify_accepted_reviews(self):
         post = "message_post"
         if hasattr(self, post):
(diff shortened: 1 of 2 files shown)


kittiu:
Test and Code Review OK,

I understand that you want to add a message on click "Restart Validation" button, as currently there is no log in chatter.

If anything was confused is the the wording `Restart Validation` which is not really  "Restarting", but only "Reset" (stop/remove tiers), and user will need to click "Request Validation" again. (but I can't think of any other suitable name)

I think the log message should be changed to "The review has been reset by Mr.A".

image

base_tier_validation/models/tier_validation.py
```diff
@@ -388,11 +391,23 @@ def request_validation(self):
         self._notify_review_requested(created_trs)
         return created_trs
 
+    def _notify_restarted_review_body(self):
+        return _("A review was restarted by %s.") % (self.env.user.name)
```
kittiu:
```suggestion
        return _("The review has been reset by %s.") % (self.env.user.name)
```

kittiu (approved):

base_tier_validation/models/tier_validation.py
newtratip: Fixed.

Saran440 (approved):
LGTM :+1:

CasVissers-360ERP (approved):
👍

LoisRForgeFlow (approved):
LGTM

PR #361, opened 2021-07-05
[14.0][ADD] sequence_range_end

kittiu:
When configure for sequence prefix/suffix, this module add option `range_end_` in addition to `range_`.
to use the ending of the range instead of the current date, e.g. %(range_end_year)s instead of %(year)s.

TODO:

- [x] Test

PR #369, opened 2021-07-21
[14.0][ADD] sequence_range_end_buddhist_calendar

newtratip:
This module is change year of sequence number from A.D. to B.E.

When select option Use B.E., year in prefix / suffix will change from Anno Domini (A.D.) to Buddhist Calendar (B.E.)

Depend on sequence_range_end module https://github.com/OCA/server-ux/pull/361

cc: @kittiu

### DIFF
--- a/sequence_range_end_buddhist_calendar/__init__.py
+++ b/sequence_range_end_buddhist_calendar/__init__.py
@@ -1,0 +1,3 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from . import models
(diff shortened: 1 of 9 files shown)


kittiu:
I think this module should name `sequence_buddhist_year` and should work in every case, start or end of year, or even when not using range_year. Ie..,

* buddhist_year
* range_buddhist_year
* range_end_buddhist_year

newtratip:
@kittiu, I agree but still can not find the best solution, however I will find the new solution and make a new PR.

Thanks for your suggest :)

kittiu:
@newtratip any update on this?

kittiu:
Moved to https://github.com/OCA/l10n-thailand/pull/244

@newtratip please can you close this one.

newtratip:
@kittiu OK.

PR #371, opened 2021-07-22
[13.0][IMP] base_tier_validation: add new review type

MateuGForgeFlow:
Add a review type based on a res.users field in the tier definition
cc @LoisRForgeFlow 

To do:

* [x] : Add tests

### DIFF
--- a/base_tier_validation/models/tier_review.py
+++ b/base_tier_validation/models/tier_review.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 
 class TierReview(models.Model):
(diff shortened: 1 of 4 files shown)


base_tier_validation/models/tier_review.py
```diff
@@ -91,4 +95,12 @@ def _compute_todo_by(self):
...
 
     def _get_reviewers(self):
-        return self.reviewer_id + self.reviewer_group_id.users
+        if self.reviewer_id or self.reviewer_group_id.users:
+            return self.reviewer_id + self.reviewer_group_id.users
+        reviewer_field = self.env["res.users"]
+        if self.reviewer_field_id:
+            resource = self.env[self.model].browse(self.res_id)
+            reviewer_field = getattr(resource, self.reviewer_field_id.name, False)
+        if not reviewer_field._name == "res.users":
```
JordiBForgeFlow: You should also check that reviewer_field has a value. Otherwise the review is initiated with no user.

LoisRForgeFlow: This opens the possibility that you set a field that can be empty, if it is empty no review is needed. If you want to always ensure a review make the "reviewer" field required in the reviewed model.

LoisRForgeFlow (approved):
Small correction in the tests done. LGTM!

PR #375, opened 2021-08-01
[14.0] Add module date_range_account

alexis-via:
Simple auto_install module that depend on account and date_range. Allows the Accounting Manager to access the date range menu entry.


bunhin:
Hi @alexis-via , i do test in my instance, working the menus shown,  and no error found when create date range type, creat date range, and generate date range, only the helper string in field allow_overlap may need to be corrected  from "if sets date range of same type must not overlap" to become as current correct behavior   "set if date range of same type to allow overlap". but it is in date_range module inherited

rvalyi (approved):

alexis-via:
Who has the superpowers on this repo ?

PR #378, opened 2021-08-06
[14.0][IMP] base_tier_validation, add tier description on stripe

kittiu:
The snippet in this enhancement is provided by @nmoturi, Thank you!

image

@nmoturi, by the way, as I tested it I found that the description will appear after a while, like 1 mins. Do you experience this too?

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -48,6 +48,7 @@ class TierValidation(models.AbstractModel):
         compute="_compute_can_review", search="_search_can_review"
     )
     has_comment = fields.Boolean(compute="_compute_has_comment")
+    next_review = fields.Char(compute="_compute_next_review")
 
     def _compute_has_comment(self):
         for rec in self:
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow:
> @nmoturi, by the way, as I tested it I found that the description will appear after a while, like 1 mins. Do you experience this too?

Regarding this message it seems that this still needs more love/testing, I will mark the PR as work in progress.

Also, why don't you do this server side? i.e. add a computed field with the message and extend the template to use that field, at a first glance I don't anticipate problems and seems more clean and easy to unit-test than an ajax call.

kittiu:
@LoisRForgeFlow you are right! It is much simpler :)

nmoturi:
> @LoisRForgeFlow you are right! It is much simpler :)

Indeed, its much simpler and cleaner approach. Thanks @kittiu I see you had already pushed the change.

kittiu:
This module is ready for review. @LoisRForgeFlow can remove the badge?

LoisRForgeFlow (approved):
Now it looks good to me :+1:

Functional test is correct

LoisRForgeFlow:
@kittiu can you squash into only one commit?

kittiu:
> @kittiu can you squash into only one commit?

Done, thanks!

LoisRForgeFlow:
@nmoturi could you drop a review here?

nmoturi:
Thanks @LoisRForgeFlow and @Kitti. The new description on stripe implementation works perfectly fine and looks functionally sound.

kittiu:
> Thanks @LoisRForgeFlow and @kitti. The new description on stripe implementation works perfectly fine and looks functionally sound.

Can you click approve?

nmoturi (approved):

PR #381, opened 2021-09-27
[13.0][IMP] base_tier_validation: review_user_count perfromance

AdriaGForgeFlow:
This PR intends to improve the method that gets the number of reviews a user has pending.
There can be cases where a user has a lot of pending reviews and every time the polling system triggers the _updateReviewPreview function it can make Odoo load for some seconds which is not nice for the user experience.

Before:
* 800 pending reviews --> executing the  review_user_count function takes around 8 seconds

Now:
* 800 pending reviews --> executing the  review_user_count function takes around 2 seconds

cc ~ @ForgeFlow

### DIFF
--- a/base_tier_validation/models/res_users.py
+++ b/base_tier_validation/models/res_users.py
@@ -12,37 +12,30 @@ class Users(models.Model):
     @api.model
     def review_user_count(self):
         user_reviews = {}
-        to_review_docs = {}
-        reviews = self.env["tier.review"].search(
-            [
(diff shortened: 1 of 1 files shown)


etobella (approved):
Code Review LGTM

LoisRForgeFlow (approved):
LGTM

LoisRForgeFlow:
Thanks, could you forward port it to v14?

/ocabot merge patch

PR #385, opened 2021-10-05
[FIX] base_tier_validation: get definition by sequence

fmdl:
Create 2 définition with A: sequence 10, B: sequence 20
On document ask validation
The order is reverse:
1 - B
2 - A


kittiu:
@fmdl this is the known fact, and has been discussed for many times. I think the final decision was to keep as is.
Yes, the higher sequence will come first.

PR #394, opened 2021-10-29
Update date_range_view.xml

robertobandera123:
Agregando etiqueta sheet al formulario para agrupar los campos, pues fecha final se quedaba fuera de zoom.


etobella:
This repo is not only fot spanish people. Comment in english please

kittiu:
My option is, this is not needed. Normally, we use <sheet> for a document, and date.range is not.

PR #397, opened 2021-11-03
[14.0][ADD] base_new_line_default

kittiu:
When add a new document line (one2many), it is now possible to use values from previous line as default.

Peek 2021-11-03 10-49

Usage:

Just add context to the one2many field to use, i.e.,

```
context="{'new_line_default': model_access, 'new_line_default_cols': ['name', 'model_id']}"
```


base_new_line_default/models/models.py
```diff
@@ -0,0 +1,36 @@
...
+    _inherit = "base"
+
+    @api.model
+    def default_get(self, field_list):
+        res = super().default_get(field_list)
+        lines = self.env.context.get("new_line_default")
+        field_list = self.env.context.get("new_line_default_cols")
+        if lines:
+            # Filter for create (0) and link (4) only
+            lines = list(filter(lambda l: l[0] in (0, 1, 4), lines))
```
simahawk:
```suggestion
            lines = [l for l in lines if l[0] in (0, 1, 4)]
```

base_new_line_default/models/models.py
```diff
@@ -0,0 +1,36 @@
...
+            last_vals = {}
+            action = last_line[0]
+            # From new line
+            if action == 0:
+                last_vals = last_line[2]
+            # From link or update
+            if action in (1, 4):
+                last_id = last_line[1]
+                last_rec = self.browse(last_id)
+                last_vals = last_rec._convert_to_write(last_rec.read()[0])
```
simahawk:
```suggestion
                last_vals = last_rec._convert_to_write(last_rec.read(field_list)[0])
```

base_new_line_default/models/models.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class BaseModel(models.AbstractModel):
+    _inherit = "base"
+
+    @api.model
+    def default_get(self, field_list):
+        res = super().default_get(field_list)
+        lines = self.env.context.get("new_line_default")
+        field_list = self.env.context.get("new_line_default_cols")
```
simahawk:
```suggestion
        field_list = self.env.context.get("new_line_default_cols", [])
```

base_new_line_default/models/models.py
```diff
@@ -0,0 +1,36 @@
...
+            if action in (1, 4):
+                last_id = last_line[1]
+                last_rec = self.browse(last_id)
+                last_vals = last_rec._convert_to_write(last_rec.read()[0])
+                if action == 1:
+                    update_vals = last_line[2]
+                    last_vals.update(update_vals)
+            # if has specific field_list, filtered out unused field.
+            if field_list:
+                last_vals = {k: last_vals[k] for k in field_list}
```
simahawk: you should also pop out `id` no?

simahawk (changes requested):

simahawk:
seems a nice tool, tnx!

base_new_line_default/models/models.py
kittiu: OK, this is cleaner, thanks.

base_new_line_default/models/models.py
kittiu: Actually, I shouldn't use the world field_list when getting value from context, as it duplicated with default_get's param. Now, I changed the name. Thanks!

base_new_line_default/models/models.py
kittiu: Done

base_new_line_default/models/models.py
kittiu: Done.

kittiu:
> seems a nice tool, tnx!

@simahawk thanks!

pedrobaeza:
I think this one can be more versatile being also handled at web client level: https://github.com/OCA/web/pull/2062

kittiu:
@simahawk I did add another option, to use default value from document header's fields too. In deed, it was our need at the first place, and see that it is now possible.

kittiu:
> I think this one can be more versatile being also handled at web client level: [OCA/web#2062](https://github.com/OCA/web/pull/2062)

Thanks @pedrobaeza I have checked that out.

For this, the original need was in a custom module, to pass values from header (3-4 fields) to line (and hide those colums) to simplify end user data input when dealing with many lines. So, I think this can still serve well.

**Note:**

But this module, the limitation is that, once default is set, the onchange is called and it may erase some default values (like the Taxes in above VDO).

Are there ways to by pass onchange call?

pedrobaeza:
You can do that without using any module. Just pass in the context of the one2many this:

`context={'default_field_name': parent.field_name}`

kittiu:
> You can do that without using any module. Just pass in the context of the one2many this:
> 
> `context='default_field_name': parent.field_name`

OK, make sense. I will drop this soon, if no other comment.

kittiu:
@ps-tubtim I will close this.

PR #399, opened 2021-11-05
[14.0][IMP] base_tier_validation, approve by sequence with option to bypass for same reviewer

kittiu:
Old behaviour with approve by sequence

#. Tier 1 to be approved by Mr. A
#. Tier 2 to be approved by Mr. A
#. Tier 3 to be approved by Mr. A

Mr. A will have to approve 3 times.

But with option approve_sequence_bypass = True in Tier 2 and Tier 3

Mr.A will only approve 1 time, then tier 2 and tier 3 will auto approve immediately.

### DIFF
--- a/base_tier_validation/models/tier_definition.py
+++ b/base_tier_validation/models/tier_definition.py
@@ -73,6 +73,9 @@ class TierDefinition(models.Model):
         default=False,
         help="Approval order by the specified sequence number",
     )
+    approve_sequence_bypass = fields.Boolean(
+        help="Bypassed (auto validated), if previous tier was validated by same reviewer",
+    )
 
     @api.onchange("review_type")
     def onchange_review_type(self):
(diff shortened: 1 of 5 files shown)


kittiu:
cc @newtratip

kittiu:
Note: I understand that this by sequence is expected behavior in the past. But it looks like sometime it is not practical in real life to have user click multiple time to steps through the sequence. So, we need this bypassing feature.

newtratip (approved):
LGTM :+1:

kittiu:
@LoisRForgeFlow do you have comment about this feature?
Another option though, is to make it global configuration than in tier definition (but it won't be as flexible)

Saran440 (approved):

kittiu:
@OCA/tools-maintainers

ps-tubtim (approved):

PR #401, opened 2021-11-09
[15.0] date_range Migration to V15

alan196:
### Proposed changes

Migration PR for module date_range.

### Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Migration Update

### Checklist

- [x] I have read the CONTRIBUTING doc
- [x] I have signed the CLA
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the necessary documentation

PR #402, opened 2021-11-10
[MIG][15.0]15.0 mig date range

JoanMForgeFlow:
Migration to version 15, module date_range,

The migration has not been completed, because the file "date_range.js",  function CustomFilterItem.patch (...) does not work properly, the following error appears:
`Error while loading date_range.CustomFilterItem: CustomFilterItem.patch is not a function` 

@AaronHForgeFlow 
@JordiMForgeflow


pedrobaeza:
Please check before starting any work if there's already one PR for the same. In this case, #401.

PR #421, opened 2021-12-17
[14.0][IMP] base_tier_validation: Add hook for state transition

rousseldenis:


base_tier_validation/models/tier_validation.py
```diff
@@ -193,6 +193,15 @@ def _check_allow_write_under_validation(self, vals):
                 return False
         return True
 
+    def _check_tier_state_transition(self, vals):
+        """
+        Check we are in origin state and no destination state
```
kittiu: no or not?
```suggestion
        Check we are in origin state and not destination state
```

kittiu (approved):

base_tier_validation/models/tier_validation.py
rousseldenis: @kittiu Done

simahawk (approved):

PR #423, opened 2021-12-20
[14.0][FIX] base_tier_validation: condition state

Saran440:
check condition `vals.get(self._state_field)` is not null for other module use write function it will error (i.e. base_exception)

- [ ] Test script

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -198,9 +198,12 @@ class TierValidation(models.AbstractModel):
         Check we are in origin state and not destination state
         """
         self.ensure_one()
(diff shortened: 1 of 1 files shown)


kittiu:
@rousseldenis can you check this?
This should be a safe fix, and make it work with base_exception.

kittiu:
@Saran440 please check Travis.

rousseldenis:
@Saran440 @kittiu IMHO, the complete content of the condition should be included in a function in order to have a better inehritance possibilities and comprehension

rousseldenis:
With a good comment

Saran440:
okay, it move on my project.
Thank you :)

PR #426, opened 2021-12-21
[14.0][IMP] base_tier_validation: Include inactive records in activities

rousseldenis:
If record validations are based on a domain with active = False,
then the activities are not visible.

### DIFF
--- a/base_tier_validation/models/res_users.py
+++ b/base_tier_validation/models/res_users.py
@@ -25,6 +25,7 @@ class Users(models.Model):
                 records = (
                     self.env[model]
                     .with_user(self.env.user)
+                    .with_context(active_test=False)
                     .search([("id", "in", reviews.mapped("res_id"))])
                     .filtered(lambda x: not x.rejected and x.can_review)
                 )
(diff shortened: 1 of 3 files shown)


kittiu (approved):

rousseldenis:
@kittiu I've added a commit because there is a drawback with the context key in js.

The view is loaded with that context and is propagated on write(). That part is a little bit ugly but the only solution

etobella:
This change is arising a lot of problems. For example, in stock.request.orders, it is using archived routes and rules. This is obviously a problem. We need to revert these changes. The right option could be to add a filter with true or false for active.

@rousseldenis @kittiu @LoisRForgeFlow

LoisRForgeFlow:
> This change is arising a lot of problems. For example, in stock.request.orders, it is using archived routes and rules. This is obviously a problem. We need to revert these changes. The right option could be to add a filter with true or false for active.
> 
> @rousseldenis @kittiu @LoisRForgeFlow

I agree, this seems very aggressive change and a active true/false filter would be cleaner and have less to none side effects.

PR #427, opened 2021-12-22
[13.0][ADD] base_custom_filter

JordiMForgeFlow:
This new module allows to define custom filters that are dynamically added in the standard 'Filters' dropdown of any model.

To create a new filter you need to go to Settings > Custom Filters. You simply need to select the model and define the filter.

image

The filter will be then dynamically added under the 'Filters' dropdown:

image

You can also group filters so that they appear under the same separator. For that purpose first create a Custom Filter Group and add the desired filters.

image

image


base_custom_filter/views/ir_custom_filter_views.xml
```diff
@@ -0,0 +1,62 @@
...
+                <sheet>
+                    <div class="oe_title">
+                        <h1>
+                            <field name="name" />
+                        </h1>
+                    </div>
+                    <group>
+                        <field name="model_id" options="{'no_create_edit': True}" />
+                        <field name="model_name" invisible="1" />
+                        <field name="group_id" />
```
LoisRForgeFlow: can you pass the model_id in the context here? something like ~ `context: default_model_id: model_id`. This way when creating a filter group on the go, the model will be defaulted.

LoisRForgeFlow (commented):
Functional test, works like a charm :+1:

base_custom_filter/views/ir_custom_filter_views.xml
JordiMForgeFlow: Done! Thank you :)

LoisRForgeFlow (approved):

HaraldPanten (approved):
A nice one! Functional review.

ValentinVinagre:
Can it be used for dynamic date filters? for example. today or tomorrow ...

JordiMForgeFlow:
> Can it be used for dynamic date filters? for example. today or tomorrow ...

If you mean something like below, yes it can.

image

HaraldPanten:
Not sure if it works dynamically. The domain "Today" will get 22/12/2021 now, but tomorrow I don't think it will consider 23/12/2021, for example.

digitalsatori:
@JordiMForgeFlow except for the custom defined filter showing in the filter dropdown menu, what could be the difference when I  define a filter in the search bar and save it as a favorite.

JordiMForgeFlow:
@digitalsatori Although from a technical perspective there is no much difference, for final users it is much different to have a filter as favorite or to have it as a 'standard'.

As a developer I tend to face customer requests to add new filters as 'standard' as they are the most commonly used across the organization and they are usually the ones users access the most. However, this implies having to add new code each time just to add a new filter there. That is the main pain that I want to ease with this module.

digitalsatori:
I c, worked as handy tool for implementers sounds reasonable to me. It will be even better if you can support dynamic filter as mentioned by @HaraldPanten. So that we can define something like ’today’, ’this week’, ’this month’ etc. instead of fixed the date value in the domain.

Thank you

JordiMForgeFlow:
@LoisRForgeFlow @HaraldPanten I am adding also the possibility to add group by filters. Could you review and let me know what do you think? :)

I am thinking on how I could add support on the domains to specify datetime parameters, but it does not seem to be possible using the domain widget. Do you have any suggestions?

pedrobaeza:
I think you shouldn't create a new model, but reuse existing one `ir.filters`, adding a checkbox to say that is of this type. The reason: on migrations, standard openupgradelib methods like `rename_models` and `rename_fields` act over this table to perform certain automatisms, like changing the name of a field in a domain, so I think that's very valuable.

And this is another reason for thinking twice if this module is a bomb mechanism or not: with it, you are converting normal code in DB data, that is not verifiable through CI when migrating a DB. On a regular basis, if someone asks you for a new filter, you add it to the custom code, and then, when migrating to the new version, you migrate and adapt such code, and even the CI may turn red if not correct. Now, you (or even the customer), creates a record in the DB specifying such filter, and then, you have to manually review these records to check if they are going to work on the new version. It's like server actions.

JordiMForgeFlow:
> I think you shouldn't create a new model, but reuse existing one `ir.filters`, adding a checkbox to say that is of this type. The reason: on migrations, standard openupgradelib methods like `rename_models` and `rename_fields` act over this table to perform certain automatisms, like changing the name of a field in a domain, so I think that's very valuable.
> 
> And this is another reason for thinking twice if this module is a bomb mechanism or not: with it, you are converting normal code in DB data, that is not verifiable through CI when migrating a DB. On a regular basis, if someone asks you for a new filter, you add it to the custom code, and then, when migrating to the new version, you migrate and adapt such code, and even the CI may turn red if not correct. Now, you (or even the customer), creates a record in the DB specifying such filter, and then, you have to manually review these records to check if they are going to work on the new version. It's like server actions.

It is true. I added a new model for simplicity but I guess that I can reuse the standard one, I will look into it.
On the other part, it is the drawback of using these dynamic domains, but it is the same that happens in any other model that uses them.

bjoliveau:
Nice feature @JordiMForgeFlow !

@digitalsatori @ValentinVinagre For feature around dates and filters, I invite you on this https://github.com/Numigi/odoo-web-addons/tree/12.0/web_search_date_range

ValentinVinagre:
> Nice feature @JordiMForgeFlow !
> 
> @digitalsatori @ValentinVinagre For feature around dates and filters, I invite you on this https://github.com/Numigi/odoo-web-addons/tree/12.0/web_search_date_range

Hi, nice modules 😃 . Have you considered taking them to OCA?

digitalsatori:
> @digitalsatori @ValentinVinagre For feature around dates and filters, I invite you on this https://github.com/Numigi/odoo-web-addons/tree/12.0/web_search_date_rang

Looks awesome, thanks for sharing @bjoliveau

elvise:
@JordiMForgeFlow any chance merge ?

PR #434, opened 2022-01-25
[14.0][IMP] base_tier_validation_server_action: Avoid recursive server actions

rousseldenis:
@kittiu


kittiu:
@rousseldenis what is the case when this happen? And where are you are passing context "server_action_tier" from?

rousseldenis:
The exact one is a user in validation group that changes a field manually that triggers a server action that changes that field.

kittiu:
> changes a field manually

IC, thanks for the fix.

kittiu (approved):

Saran440 (approved):

PR #447, opened 2022-02-12
[14.0][FIX] base_tier_validation_server_action:  rejected with server action error

Saran440:
[BUG]
you can not rejected document when you use tier validation server action with state cancel
because it remove **tier.review** when state cancel https://github.com/OCA/server-ux/blob/14.0/base_tier_validation/models/tier_validation.py#L243

Step to error:

1. Create tier and Post Reject Action = change state to cancel 
2. Reject document and it will error

[FIX]
- Reject first and run server
- cherry-pick PR https://github.com/OCA/server-ux/pull/434 because it will conflict file.

@kittiu @rousseldenis

### DIFF
--- a/base_tier_validation_server_action/models/__init__.py
+++ b/base_tier_validation_server_action/models/__init__.py
@@ -2,3 +2,4 @@
 
 from . import tier_definition
 from . import tier_review
+from . import tier_validation
--- a/base_tier_validation_server_action/models/tier_review.py
+++ b/base_tier_validation_server_action/models/tier_review.py
@@ -9,13 +9,9 @@ class TierReview(models.Model):
 
     def write(self, vals):
         res = super().write(vals)
-        if vals.get("status") in ["approved", "rejected"]:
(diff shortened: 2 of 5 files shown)


kittiu:
Just FYI, your error come from this fix. https://github.com/OCA/server-ux/pull/343
You may need to abide by the latest version.

kittiu:
Look like this is a difficult fix. Let's ask for the reason to remove the reviews https://github.com/OCA/server-ux/pull/343#issuecomment-1063592275

kittiu:
@Saran440 the solution should be as simple as this,

kittiu:
@Saran440 the solution can be as simple as this, and I propose the changes to @rousseldenis 

https://github.com/OCA/server-ux/pull/495#issuecomment-1143107010

Can you close this PR?

kittiu:
> @Saran440 the solution should be as simple as this,

https://github.com/OCA/server-ux/pull/495#issuecomment-1143107010

PR #448, opened 2022-02-14
[MIG] base_custom_filter: migrate to v14.0 from 13.0

AshishHirapara:
migrated base_custom_filter module from  version 13 to version 14


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

PR #449, opened 2022-02-14
[14.0][IMP] base_revision: Add revisions_count field + update the state field if it exists in the model.

victoralmau:
Add revisions_count field + update the state field if it exists in the model.

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa

### DIFF
--- a/base_revision/models/base_revision.py
+++ b/base_revision/models/base_revision.py
@@ -38,6 +38,22 @@ class BaseRevision(models.AbstractModel):
     )
     active = fields.Boolean(default=True)
     has_old_revisions = fields.Boolean(compute="_compute_has_old_revisions")
+    revision_count = fields.Integer(
+        compute="_compute_revision_count", string="Previous versions count"
+    )
+
+    @api.depends("old_revision_ids")
+    def _compute_revision_count(self):
+        res = self.with_context(active_test=False).read_group(
+            domain=[("current_revision_id", "in", self.ids)],
+            fields=["current_revision_id"],
+            groupby=["current_revision_id"],
+        )
+        revision_dict = {
+            x["current_revision_id"][0]: x["current_revision_id_count"] for x in res
+        }
+        for rec in self:
+            rec.revision_count = revision_dict.get(rec.id, 0)
 
     _sql_constraints = [
         (
@@ -67,6 +83,9 @@ class BaseRevision(models.AbstractModel):
             "old_revision_ids": [(4, self.id, False)],
         }
 
+    def _prepare_revision_data(self, new_revision):
+        return {"active": False, "current_revision_id": new_revision.id}
+
     def copy_revision_with_context(self):
         default_data = self.default_get([])
         new_rev_number = self.revision_number + 1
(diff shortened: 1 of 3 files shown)


base_revision/models/base_revision.py
```diff
@@ -38,6 +38,20 @@ def _compute_has_old_revisions(self):
     )
     active = fields.Boolean(default=True)
     has_old_revisions = fields.Boolean(compute="_compute_has_old_revisions")
+    revisions_count = fields.Integer(compute="_compute_revisions_count")
```
pedrobaeza: Don't use plural:
```suggestion
    revision_count = fields.Integer(compute="_compute_revisions_count")
```

base_revision/models/base_revision.py
```diff
@@ -74,9 +88,10 @@ def copy_revision_with_context(self):
         default_data.update(vals)
         new_revision = self.copy(default_data)
         self.old_revision_ids.write({"current_revision_id": new_revision.id})
-        self.write(
-            {"active": False, "state": "cancel", "current_revision_id": new_revision.id}
-        )
+        vals = {"active": False, "current_revision_id": new_revision.id}
```
pedrobaeza: Add a `_prepare` hook method for this for easing possible extension

base_revision/models/base_revision.py
```diff
@@ -38,6 +38,20 @@ def _compute_has_old_revisions(self):
...
+        res = self.with_context(active_test=False).read_group(
+            domain=[("current_revision_id", "in", self.ids)],
+            fields=["current_revision_id"],
+            groupby=["current_revision_id"],
+        )
+        revision_dict = {
+            x["current_revision_id"][0]: x["current_revision_id_count"] for x in res
+        }
+        for rec in self:
+            rec.revision_count = revision_dict.get(rec.id, 0)
```
pedrobaeza, line 56: Shouldn't be + 1 for the current revision + old revisions?

base_revision/models/base_revision.py
```diff
@@ -67,16 +81,20 @@ def _get_new_rev_data(self, new_rev_number):
             "old_revision_ids": [(4, self.id, False)],
         }
 
+    def _prepare_revision_data(self, new_revision):
+        vals = {"active": False, "current_revision_id": new_revision.id}
+        if "state" in self._fields:
+            vals.update({"state": "cancel"})
```
pedrobaeza: Why assume that the inherited model is going to have such selection value in `state`? Do this in the inherited model. That's why we have put the hook.

dreispt (approved):

pedrobaeza:
Aren't you labeling fields with the texts we talked?

pedrobaeza:
We talked about putting the smart button as "Prev. revisions", and the fields as "Previous versions count".

victoralmau:
> We talked about putting the smart button as "Prev. revisions", and the fields as "Previous versions count".

I have defined "Previous versions count" in the field, but the smart-button will be added here https://github.com/OCA/sale-workflow/pull/1911 when this PR is merged.

pedrobaeza (approved):

PR #458, opened 2022-02-25
[14.0][FIX] mass_editing: deleted wrong FKs in pre-migration

MiquelRForgeFlow:
The obsolete table if `mass_editing`.


MiquelRForgeFlow:
@pedrobaeza

pedrobaeza (approved):

PR #462, opened 2022-03-08
[13.0][FIX+IMP] base_tier_validation: Don't mess searches with NewId + non saved data

pedrobaeza:
You can't search for a NewId in a domain, and doing a search requires to flush cache contents to DB for being effective, so this disregards the correct use of this on the fly computed field.

Fortunately, we have a tool at ORM level for applying domains on cached content and without worrying about NewId, which is `filtered_domain`, so let's apply it.


sergio-teruel (approved):

LoisRForgeFlow (approved):
LGTM

PR #465, opened 2022-03-09
[15.0][MIG] user_role_analysis: Migration to 15.0

JaumeBforgeFlow:
Migration of the user_role_analysis module added on pull request #464

### DIFF
--- a/user_role_analysis/__manifest__.py
+++ b/user_role_analysis/__manifest__.py
@@ -1,0 +1,19 @@
+# Copyright 2021 ForgeFlow, S.L. (https://www.forgeflow.com)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+{
+    "name": "User Role Analysis",
+    "version": "15.0.1.0.0",
+    "summary": "specific user role analysis reports",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/server-ux",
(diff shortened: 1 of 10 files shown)


user_role_analysis/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2021 ForgeFlow, S.L. (https://www.forgeflow.com)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+{
+    "name": "User Role Analysis",
+    "version": "15.0.1.0.0",
+    "summary": "specific user role analysis reports",
```
simahawk, line 6: can you elaborate a bit more here? Also, having a readme fragment + screenshot of what you get as result would be nice :pray:

user_role_analysis/report/user_role_reports.py
```diff
@@ -0,0 +1,213 @@
...
+from odoo.tools import sql
+
+
+class UserModelPermissionReport(models.Model):
+    _name = "user.model.permission.report"
+    _auto = False
+    _description = "User Model Permission Report"
+    _order = "x_user_id, x_group_name, x_model_name, id"
+
+    x_user_id = fields.Many2one(comodel_name="res.users", string="User", readonly=True)
```
simahawk, line 14: why all `x_`? This is a real model not a manual one.

PR #467, opened 2022-03-09
[14.0][FIX+IMP] base_tier_validation: Don't mess searches with NewId + non …

LoisRForgeFlow:
…saved data

You can't search for a NewId in a domain, and doing a search requires
to flush cache contents to DB for being effective, so this disregards
the correct use of this on the fly computed field.

Fortunately, we have a tool at ORM level for applying domains on cached
content and without worrying about NewId, which is filtered_domain, so
let's apply it.

Forward port of #462.

cc @pedrobaeza

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -7,7 +7,6 @@ from lxml import etree
 
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
-from odoo.osv import expression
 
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

pedrobaeza:
Fast-tracking as already merged in previous version:

/ocabot merge patch

LoisRForgeFlow:
it got stuck.

/ocabot merge patch

PR #469, opened 2022-03-11
[FIX] chained_swapper - search chain field with correct model

enriquemartin:
When searching for the field to be modified (to compare the types), the source model was being applied instead of the target model.


pedrobaeza (approved):

pedrobaeza:
Cherry-picked to v13 at ad89e0fa2509867dde8636abf009df1459e604bc and v15 at 3ca6329bdd3986ff7134bb5bf35ddb2f2a8348fb

PR #470, opened 2022-03-15
[13.0][IMP] base_tier_validation: Allow sequential tier reviews

JordiBForgeFlow:
It is possible that after one validation has been accepted, and before proceeding
with the transition of the status the conditions of the object have changed,
requiring an additional validation.
This change allows the engine to support this scenario.

cc @LoisRForgeFlow @judithaforgeflow @jaumebforgeflow

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -159,10 +159,10 @@ class TierValidation(models.AbstractModel):
 
     def _compute_validated_rejected(self):
         for rec in self:
-            rec.validated = not self.need_validation and self._calc_reviews_validated(
+            rec.validated = not rec.need_validation and rec._calc_reviews_validated(
                 rec.review_ids
             )
-            rec.rejected = not self.need_validation and self._calc_reviews_rejected(
+            rec.rejected = not rec.need_validation and rec._calc_reviews_rejected(
                 rec.review_ids
             )
 
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -37,6 +37,11 @@ class TierValidation(models.AbstractModel):
     rejected = fields.Boolean(
         compute="_compute_validated_rejected", search="_search_rejected"
     )
+    validation_max_sequence = fields.Integer(
+        string="Validation Maximum Sequence",
+        compute="_compute_validation_max_sequence",
+        store=True,
+    )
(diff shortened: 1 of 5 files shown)


JordiBForgeFlow:
@JaumeBforgeFlow Please run pre-commit.

base_tier_validation/models/tier_validation.py
```diff
@@ -49,6 +54,17 @@ class TierValidation(models.AbstractModel):
     has_comment = fields.Boolean(compute="_compute_has_comment")
     next_review = fields.Char(compute="_compute_next_review")
 
+    @api.depends("review_ids", "review_ids.status")
+    def _compute_validation_max_sequence(self):
```
etobella, line 91: You probably need to make a migration script for this. Otherwise, it will take a long time to reprocess all existent data.

JordiBForgeFlow: We have fixed in https://github.com/OCA/server-ux/pull/470/commits/fe3cf61f4b634e11675d2e00ac6c0d03c08bf524 making the field non-stored, searchable.

kittiu:
Hello Jordi this is a good feature for some use case.
But in some organization, this feature may not comply.

Can this feature be option in tier definition, WDYT?

JordiBForgeFlow:
@kittiu it's really an option. By default all reviews will be triggered at once. Only if you use the field 'validation_max_sequence' in your tier definition will you be able to achieve a sequential review.

For example, review with sequence 20 will only be achieved when the validation_max_sequence = 10 in the given object (purchase order, etc.)

kittiu:
> @kittiu it's really an option. By default all reviews will be triggered at once. Only if you use the field 'validation_max_sequence' in your tier definition will you be able to achieve a sequential review.
> 
> For example, review with sequence 20 will only be achieved when the validation_max_sequence = 10 in the given object (purchase order, etc.)

Thanks 👍 @JordiBForgeFlow 
May be a field help / readme on this would be nice.

JordiBForgeFlow:
@kittiu done!

JordiBForgeFlow:
@JaumeBforgeFlow can you cherry-pick fb67dfa927d0b9d157e7ed2c46b8b9c43e3a6ce0 to  PR's in 14.0 and 15.0?

JordiBForgeFlow:
@kittiu Can you approve this one? Thanks!

bosd:
Any change to get this one merged?

PR #481, opened 2022-03-31
[FIX] mass_operation_abstract (mass_editing, ...) : Prevent message if mass operation is realized on archived items

legalsylvain:
**Step to reproduce**

- select a user, and disable it.
- in the tree view, select it and click on "mass edit" action

A first error is displayed : 
image

When clicking on "Apply" button, a second message is displayed.

image

This trivial patch fixes the problem

GRAP#258

CC : @quentinDupont


flotho (approved):
easy fix indeed, LGTM

quentinDupont (approved):

PR #491, opened 2022-04-23
[14.0][ADD] base_disable_field_translate

kittiu:
When it is sure user don't want to use data translation. Translation of data can cause problem than being useful.
This technical module simply remove translation from all fields' data (same as set translate=False)

Selection_555

After install and restart server

image


kittiu:
Looks like this disable of translation will affect tests in other module, i.e.,

```
FAIL: TestMassEditing.test_mass_edit_partner_title
  File "/home/travis/build/OCA/server-ux/setup/mass_editing/odoo/addons/mass_editing/tests/test_mass_editing.py", line 110, in test_mass_edit_partner_title
    "Translation for Partner Title's Abbreviation " "was not loaded properly.",
AssertionError: 0 != 1 : Translation for Partner Title's Abbreviation was not loaded properly.
```

Anyone know how solve this? Anyway to ensure this module is installed last?

newtratip:
This module is disable translate `menu item`, `field label`, ... too (No only field's data). I think these should be translated if install multiple language in the system.

kittiu:
When this module is installed, or uninstalled. The server need restart to take effect (this is in README). 

I still not find the problem, i.e., th_TH

Selection_591

With this module installed + restart

Selection_590

Can you give me exact steps to get the problem?

kittiu:
OK, I think I know how you get to the problem. Please make sure this module is uninstalled when generate missing terms. I added in the README as following,

* Restart server after install or uninstall for this module to take effect
* Make sure this module is not installed, if you want to use Translations > Generate Missing Terms, as translated terms won't be loaded

kittiu:
@newtratip this should work now. Can you try again?

kittiu:
@newtratip have you test yet?

Issue #493, opened 2022-05-03
base_tier_validation_server_action: Execute action after the write() on record.


rousseldenis:
@kittiu I think there is a limitation in the behaviour doing the trigger of the action on reviews write() as the actual record is not still really written at that time. Therefore, we cannot rely on record values (that are in the write vals) to filter the action records.

I will propose a hook method in the write method in order to trigger the server action just after the super() call, there:

https://github.com/OCA/server-ux/blob/14.0/base_tier_validation/models/tier_validation.py#L283

What do you think ?

kittiu:
I agree 👍

PR #495, opened 2022-05-03
[14.0][IMP] base_tier_validation/server_action: Improve flow for validation conditions

rousseldenis:


rousseldenis:
Following this conversation: https://github.com/OCA/server-ux/issues/493

rousseldenis:
@kittiu

This was not so obvious but this prevent wrong record fields values evaluation when running record action.

I've split modifications in several commits to easy review. But I can split PR's if needed.

kittiu:
@rousseldenis I saw you are changing version too. Won't we just use ocabot merge minor ?
    "version": "14.0.2.10.0",
    "version": "14.0.3.0.0",

rousseldenis:
> @rousseldenis I saw you are changing version too. Won't we just use ocabot merge minor ?
>     "version": "14.0.2.10.0",
>     "version": "14.0.3.0.0",
> 

Yes that's done at purpose as I've added migration scripts.

kittiu:
Functional tested OK.
I review the code, am I understand correctly that the one thing you try to achieved is to run the "server action" only after the record is validated?

base_tier_validation_server_action/models/tier_review.py
```diff
@@ -7,24 +7,33 @@
 class TierReview(models.Model):
     _inherit = "tier.review"
 
+    def _run_server_action(self):
+        for rec in self.filtered(
```
kittiu, line 11: OK, I understand now. But since it is quite complicated. Can we add some comment here and there? WDYT?

kittiu (approved):

base_tier_validation/models/tier_review.py
```diff
@@ -20,6 +20,14 @@ class TierReview(models.Model):
     )
     model = fields.Char(string="Related Document Model", index=True)
     res_id = fields.Integer(string="Related Document ID", index=True)
+    record_id = fields.Reference(
```
rousseldenis, line 23: @LoisRForgeFlow I would like to have your review on this. Thanks

I kept former fields for compatibility.

kittiu:
@rousseldenis I still get the error as in #447 and I found the solution. Let's fix it here altogether?

Step to error:

- Setup a Tier Definition with Reject Server Action = Set document state to rejected
- And because when server action set rejection and the tier.review will all be removed.
- The error occur because tier.review was still in cache and the _rejected_tier() function will throw error `Record does not exist or has been deleted.`

https://github.com/OCA/server-ux/blob/84bb760a1c27479f2adc7f4f6db2eac5ccfd21d2/base_tier_validation/models/tier_validation.py#L434

Just change from
```
for review in user_reviews:
```
to
```
for review in user_reviews.exists():  # to only get record that really exists
```

kittiu:
@rousseldenis still this make sense to you to fix?

rousseldenis:
> @rousseldenis still this make sense to you to fix?

Ok, I'll add on my todo for today :smile:

bosd:
ping @rousseldenis

rousseldenis:
@bosd @kittiu I've implemented the exists() as suggested

bosd (approved):
Quick code review, change ok 👍

bosd:
@rousseldenis Can you forward port this to V15 & 16?

rousseldenis:
> @rousseldenis Can you forward port this to V15 & 16?

As soon as this is merged yes

kittiu (approved):

francesco-ooops:
@rousseldenis can this be merged?

francesco-ooops:
or rather @LoisRForgeFlow

base_tier_validation/models/tier_review.py
```diff
@@ -20,6 +20,14 @@ class TierReview(models.Model):
...
+        selection=lambda self: [
+            (m.model, m.name) for m in self.env["ir.model"].search([])
```
simahawk, line 25: maybe a search_read on model and name only is better?

base_tier_validation_server_action/models/tier_review.py
simahawk: no comment? :)

base_tier_validation_server_action/models/tier_review.py
```diff
@@ -7,24 +7,33 @@
...
+            lambda review: not review.record_id.tier_validation_before_write
```
simahawk, line 12: this will loop twice, can we skip records in the loop itself?

simahawk (changes requested):
Some comments to address.
Furthermore: @rousseldenis can you please rewrite the commits to shorten them and remove the odoo version? Thanks :pray:

francesco-ooops:
@rousseldenis any chance to complete this one as well?

bosd:
@rousseldenis Would you be so kind to attend to Simahawk's review?

rousseldenis:
> @rousseldenis Would you be so kind to attend to Simahawk's review?

Mmmh, I need to get time to finish this indeed. Mark that for later today

francesco-ooops:
@rousseldenis any chance to finish this one?

francesco-ooops:
@rousseldenis do you think you will pick it up again soon?

francesco-ooops:
hi @rousseldenis , any plan to bring this to completion? 

Thank you!

francesco-ooops:
@rousseldenis just trying :)

bosd:
@rousseldenis Friendly reminder ;)

base_tier_validation/models/tier_review.py
renda-dev: I don't totally agree with the `search_read` here since it would return a list of dictionaries including the ids of every record which it seems a little too much for me

base_tier_validation_server_action/models/tier_review.py
bosd: @renda-dev Do you have any thoughts on this?

renda-dev: I think that is not a really big deal, as long as it's not recursive and there are not too many operations inside the loop

francesco-ooops:
@rousseldenis you can close this as https://github.com/OCA/server-ux/pull/794 was merged. thanks!

PR #497, opened 2022-05-17
[14.0][ENH] base_cancel_confirm, allow disable cancel confirm

kittiu:
By default, the cancel confirm will be disabled (to ensure no side effect on other module unit test)

To enable cancel confirm wizard, please add System Parameter (ir.config_parameter) for each extended module.

For example,

* sale_cancel_confirm, add `sale.order.cancel_confirm_disable = False`
* purchase_cancel_confirm, add `purchase.order.cancel_confirm_disable = False`

### DIFF
--- a/base_cancel_confirm/model/base_cancel_confirm.py
+++ b/base_cancel_confirm/model/base_cancel_confirm.py
@@ -4,7 +4,8 @@ import inspect
 
 from lxml import etree
 
-from odoo import fields, models
+from odoo import _, fields, models, tools
+from odoo.exceptions import ValidationError
 
 
 class BaseCancelConfirm(models.AbstractModel):
@@ -16,6 +17,7 @@ class BaseCancelConfirm(models.AbstractModel):
 
     cancel_confirm = fields.Boolean(
         string="Cancel Confirmed",
+        default=lambda self: self._cancel_confirm_disabled(),
         copy=False,
         help="A flag signify that this document is confirmed for cancellation",
     )
@@ -24,6 +26,17 @@ class BaseCancelConfirm(models.AbstractModel):
         copy=False,
         help="An optional cancel reason",
     )
+
+    def _cancel_confirm_disabled(self):
+        key = "%s.cancel_confirm_disable" % self._name
+        res = self.env["ir.config_parameter"].sudo().get_param(key)
+        if not res:
+            return True
+        if res not in ("True", "False"):
+            raise ValidationError(
+                _("Configuration Error (%s), should be 'True' or 'False'") % key
+            )
+        return tools.str2bool(res)
 
     def open_cancel_confirm_wizard(self):
         xmlid = "base_cancel_confirm.action_cancel_confirm_wizard"
--- a/base_cancel_confirm/readme/CONFIGURE.rst
+++ b/base_cancel_confirm/readme/CONFIGURE.rst
@@ -1,0 +1,8 @@
+By default, the cancel confirm will be disabled (to ensure no side effect on other module unit test)
+
+To enable cancel confirm wizard, please add System Parameter (ir.config_parameter) for each extended module.
+
+For example,
+
+* sale_cancel_confirm, add `sale.order.cancel_confirm_disable = False`
+* purchase_cancel_confirm, add `purchase.order.cancel_confirm_disable = False`
--- a/base_cancel_confirm/tests/test_cancel_confirm.py
+++ b/base_cancel_confirm/tests/test_cancel_confirm.py
@@ -20,7 +20,9 @@ class TestCancelConfirm(common.SavepointCase):
         cls.tester_model = cls.env["ir.model"].search(
             [("model", "=", "cancel.confirm.tester")]
         )
-
+        cls.env["ir.config_parameter"].create(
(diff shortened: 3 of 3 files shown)


rousseldenis (commented):
@kittiu I'm not sure of this as it should be disabled by default.

The cancel operation should flow normally except you enable the feature(that breaks that flow as waiting for the wizard confirmation)

rousseldenis:
@kittiu

kittiu:
@rousseldenis yes, it is disabled by default on the extended module such as sale_cancel_confirm https://github.com/OCA/sale-workflow/pull/2022
But in this base module, I just want to make it compatible with existing extended modules, not yet corrected as sale_cancel_confirm, to be enabled by defaul (no system param).

I try to make it backward compat.
WDYT?

base_cancel_confirm/model/base_cancel_confirm.py
```diff
@@ -25,6 +27,19 @@ class BaseCancelConfirm(models.AbstractModel):
         help="An optional cancel reason",
     )
 
+    def _cancel_confirm_disabled(self):
+        key = "%s.cancel_confirm_disable" % self._name
+        res = self.env["ir.config_parameter"].get_param(key)
+        if not res:
+            return False
```
rousseldenis: @kittiu What I wanted to say is this is the default if no ir config parameter entry.

Should be true here

kittiu: You are genius ! :)

rousseldenis: > You are genius ! :)

Don't over estimate me ;-)

rousseldenis (approved):
Code review

base_cancel_confirm/model/base_cancel_confirm.py
kittiu: I think you are rock star now. Very active contributor 🥇

base_cancel_confirm/model/base_cancel_confirm.py
```diff
@@ -25,6 +27,19 @@ class BaseCancelConfirm(models.AbstractModel):
         help="An optional cancel reason",
     )
 
+    def _cancel_confirm_disabled(self):
+        key = "%s.cancel_confirm_disable" % self._name
+        res = self.env["ir.config_parameter"].get_param(key)
```
ps-tubtim:
```suggestion
        res = self.env["ir.config_parameter"].sudo().get_param(key)
```
@kittiu Users other than admin can't work.
Selection_002

kittiu: OK thanks!

ps-tubtim (approved):

kittiu:
@rousseldenis ok to merge?

kittiu:
@rousseldenis can we merge this?

rousseldenis:
> @rousseldenis can we merge this?

No superpowers here. @OCA/core-maintainers

base_cancel_confirm/model/base_cancel_confirm.py
```diff
@@ -25,6 +27,19 @@ class BaseCancelConfirm(models.AbstractModel):
         help="An optional cancel reason",
     )
 
+    def _cancel_confirm_disabled(self):
+        key = "%s.cancel_confirm_disable" % self._name
+        res = self.env["ir.config_parameter"].sudo().get_param(key)
+        if not res:
+            return True
+        if res == "True":
```
simahawk: you can use `tools.str2bool` here

base_cancel_confirm/model/base_cancel_confirm.py
```diff
@@ -25,6 +27,19 @@ class BaseCancelConfirm(models.AbstractModel):
         help="An optional cancel reason",
     )
 
+    def _cancel_confirm_disabled(self):
+        key = "%s.cancel_confirm_disable" % self._name
```
simahawk, line 31: I think `_enable` makes more sense, but it's likely a matter of taste :)

simahawk (commented):
Can you please remove the odoo version from the commit?

base_cancel_confirm/model/base_cancel_confirm.py
kittiu: We use disable for backward compatibility too, there are extended modules what should be enable by default.

base_cancel_confirm/model/base_cancel_confirm.py
kittiu: Ok, thank you!

kittiu: Fixed, thank you!

Saran440 (approved):

Saran440:
This PR is ready to merge. someone can merge, please?

simahawk (approved):
@kittiu LG, thanks.
One more thing, if you don't mind: please never include the odoo version in the commit -> when you port commits from one branch to another it makes little sense.
Also, if you keep the tag, improvements should be tagged w/ `IMP`.
Not a blocker of course :wink:

Saran440:
@simahawk Thank you for your advise, I rename commit without version.

PR #498, opened 2022-05-26
[13.0][FIX] document_quick_access_folder_auto_classification: remove dead code

LoisRForgeFlow:
Since the refactor to use edi_storage capabilities the cron is not needed
because the EDI and EDI backend crons do the job of retrieving the files
and create the exchange records to process and create the jobs to
process them. Actually the cron job in this module is failling in most of
the cases because it is using the storage "home" path instaed of the
input path.

cc @etobella 

@ForgeFlow


etobella (approved):

etobella:
You probably need to remove the tests....

LoisRForgeFlow:
@etobella better late than never, tests are reworked :)

JoanSForgeFlow (approved):
LGTM! :+1:

PR #500, opened 2022-06-02
[14.0][FIX] date_range: fix TZ

hparfr:
tz is added by odoo (in js) at some point so no need to handle it here.

Without this patch, with a tz set to Europe/Paris, a range starting at 01/01/2022, is then converted to in this module 01/01/2022 00:00 UTC+1 (= 31/12/2021 23:00) then (in some odoo js) 31/12/2021 22:00.


hparfr:
ping @florian-dacosta

lmignon (approved):

florian-dacosta (approved):
Tested, LGTM
Thanks for this important fix!

PR #509, opened 2022-07-06
[FIX] mass_editing: Removed non-existing column from migration query.

ddejong-therp:
The current migration script of module `mass_editing` uses a query that expects there to be a column `domain` in table `mass_editing`. I have not found this field in Odoo 13.0 or 14.0, so I removed it from the query.

### DIFF
--- a/mass_editing/migrations/14.0.1.0.0/pre-migrate.py
+++ b/mass_editing/migrations/14.0.1.0.0/pre-migrate.py
@@ -59,7 +59,7 @@ def migrate_mass_editing(env):
             'mass_edit',
             me.model_id,
             mo.model,
-            me.domain
+            NULL
         FROM mass_editing me
         LEFT JOIN ir_model mo ON (me.model_id = mo.id)
         """


pedrobaeza:
I'm suspecting you haven't included the module `mass_editing` in the intermediate versions, and thus such DB layout unexpected changes.

ddejong-therp:
That is somewhat true. I'm dealing with an enterprise migration here, which doesn't play well with OpenUpgrade of course. Enterprise requires me to upgrade to 13.0 in one go, and so the odoo 12.0 migration scripts of `mass_editing` never gets called. I've done that manually now, and that seems to 'work'. Except when I upgrade in 14.0.

Now I assumed it was the upgrade script that was wrong, considering the fact that I couldn't find the origin of the `domain` column anywhere. Not even in 11.0 or 10.0.
I can try a clean installation of Odoo and upgrade that to 14.0 without enterprise, and see if the same problem occurs...

pedrobaeza:
It's not only migration scripts, but regular ORM update process that changes DB layout, so I'm afraid you have to create specific migration scripts to deal with this situation or go the OpenUpgrade path.

ddejong-therp:
Alright.

PR #516, opened 2022-07-22
[15.0][FIX] document_quick_access_folder_auto_classification: remove dead code

LoisRForgeFlow:
Since the refactor to use edi_storage capabilities the cron is not needed
because the EDI and EDI backend crons do the job of retrieving the files
and create the exchange records to process and create the jobs to
process them. Actually the cron job in this module is failling in most of
the cases because it is using the storage "home" path instead of the
input path.

Forward port of #498

@ForgeFlow


JoanSForgeFlow (approved):
:+1:

DavidJForgeFlow (approved):
:+1:

etobella (approved):

PR #521, opened 2022-08-10
[13.0][ADD] announcement: New module

chienandalu:
Create announments:
admin

How users see them:
user

Logging:
logs

cc @Tecnativa TT38174


chienandalu:
@victoralmau @CarlosRoca13 can you review?

chienandalu:
BTW: Still struggling with the on log in popup being closed after full loading...

CarlosRoca13 (approved):
Tested on runbot! Nice solution for the problem!

announcement/models/announcement.py
```diff
@@ -0,0 +1,66 @@
...
+        if self.announcement_type == "specific_users":
+            self.user_group_ids = False
+        elif self.announcement_type == "user_group":
+            self.specific_user_ids = False
+
+    def _compute_read_announcement_count(self):
+        for announcement in self:
+            announcement.read_announcement_count = self.env[
+                "announcement.log"
+            ].search_count([("announcement_id", "=", announcement.id)])
```
victoralmau:
```suggestion
        logs = self.env["announcement.log"].read_group(
            [("announcement_id", "in", self.ids)],
            ["announcement_id"],
            ["announcement_id"],
        )
        result = {
            data["announcement_id"][0]: (data["announcement_id_count"]) for data in logs
        }
        for announcement in self:
            announcement.read_announcement_count = result.get(announcement.id, 0)
```

announcement/models/announcement.py
```diff
@@ -0,0 +1,66 @@
...
+
+    announcement_id = fields.Many2one(comodel_name="announcement")
+
+
+class Announcement(models.Model):
+    _name = "announcement"
+    _description = "User announcements"
+    _order = "sequence asc, id"
+
+    active = fields.Boolean()
```
victoralmau:
```suggestion
    active = fields.Boolean(default=True)
```

announcement/views/announcement_views.xml
```diff
@@ -0,0 +1,157 @@
...
+                                name="user_group_ids"
+                                string="Groups"
+                                options="{'no_create': True}"
+                                attrs="{'invisible': [('announcement_type', '!=', 'user_group')], 'required': [('announcement_type', '=', 'user_group')]}"
+                                widget="many2many_tags"
+                            />
+                            </group>
+                        <group name="right">
+                            <field name="notification_date" string="Announce at" />
+                            <field name="sequence" />
```
victoralmau:
```suggestion
```

victoralmau (changes requested):
Very good job!

announcement/models/announcement.py
chienandalu: No. It's on purpose. We don't want announcements released until they're unarchived.

announcement/views/announcement_views.xml
chienandalu: It's on purpose.

victoralmau (approved):
Code and functional review OK.

announcement/models/announcement.py
chienandalu: Thanks :) I was going to use this field only in the form, but thinking twice I get your code and I'll be showing the count as well in the tree :smile: :+1:

marcelsavegnago (approved):

announcement/README.rst
```diff
@@ -0,0 +1,117 @@
...
+#. Create a new one and define a title. This title will be shown in the announcement
+   header.
+#. Define the announcement scope:
+
+   - Specific users: manually select which users will see the announcement.
+   - User groups: users from the selected groups will be the ones to see the
+     announcement.
+#. Optionally you can set an announcement date to schedule the announcement. The
+   announcement won't show up until that date.
+#. Define the announcement body. You can use rich formatting and event paste your
```
pedrobaeza, line 50: This is missing the new feature "Valid until".

announcement/README.rst
```diff
@@ -0,0 +1,117 @@
...
+
+#. Go to *Discuss > Announcements*
+#. Create a new one and define a title. This title will be shown in the announcement
+   header.
+#. Define the announcement scope:
+
+   - Specific users: manually select which users will see the announcement.
+   - User groups: users from the selected groups will be the ones to see the
+     announcement.
+#. Optionally you can set an announcement date to schedule the announcement. The
```
pedrobaeza: Put this after the unarchive text

announcement/README.rst
```diff
@@ -0,0 +1,117 @@
...
+#. By default, the announcement will be archived. This is to prevent the announcement
+   to show up before time.
+#. Once the announcement is ready, unarchive it going to the *Actions* menu an choosing
+   the *Unarchive* option.
+
+Usage
+=====
+
+When a user in the scope of active announcements logs in, those will popup. The user
+has to mark them as read to continue working. If the announcement is set during this
```
pedrobaeza:
```suggestion
has to mark them as read to continue working. If the announcement is set during the
```

announcement/README.rst
```diff
@@ -0,0 +1,117 @@
...
+user session, the announcement will be eventually propted in the top bar. The user can
```
pedrobaeza:
```suggestion
user session, the announcement will be eventually prompted in the top bar on the right part. The user can
```

announcement/README.rst
```diff
@@ -0,0 +1,117 @@
...
+click on the unread announcements icon and the announcements will popup for the user
```
pedrobaeza:
```suggestion
click on the unread announcements icon (a speaker) and the announcements will popup for the user
```

announcement/README.rst
```diff
@@ -0,0 +1,117 @@
...
+to check them.
+
+For the announcement managers, going to *Discuss > Announcements* an clicking on an
```
pedrobaeza:
```suggestion
For the announcement managers, going to *Discuss > Announcements* and clicking on a
```

announcement/README.rst
```diff
@@ -0,0 +1,117 @@
...
+specific announcement, they can track which users have read the announcement.
+
+Known issues / Roadmap
+======================
+
+* It could be integrated in Discuss app to review past announcements.
```
pedrobaeza, line 76: Add  `* Log other information like geolocation, IP, browser agent, etc when marking announcement as read.`

announcement/models/announcement.py
```diff
@@ -0,0 +1,73 @@
...
+    _description = "Log user reads"
+    _order = "create_date desc"
+
+    announcement_id = fields.Many2one(comodel_name="announcement")
+
+
+class Announcement(models.Model):
+    _name = "announcement"
+    _description = "User announcements"
+    _order = "sequence asc, id"
```
pedrobaeza:
```suggestion
    _order = "notification_date, sequence asc, id"
```

announcement/models/announcement.py
```diff
@@ -0,0 +1,73 @@
...
+            ("specific_users", "Specific users"),
+            ("user_group", "User groups"),
+        ],
+        default="specific_users",
+        required=True,
+    )
+    specific_user_ids = fields.Many2many(
+        comodel_name="res.users",
+        domain=[("share", "=", False)],
+        inverse="_inverse_specific_user_ids",
```
pedrobaeza, line 34: Document this trick

announcement/models/announcement.py
```diff
@@ -0,0 +1,73 @@
...
+    )
+    user_group_ids = fields.Many2many(comodel_name="res.groups",)
+    notification_date = fields.Datetime()
+    notification_expiry_date = fields.Datetime()
+    read_announcement_count = fields.Integer(
+        compute="_compute_read_announcement_count",
```
pedrobaeza: Why not being stored?

announcement/models/announcement.py
```diff
@@ -0,0 +1,73 @@
...
+        )
+        result = {
+            data["announcement_id"][0]: (data["announcement_id_count"]) for data in logs
+        }
+        for announcement in self:
+            announcement.read_announcement_count = result.get(announcement.id, 0)
+
+    def copy(self, default=None):
+        """When duplicating an announcement, prevent it to show up before time"""
+        default = dict(default or {}, active=False)
```
pedrobaeza: Put `copy=False` in the field instead.

announcement/models/res_users.py
```diff
@@ -0,0 +1,62 @@
...
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    pendant_announcement_ids = fields.Many2many(
+        comodel_name="announcement", relation="pendant_announcement_user_rel",
+    )
+    read_announcement_ids = fields.Many2many(
+        comodel_name="announcement", relation="read_announcement_user_rel",
+    )
+    popup_announcements = fields.Boolean()
```
pedrobaeza, line 15: `store=False` ?

announcement/models/res_users.py
```diff
@@ -0,0 +1,62 @@
...
+        announcements = self.env["announcement"].browse(
+            {
+                x["id"]
+                for x in group_announcements
+                if any(
+                    [g for g in x["user_group_ids"] if g in self.env.user.groups_id.ids]
+                )
+            }
+        )
+        announcements |= self.env.user.pendant_announcement_ids.filtered(
```
pedrobaeza, line 43: Why this OR ? Can't be done with the initial search?

chienandalu: It's not an `OR`. We're joining the two types of announcements in a single recordset of unique records.

announcement/models/res_users.py
chienandalu: It's not a computed field. Would it be stored in cache?

announcement/models/res_users.py
pedrobaeza: And this is can't be done with a bigger domain?

pedrobaeza: I want to avoid the heavy field `pendant_announcement_ids` + later `filtered`

announcement/models/res_users.py
pedrobaeza: Yes, it is

chienandalu: I just tried and it doesn't seem to work for this purpose

announcement/models/res_users.py
chienandalu: But pendant announcement_ids will only contain a handful of records at the very best. It's not going to be that big.

pedrobaeza: OK then.

announcement/models/res_users.py
pedrobaeza: OK

announcement/models/announcement.py
```diff
@@ -0,0 +1,128 @@
...
+        self.allowed_user_ids = False
+        self.allowed_users_count = False
+        for announcement in self.filtered(
+            lambda x: x.announcement_type == "specific_users"
+        ):
+            announcement.allowed_user_ids = announcement.specific_user_ids
+            announcement.allowed_users_count = len(announcement.specific_user_ids)
+        for announcement in self.filtered(
+            lambda x: x.announcement_type == "user_group"
+        ):
```
simahawk, line 74: Is it really needed to loop 4 times (2 loops + filtered) on these records ?

announcement/models/announcement.py
```diff
@@ -0,0 +1,128 @@
...
+    @api.onchange("announcement_type")
+    def _onchange_announcement_type(self):
+        if self.announcement_type == "specific_users":
+            self.user_group_ids = False
+        elif self.announcement_type == "user_group":
+            self.specific_user_ids = False
+
+    def action_announcement_log(self):
+        """Return a set of fungible transient records to see altogether read logs
+        and pendant users"""
```
simahawk, line 99: Please try to keep docstrings short w/ 1 line + description (as for commits). See https://peps.python.org/pep-0257/#multi-line-docstrings

announcement/models/announcement.py
```diff
@@ -0,0 +1,128 @@
...
+        )
+        read_pendant_log += self.env["read.announcement.wizard"].create(
+            [{"user_id": user.id, "read_state": "pendant"} for user in pendant_users]
+        )
+        return {
+            "type": "ir.actions.act_window",
+            "res_model": "read.announcement.wizard",
+            "views": [[False, "tree"]],
+            "domain": [("id", "in", read_pendant_log.ids)],
+            "context": dict(self._context, create=False, group_by=["read_state"]),
```
simahawk, line 126: self.env.context

announcement/models/res_users.py
```diff
@@ -0,0 +1,62 @@
...
+            lambda x: (not x.notification_date or x.notification_date <= now)
```
simahawk, line 44: this could have been done more readable

announcement/models/res_users.py
```diff
@@ -0,0 +1,62 @@
...
+            and (not x.notification_expiry_date or x.notification_expiry_date >= now)
+        ).sorted(lambda k: k.sequence)
+        return announcements.read()
```
simahawk, line 47: here you are reading again everything from the db, can't you read only what is needed from the recordset?

simahawk (commented):
I'm late to the party :sweat_smile: 
1st of all: thanks for this useful module. 
I left some tech remarks, maybe you want to have a look for your next iteration.

One more thing: please DO NOT include the odoo version in the commits. 
Is useless and misleading when forward/back porting changes from version to version.

chienandalu:
Hi @simahawk I'm attending some of your remarks here: https://github.com/OCA/server-ux/pull/525

PR #525, opened 2022-08-22
[13.0][FIX] announcement: assorted fixes

chienandalu:
- Now that the reads count is stored, we need to add a depends that triggers the recomputation. To do so, we need to link announcements and read logs.
- Simplified popup on log in mechanism.
- The notification_expiry_date field was duplicated in the form view
- Add some thechnical suggestions
- Change the term `pendant` to a more convenient `unread`
- Regular users can't read announcements with dates in the future or which expiration date has passed nor archived announcements.
- Fix typos

cc @Tecnativa TT38174 TT38665


announcement/models/announcement.py
```diff
@@ -95,8 +98,7 @@ def _onchange_announcement_type(self):
             self.specific_user_ids = False
 
     def action_announcement_log(self):
-        """Return a set of fungible transient records to see altogether read logs
-        and pendant users"""
+        """See altogether read logs and pendant users"""
```
pedrobaeza:
```suggestion
        """See altogether read logs and pending users"""
```

pedrobaeza (commented):
Please change everywhere `pendant` by `pending`:

https://twitter.com/ggellatly/status/1561688338859700225?s=20&t=wZrpw5MFXkPsZWeCaYOzaw

gdgellatly:
> Please change everywhere `pendant` by `pending`:
> 
> https://twitter.com/ggellatly/status/1561688338859700225?s=20&t=wZrpw5MFXkPsZWeCaYOzaw

Hi, Just further to this, why not just `Unread`, or am I not understanding what this field is doing.

pedrobaeza:
Yes, unread is valid as well and fits well in the current context.

announcement/models/announcement.py
chienandalu: Fixed changing it to *unread* as @gdgellatly suggested

announcement/models/res_users.py
```diff
@@ -18,6 +18,8 @@ class ResUsers(models.Model):
     def announcement_user_count(self):
         """The js widget gathers the announcements from this method"""
         now = fields.Datetime.now()
+        # It would be better to rely on record rules, but then announcement managers
+        # would see every announment, which would be annoying.
```
pedrobaeza:
```suggestion
        # would see every announcement, which would be annoying.
```

pedrobaeza (approved):

announcement/models/res_users.py
chienandalu: Oops... it wasn't the only one...

chienandalu:
@CarlosRoca13 @victoralmau your review will be very welcome :)

victoralmau (approved):

pedrobaeza:
As record rule is noupdate, you have to add migration script.

chienandalu:
> As record rule is noupdate, you have to add migration script.

Done

announcement/models/res_users.py
```diff
@@ -40,23 +35,15 @@ def announcement_user_count(self):
...
         )
-        announcements |= self.env.user.pendant_announcement_ids.filtered(
-            lambda x: (not x.notification_date or x.notification_date <= now)
-            and (not x.notification_expiry_date or x.notification_expiry_date >= now)
-        ).sorted(lambda k: k.sequence)
-        return announcements.read()
+        # Unread announcements are directly linked to the user. Normally, only a
+        # handful of records will be evaluated at best.
+        announcements |= self.env.user.unread_announcement_ids.filtered("in_date")
+        return announcements.sorted(lambda k: k.sequence).read(["name", "content"])
```
simahawk:
```suggestion
        return [
            {"name": rec.name, "content": rec.content}
            for announcements.sorted(lambda k: k.sequence)
        ]
```

simahawk (approved):
I would not go for 2.0.0 as these are fixes on a new immature module. Anyway, do it as you prefer :)

announcement/models/res_users.py
chienandalu: Let's save a query, yes :smile: :+1:

chienandalu:
> I would not go for 2.0.0 as these are fixes on a new immature module. Anyway, do it as you prefer :)

Some of the changes are major. Despite the short lifespan, the module was being tested in production already.

PR #364, opened 2021-07-08
[14.0][FIX] - base_search_custom_field_filter: fix filter init

sbejaoui:
When using the custom field filter, the user gets an error message:

```ValueError: Invalid field res.partner.ir_ui_custom_filter_3 in leaf ('ir_ui_custom_filter_3', 'ilike', 'azure')```

PR #526, opened 2022-08-23
[14.0][FIX] base_search_custom_field_filter: Resolve the error when the expression has related fields (dot notation).

victoralmau:
**Steps to reproduce the error**:
- Install `contacts` addon
- Go to Settings > Technical > User Interface > Custom Field Filters and create a new record with the following data: 
- Model: Contact
- Name: Country
- Expression: state_id.country_id
- Go to Contacts

Before this change there was an error (field does not exist).

@Tecnativa TT38645

### DIFF
--- a/base_search_custom_field_filter/models/base.py
+++ b/base_search_custom_field_filter/models/base.py
@@ -1,5 +1,6 @@
 # Copyright 2020 Tecnativa - Carlos Dauden
 # Copyright 2020 Tecnativa - Pedro M. Baeza
+# Copyright 2022 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 from lxml import etree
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I think the fix should be the contrary: to restore how it was in this part in v13:

https://github.com/OCA/server-ux/blob/05573aed83d364bd0470f601241a72109ff174b5/base_search_custom_field_filter/models/base.py#L30

Please check also if this problem is present without the JS code:

https://github.com/OCA/server-ux/commit/01e29cbd1050550c1006f97ceaa6a6dd3085be22

victoralmau:
I think it is not necessary to restore the mentioned from v13.

What you indicate was commented here that it did not happen https://github.com/OCA/server-ux/pull/364#issuecomment-1147100106

This is an example of the current operation
ejemplo-local-v14

pedrobaeza:
Thanks for the info on last point. And what about the other?

victoralmau:
> I think the fix should be the contrary: to restore how it was in this part in v13:
> 
> https://github.com/OCA/server-ux/blob/05573aed83d364bd0470f601241a72109ff174b5/base_search_custom_field_filter/models/base.py#L30

Making this change would cause the error https://github.com/OCA/server-ux/pull/364 to appear again.

The error that currently exists occurs in `.js` specifically here https://github.com/odoo/odoo/blob/14.0/addons/web/static/src/js/control_panel/control_panel_model_extension.js#L912 because the field does not exist.

pedrobaeza (approved):
OK, if the field is inferred from dot notation, then it's good to go.

PR #527, opened 2022-08-30
[13.0][FIX] announcement: log read just once

chienandalu:
If a user has an announcement open in multiple windows, the read will be logged multiple times. The relevant one is only the first one.

cc @Tecntiva TT38705

please review @victoralmau @pedrobaeza

### DIFF
--- a/announcement/models/res_users.py
+++ b/announcement/models/res_users.py
@@ -52,5 +52,8 @@ class ResUsers(models.Model):
         """Used as a controller for the widget"""
         announcement = self.env["announcement"].browse(int(announcement_id))
         self.env.user.unread_announcement_ids -= announcement
-        self.env.user.read_announcement_ids += announcement
-        self.env["announcement.log"].create({"announcement_id": announcement.id})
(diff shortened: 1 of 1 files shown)


victoralmau (approved):
Could you add a small test to validate this behavior?

announcement/models/res_users.py
```diff
@@ -52,5 +52,8 @@ def mark_announcement_as_read(self, announcement_id):
         """Used as a controller for the widget"""
         announcement = self.env["announcement"].browse(int(announcement_id))
         self.env.user.unread_announcement_ids -= announcement
-        self.env.user.read_announcement_ids += announcement
-        self.env["announcement.log"].create({"announcement_id": announcement.id})
+        # Log only the first time. Users with the announment in multiple windows would
```
pedrobaeza:
```suggestion
        # Log only the first time. Users with the announcement in multiple windows would
```

chienandalu: Fixed

PR #529, opened 2022-09-09
[FIX] mass_editing: fix migration to 14.0

sebalix:


pedrobaeza:
Isn't simpler to add in the XML `<field name="inherit_id" eval=False> <!-- make sure the field is emptied -->` ?

sebalix:
Yes we can do that of course, I thought as it's a migration issue I didn't want to put garbage in functional code, but as you wish.

sebalix:
Maybe I prefer to keep things like this, because we are migrating directly from 13.0 to 15.0 here, but we retrieve the migration scripts of 14.0 with `git format-patch` (by targeting `migrations` folder of the addons). By putting this fix in the 14.0 view directly, this one won't get updated during the migration.

ivantodorovich (approved):

pedrobaeza:
Well, that approach is condemned to fail IMO, as migration scripts rely on the regular Odoo update mechanism for the things they expect to have, so if you skip one version, there may be a missing DB column, some data not computed, etc. But yeah, it's the enterprise price (or enterprice as I have seen around)!

sebalix:
Yes of course. In that case we patch ourself the migration scripts to make it working, but often they work as-is.
I think it's the same issue with OpenUpgrade: if you migrate from 12.0 to 15.0, you don't bother to enable/migrate all OCA addons for 13.0+14.0 versions while you target the 15.0.

pedrobaeza:
>  if you migrate from 12.0 to 15.0, you don't bother to enable/migrate all OCA addons for 13.0+14.0 versions while you target the 15.0.

Every OCA module already migrated to an intermediate version should be included in the addons path for that step. If not, errors may arise, or worst, no error, but inconsistent data that later leads to problems.

PR #537, opened 2022-10-10
[14.0][ADD] base_delegation

remi-filament:
Looking for a solution to add delegations and track delegation action, I could not find one, so I propose that solution with associated purchase functionalities.
Any feedback welcome !


kittiu:
Some picture short video would be nice :)
What are you trying to achieve?

remi-filament:
Hi @kittiu 

I should try to provide some pictures / video.
In the meantime, here is what I try to achieve : 

In many org, you have one of these 2 expectations, or both :
- people want to delegate their rights to validate / sign permanently to some other (for instance a director delegating holidays approval to her secretary)
- people want to delegate their rights to validate / sign temporarily (for instance a manager delegating his validation rights during its holidays)

I have found a couple of modules that adds Odoo roles for that, however we also want both traceability and accountability on these delegations, so that we need to trace that the delegate signed / validated on behalf of delegator.

This module intends to provide list of delegations that can then be used in functional specific modules (same way as tier_validation is built).
One example on purchase here : https://github.com/OCA/purchase-workflow/pull/1591

This is a proposed tentative implementation for the need above, I am open to any feedback / improvements !

kittiu:
Hello @remi-filament I read the code and try to understand what it is doing.

First of all, I just got this error, when try to "Approve As", after I install your module in fresh database.

image

But, despite that error, I still don't get why do we need to make this base module. I think the delegate will be useful only on Model that require approval, like PO, correct? So, it is not that generic, is it?
We still need to write a lot of code on the extended module, in this case, purchase_tier_delegation (the idea of having base module is that, the extended module also no coding required).

IMHO, if you need the delegation feature on PO approval, why not make a single specific module, i..e, purchase_approval_delegation.

**Note:** The word "Tier" also not make much sense to me, as it is not step by step approval thing, is it?

Note sure I understand it OK, let's see how other is thinking :)

remi-filament:
Hi @kittiu , thanks for looking into it !
I fixed the error you reported on purchase module.

Regarding your comments :
- about not needing base module, my idea is that if delegations are used in many modules (holidays, purchase, sale, expense, etc.) you can get a unique place where to list delegations (and therefore not redefining these in each specific module)
- about the naming tier, indeed it is a mistake of mine (I understood tier as third party which is incorrect translation from my side). I should rename it !

kittiu:
> Hi @kittiu , thanks for looking into it ! I fixed the error you reported on purchase module.
> 
> Regarding your comments :
> 
> * about not needing base module, my idea is that if delegations are used in many modules (holidays, purchase, sale, expense, etc.) you can get a unique place where to list delegations (and therefore not redefining these in each specific module)
> * about the naming tier, indeed it is a mistake of mine (I understood tier as third party which is incorrect translation from my side). I should rename it !

Though I not sure if I will be ok with the base module still ;) but if you can try make base module more base feature builtin (not just a data model) and make extending module more breeze to develop, I think it may be ok.

But it is my opinion only, we can wait for others opinion too :)

Also, I can see non-English in your code, you can make it English? Thanks!

remi-filament:
I fixed the non-English terms (I think all of them), do not hesitate to tell me in case you still find some.
I should wait for more opinions then before changing name or structure.
Thanks for your time !

bosd:
Interesting module! I would like to test it on runboat. But the runboat is not working.
@remi-filament Can you push an empty commit to retrigger the github actions?

bosd:
The problem with the runbot seems to be on my end. Chrome had some issues.
I like the idea of being able to manage the delagations in a central place.

Much more convenient then forwarding an approval.
In a use case where an specific user is set to do the validation. I doubt the will login to odoo in their holiday to forward the approval to an colleague. :)

Just a thought....
Some validation rules can be complex..and spread across multiple models..
Would it be simpler to do the delegation on an user level?
e.g. When `admin` is on leave, user `demo` will temporarily have the same validation rights on the set period.

Sticking to the regular workflow in the validation modules. So instead of the admin user getting the notifiction that his approval is needed, the notification is sent to the demo user.

Oops, maybe I'm also mislead by the name of this module. It does not seem to be about temporary delagating rights for tier validations.

remi-filament:
Hi @bosd I pushed an empty commit.
Regarding your comments, indeed it is not (yet ?) linked to tier validation, but the idea is that central delegation could be used for anything if it makes sense for others !
About notifications, it is a good idea, although in theory one could delegate rights to more than one people so that would be notifications to all (and probably only one of those would validate, so we could discuss if this is appropriate)

bosd:
> About notifications, it is a good idea, although in theory one could delegate rights to more than one people so that would be notifications to all 
I see, that would become complex very quickly.

I did'nt manage to do a `validate as` on a purchase order. (On the runboat of the other pr.)
Maybe I don;t understand this module :)
(There was no validate as button).

For now, I don't think I have enough knowledge and experience to have an educated opinion on this one.
From the use cases I have in mind right now, I think the user role module will fulfill the needs.

bosd:
Can this one be re-opened?

bosd:
@remi-filament Can you please fix pre-commit? :pray:

remi-filament:
Hi @bosd 
I just updated my commit and renamed also the module to base_delegation (to remove the incorrect _tier part).
I also updated https://github.com/OCA/purchase-workflow/pull/1591 accordingly

Issue #496, opened 2022-05-12
[base_tier_validation] BAD PERFORMANCE to compute and search field rejected


celm1990:
## Module
base_tier_validation
## Describe the bug
On Database with more than 80K records, at open sidebar and click Pending Reviews is very very slow
![image](https://user-images.githubusercontent.com/7775116/168171237-4e67697f-6202-4fe4-88d5-cd11f1f52911.png)

## To Reproduce
**Affected versions**:
All, I tested on V13
- on SideBar click to open pending records this call to search with domain [[["can_review", "=", true]]](https://github.com/OCA/server-ux/blob/13.0/base_tier_validation/static/src/js/systray.js#L147) 

![image](https://user-images.githubusercontent.com/7775116/168171403-c8fb08c8-551b-474a-bf97-3594d1e74d64.png)

- `can_review `is field  `computed `and implement [search function](https://github.com/OCA/server-ux/blob/13.0/base_tier_validation/models/tier_validation.py#L90), field `rejected ` is also computed and implement [search funcion](https://github.com/OCA/server-ux/blob/13472458f3897daf995c1ca99fa5c4e4cbbe0586/base_tier_validation/models/tier_validation.py#L121-L127) 
for `sale.order` this function search is done with state ["draft", "sent", "to approve"](https://github.com/OCA/sale-workflow/blob/3883f38a79a725acc58ee57b90b3944fe03b0aa1/sale_tier_validation/models/sale_order.py#L10)
On my case search is done with 30K records approx, but when use filtered for field `rejected `is where is more slower see [_compute_validated_rejected](https://github.com/OCA/server-ux/blob/13472458f3897daf995c1ca99fa5c4e4cbbe0586/base_tier_validation/models/tier_validation.py#L171-L175)

![image](https://user-images.githubusercontent.com/7775116/168172765-6c70c7e2-455f-49d0-95a0-e23abfd05815.png)

**Expected behavior**
Better performance at search records, fields computed not storable is very slower

Optimize _compute_validated_rejected and use SQL on some cases(if possible)
Try convert Fields to store=True

Optimize _search_can_review
![image](https://user-images.githubusercontent.com/7775116/168180872-ea39dd49-795f-4125-a0a7-7574bb27dccc.png)

https://user-images.githubusercontent.com/7775116/168182672-122d4b31-52c8-4d14-b42b-1103794371b3.mp4

rousseldenis:
Indeed, this has to be improved.

If I have time, I will try to fix it

[...]

PR #540, opened 2022-10-19
[16.0] base_tier_validation: migration to 16

bizzappdev:


celm1990:
Any plan to improve performance??
please see https://github.com/OCA/server-ux/issues/496

bizzappdev:
> Any plan to improve performance?? please see #496

Not in this MR. 
Prio is the Migration first and then might be new MR for improvement of performance.

Bart-dh:
#534

StefanRijnhart:
@bizzappdev can you tell us what still needs to be done?

bizzappdev:
> @bizzappdev can you tell us what still needs to be done?

only testing from the QA team was remaining. now I have confirmation from the QA team also.

it is ready to review

StefanRijnhart:
@bizzappdev thanks! In that case, please check the failing pre-commit linting.

base_tier_validation/__manifest__.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2017-19 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Base Tier Validation",
+    "summary": "Implement a validation process based on tiers.",
+    "version": "16.0.0.0.0",
```
StefanRijnhart: As per convention
```suggestion
    "version": "16.0.1.0.0",
```

StefanRijnhart:
@bizzappdev do you know of any of the OCA tier validation implementation modules for 16.0 to test this one with? A WIP branch or PR perhaps?

bizzappdev:
> @bizzappdev do you know of any of the OCA tier validation implementation modules for 16.0 to test this one with? A WIP branch or PR perhaps?

@StefanRijnhart we tested sale_tier_validation from v15. we still have to prepare the MR for that. also, automated test cases worked properly.

StefanRijnhart:
Please check the js linting
```
 error  Parsing error: 'import' and 'export' may appear only with 'sourceType: module'
 ```

**Comment by bizzappdev** [comment]
> Please check the js linting
> 
> ```
>  error  Parsing error: 'import' and 'export' may appear only with 'sourceType: module'
> ```

I really do not have much idea about how to fix this. as in the entire odoo also they have imported like this only. as some suggestions are changing .eslintrc file. but that will be needed at 16.0 level.

File: base_tier_validation/views/res_config_settings_views.xml [hunk 2]
```diff
@@ -0,0 +1,81 @@
...
+            <xpath expr="//div[@id='companies']" position="inside">
+                <div id="base_tier_validation">
+                    <h2>Tier Validation</h2>
+                    <div
+                        class="row mt16 o_settings_container"
+                        name="base_tier_validation_option_setting_container"
+                    >
+                        <div
+                            class="col-12 col-lg-6 o_setting_box"
+                            id="module_base_tier_validation_formula"
```
StefanRijnhart, line 20: All these modules are not yet migrated, and some of them have not even been migrated to 15 yet, so offering these options here seems misleading. Maybe these options should be commented out here and reenabled in each of these modules' migration PRs?

StefanRijnhart:
> I really do not have much idea about how to fix this. as in the entire odoo also they have imported like this only. as some suggestions are changing .eslintrc file. but that will be needed at 16.0 level.

You can fix the errors by renaming the .js files to .esm.js files as mentioned in the 15.0 migration issue (https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0#tasks-to-do-in-the-migration). This will trigger further eslint messages that need to be followed up.

StefanRijnhart (changes requested):
Thanks for the huge work migrating this module and all its javascript! I tested the module successfully. The code still needs some work on the JS linting.

rven:
``> Thanks for the huge work migrating this module and all its javascript! I tested the module successfully. The code still needs some work on the JS linting.

@StefanRijnhart
We discovered the same problem with the linting

I think the solution for the JS linting is the following:
```yml
parserOptions:
  ecmaVersion: 2019

overrides:
  - files:
      - "**/*.esm.js"
    parserOptions:
      sourceType: module
```
Should be adapted to
```yml
parserOptions:
  ecmaVersion: 2019
  sourceType: module

overrides:
  - files:
      - "**/*.esm.js"
    parserOptions:
      sourceType: module
```
After this change, the linting error will dissapear.

pedrobaeza:
JS files should be renamed to `.esm.js` if they use the new ES6 syntax.

bizzappdev:
> > I really do not have much idea about how to fix this. as in the entire odoo also they have imported like this only. as some suggestions are changing .eslintrc file. but that will be needed at 16.0 level.
> 
> You can fix the errors by renaming the .js files to .esm.js files as mentioned in the 15.0 migration issue (https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0#tasks-to-do-in-the-migration). This will trigger further eslint messages that need to be followed up.

Pre-commits are fixed. you can review again.
code coverage is already more than 90%. should we increase the coverage?

bizzappdev:
Pre-commits are fixed. you can review it again.
code coverage is already more than 90%. should we increase the coverage?
@StefanRijnhart

base_tier_validation/__manifest__.py
StefanRijnhart: Would you consider changing this?

StefanRijnhart:
Thanks for the js updates! Looking good. Please check my version number comment. It's always good to improve coverage, but it's going to be hard to make that 96% mark. I had a look at what's still missing coverage, and I don't see anything critical.

base_tier_validation/__manifest__.py
bizzappdev: done

StefanRijnhart (approved):
Thanks! Tested + code reviewed.

alan196 (approved):
Tested functionally 👍

alan196:
I tested the module in a custom model, and I get an issue when I try to click on the notification in the tray menu.

It looks that is an issue of Odoo, I made the following PR https://github.com/odoo/odoo/pull/108609 to fix this issue.

I post this in case you face the same issue 😄

PR #548, opened 2022-11-02
[12.0][FIX] base_tier_validation: Ignore active_test coming from upstream methods

HviorForgeFlow:
The use case is when filtering account invoice, active_test = False is set in the context down-streaming the context to the search inside evaluate tier.

Not happening on v14. Need to be check for v13

CC @ForgeFlow


LoisRForgeFlow (approved):

MiquelRForgeFlow:
@HviorForgeFlow please rebase

MiquelRForgeFlow (approved):

PR #556, opened 2022-11-23
[15.0][FIX] base_export_manager: Remove default because it is redundant

victoralmau:
Remove default because it is redundant

`WARNING odoo.fields: Redundant default on ir.exports.line.model1_id`

@Tecnativa


pedrobaeza (approved):

pedrobaeza:
Fast tracking this:

/ocabot merge patch

PR #560, opened 2022-11-29
[15.0][IMP] mass_editing: Do not re-use demo and admin users to avoid incompatibilities with data from those users.

victoralmau:
Do not re-use demo and admin users to avoid incompatibilities with data from those users.

If module `mass_mailing_partner` was installed the tests showed the error https://github.com/OCA/social/blob/15.0/mass_mailing_partner/models/res_partner.py#L41-L47 related to admin user.

@Tecnativa

### DIFF
--- a/mass_editing/tests/test_mass_editing.py
+++ b/mass_editing/tests/test_mass_editing.py
@@ -6,7 +6,7 @@
 from ast import literal_eval
 
 from odoo.exceptions import ValidationError
-from odoo.tests import Form, common
+from odoo.tests import Form, common, new_test_user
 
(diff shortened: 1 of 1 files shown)


legalsylvain (commented):
Maybe this patch should be ported upstream

https://github.com/OCA/server-ux/tree/16.0/server_action_mass_edit

pedrobaeza (approved):

pedrobaeza:
@legalsylvain do you agree with the change then?

legalsylvain (approved):
Yes

PR #561, opened 2022-11-30
[16.0][IMP] server_action_mass_edit: Do not re-use demo and admin users to avoid incompatibilities with data from those users.

victoralmau:
FWP from 15.0: https://github.com/OCA/server-ux/pull/560

Do not re-use demo and admin users to avoid incompatibilities with data from those users.

If module `mass_mailing_partner` was installed the tests showed the error https://github.com/OCA/social/blob/15.0/mass_mailing_partner/models/res_partner.py#L41-L47 related to admin user.

@Tecnativa

### DIFF
--- a/server_action_mass_edit/tests/test_mass_editing.py
+++ b/server_action_mass_edit/tests/test_mass_editing.py
@@ -6,7 +6,7 @@
 from ast import literal_eval
 
 from odoo.exceptions import ValidationError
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
You should change the module reference in the commit/PR message

legalsylvain (approved):
/ocabot merge patch

PR #565, opened 2022-12-09
[15.0][FIX] date_range: fix TZ

vincent-hatakeyama:
tz is added by odoo (in js) at some point so no need to handle it here.

Port of #500 for Odoo 15.


flotho (approved):
easy port LGTM
thanks

lmignon (approved):

PR #582, opened 2023-01-19
[14.0][IMP] sequence_check_digit: add ean

wouitmil:
ref: qoq-1346

### DIFF
--- a/sequence_check_digit/__manifest__.py
+++ b/sequence_check_digit/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "Check Digit on Sequences",
-    "version": "14.0.1.0.2",
+    "version": "14.0.1.1.0",
     "category": "Reporting",
     "website": "https://github.com/OCA/server-ux",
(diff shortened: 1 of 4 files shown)


sequence_check_digit/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     "name": "Check Digit on Sequences",
-    "version": "14.0.1.0.2",
+    "version": "14.0.1.1.0",
```
simahawk, line 6: don't change the version nr, it will get bumped by the ocabot ;)

simahawk (changes requested):
Can you please amend your commit msg on the next push?
Something like `sequence_check_digit: support ean` or smth similar will work ;)
You can add `[IMP] in front of it if you like.
Guidelines here https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

gurneyalex (approved):

gurneyalex:
ocabot merge minor

PR #592, opened 2023-02-06
[13.0][ADD] base_action_visibility_restriction

JordiMForgeFlow:


legalsylvain (approved):
thanks ! (code review / no test).

Just one point. (non blocking) : 
- could you add in the description, a use case where the module is usefull. (for the time being, using group_ids is enough for me, and I don't see the interest of the excluded_group_ids). maybe other people will ask the same question.

JordiMForgeFlow:
Hi @legalsylvain 

In our case we are already using other similar modules like base_menu_visibility_restriction, as they are really helpful when you just want to restrict access to a specific user or small group of users. Then you only need to handle one group to apply the restriction and you just need to assign it to the few users you want to restrict.

In this case we also needed to restrict access to the actions (mainly the server actions, but I have also added the feature for the window actions).

JaumeBforgeFlow (approved):
Code + functional review :+1:

LoisRForgeFlow (approved):

AaronHForgeFlow (approved):
Code review LG!

PR #600, opened 2023-02-16
[16.0][FIX] pre-commit warnings

CRogos:
The branch 16.0 pre-commit job is not finishing without warnings

### DIFF
--- a/base_tier_validation/static/src/js/ir_model.esm.js
+++ b/base_tier_validation/static/src/js/ir_model.esm.js
@@ -1,7 +1,7 @@
 /** @odoo-module **/
 
-import {registerModel} from "@mail/model/model_core";
 import {attr, one} from "@mail/model/model_field";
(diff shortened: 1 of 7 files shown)


setup/_metapackage/VERSION.txt
```diff
@@ -1 +0,0 @@
-16.0.20230119.0
```
pedrobaeza: Why are these files removed?

CRogos: I just checked out branch 16.0 and executed pre-commit. 
Good question why it is removed?

CRogos: I still don't know what is wrong with my other computer but on this these files are not deleted.

pedrobaeza (approved):
Fast-tracking to remove warnings:

/ocabot merge patch

PR #601, opened 2023-02-17
[16.0][MIG] mass_editing

CRogos:
Migration using oca-port tool


legalsylvain (changes requested):
The module bas been renamed and existe in v16

https://github.com/OCA/server-ux/tree/16.0/server_action_mass_edit

PR #472, opened 2022-03-15
[15.0][MIG] mass_operation_abstract

ChrisOForgeFlow:
Standard Migration
@ForgeFlow

PR #603, opened 2023-02-23
[15.0][MIG] mass_operation_abstract

MiquelRForgeFlow:
Standard migration.


legalsylvain:
why do you migrate this module ?

MiquelRForgeFlow:
Why not? It was on my list of modules to migrate.

legalsylvain:
> Why not? It was on my list of modules to migrate.

;-). Well, that module doesn't do anything but is usefull as a dependency. my question is for which module ? Note : AFAIK, this module is obsolete for V16. (But I don't know for V15).

LoisRForgeFlow:
Indeed, this migration is not needed, see https://github.com/OCA/server-ux/pull/472#issuecomment-1185593749.

MiquelRForgeFlow:
Nice

PR #604, opened 2023-02-23
[FIX] Remove unneeded external_dependencies odoo_test_helper

rven:
The python package odoo_test_helper is only needed on the CI environments.


rven:
Shouldn't the python package odoo_test_helper being installed by the CI?

lmignon (changes requested):
@rven  You must add the dependency into a test-requirements.txt file to be installed by the CI

etobella:
The right way, would be your changes and adding it to test-requirements.txt file

lmignon (approved):

PR #605, opened 2023-02-27
Improving conditions to double check if the operation is under valida…

BT-rmartin:
…tion

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -243,7 +243,9 @@ class TierValidation(models.AbstractModel):
                         )
                     )
             if (
+                # Only check if a state change is included in vals                
                 rec.review_ids
+                and self._state_field in vals
                 and getattr(rec, self._state_field) in self._state_from
                 and not vals.get(self._state_field)
                 in (self._state_to + [self._cancel_state])


LoisRForgeFlow (changes requested):
This is not correct. If you add that condition any change could be done to a document under validation.

Plus, you just need to look 2 lines below to see that the exact opposite condition is there: `and not vals.get(self._state_field)` so your change will create a condition imposible to meet `and self._state_field in vals and not vals.get(self._state_field)`

BT-rmartin:
@LoisRForgeFlow Sorry for disturbing, it was not yet ready to test, it should have been a draft PR

BT-rmartin:
@LoisRForgeFlow The origin of the change is that as it is right now you cannot change almost anything, just "message_follower_ids", "access_token" as defined in _get_under_validation_exceptions. 
 This makes the module unusable, as for instance to confirm the invoice after validation, Odoo automatically adds some other values like "invoice_date", "invoice_date_due", "line_ids", "tax_country_id", "payment_reference". 
All those should be whitelisted as well, but even with more control, as otherwise you can change anything in line_ids. 
Conceptually, after validation, some changes should be also possible by the accountant at least, like changing the account of a line, etc. Or dummy changes as changing the reference of the Customer Reference.
What's your opinion? Which would be the best approach to get that behaviour?

I would start with something like this
image

LoisRForgeFlow:
Hi @BT-rmartin 

`_get_under_validation_exceptions` is exactly meant for that, for whitelisting more fields in each specific submodule. E.g: for generic fields (say it is invoice_date, then it goes to account_move_tier_validation - it is a bad example because I think invoice_date is arguable so it is probably better that you do it in a customization) or custom modules for custom fields (say you have a x_custom_field and you want to allow to modify it even on invoices under validation, then you should create your own custom module extending _get_under_validation_exceptions to add that field to the whitelist).

You also suggest a more extensible and powerful system in which you can whitelist via configuration and based on groups, I'm open to any proposal implementing this as long as it respect current installations. 

Regards,

PR #533, opened 2022-09-30
[15.0][IMP] base_tier_validation: Avoid tier validation (error) when only browsing attachments

bobslee:

PR #607, opened 2023-03-01
[14.0][FIX] base_tier_validation - check state field exists

### LINKED ISSUE #473: [15.0] Base tier validation - Incorrect: The operation is under validation.
## Module
base_tier_validation

## Describe the bug
def write is inherited to check for state changes, one of the checks is: 
`and not vals.get(self._state_field) in (self._state_to + [self._cancel_state])`

If vals doesn't contain state the warning is shown.

I suggest adding a check to see if the write should be validated, something like:
```
            if (
                rec.review_ids
                **and self._state_field in vals**
                and getattr(rec, self._state_field) in self._state_from
                and not vals.get(self._state_field) in (self._state_to + [self._cancel_state])
                and not rec._check_allow_write_under_validation(vals)
            ):
```

ilyasProgrammer:
fix https://github.com/OCA/server-ux/issues/473


LoisRForgeFlow (changes requested):
@ilyasProgrammer I don't see this correct for the same reasons I exposed here https://github.com/OCA/server-ux/pull/605.

tarbib:
Hello @ilyasProgrammer , 
Not a developer here :  I get a "The operation is under validation" validation error when users try to send a message / log note with an attachment
Wondering if your fix would cover this issue as well?
Thanks
Screenshot 2023-03-09 at 09 53 28

bosd:
@tarbib  Check #533

LoisRForgeFlow:
Finally solved in https://github.com/OCA/server-ux/pull/612.

We can close this one, please, reopen if you think I'm mistaken.

PR #612, opened 2023-03-09
[14.0][IMP] base_tier_validation: skip_validation_check

ilyasProgrammer:
Some modules have compatibility issues with this one. E.g. delivery_auto_refresh.
_get_under_validation_exceptions() does not cover cases when you dont want to exclude field from a check in general, but only for certain cases, like specific methods calls. Or when multiple fields are changed, but validation check should be skipped.
I suggest to add context  check to avoid "The operation is under validation." raise. 
Im open to any alternative solutions. 
Related:
https://github.com/OCA/server-ux/issues/473
https://github.com/OCA/server-ux/pull/605
https://github.com/OCA/server-ux/pull/607#issuecomment-1461595114

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -274,6 +274,7 @@ class TierValidation(models.AbstractModel):
                 rec.review_ids
                 and rec._check_tier_state_transition(vals)
                 and not rec._check_allow_write_under_validation(vals)
+                and not rec._context.get("skip_validation_check")
             ):
                 raise ValidationError(_("The operation is under validation."))
         if vals.get(new_self._state_field) in (


francesco-ooops:
@LoisRForgeFlow what do you think of this solution?

base_tier_validation/models/tier_validation.py
```diff
@@ -274,6 +274,7 @@ def write(self, vals):
                 rec.review_ids
                 and rec._check_tier_state_transition(vals)
                 and not rec._check_allow_write_under_validation(vals)
+                and not rec._context.get("skip_validation_check")
```
LoisRForgeFlow, line 277: I'm still having a hard time understanding why is not enough the hook method `_check_allow_write_under_validation`.

Adding a context, will require you to do changes in other modules to send the context anyway. Why not extending _check_allow_write_under_validation hook? it is there for this exact purpose.

Also if you find something obvious that should be considered as a global exception (in some of the related issues messages with attachments are mentioned for instance) you could also propose a enhancement in the default logic here in the base module - with this enhancement proposals some steps to reproduce the issue in runbot will also help getting things approved faster.

Thanks for contributing!

LoisRForgeFlow (changes requested):

base_tier_validation/models/tier_validation.py
ilyasProgrammer: @LoisRForgeFlow Hi. _check_allow_write_under_validation works based on fields list (from _get_under_validation_exceptions).
But in some cases need to allow to write all fields. In this case you cant extend fields list in _get_under_validation_exceptions. What if we add same context check into _check_allow_write_under_validation ?

LoisRForgeFlow: @ilyasProgrammer _check_allow_write_under_validation can be extended in any specific model to add extra logic, pspeudo code:

```python

class Invoice()
    _name = "account.move"

    def _check_allow_write_under_validation(self, vals):
         res = super()._check_allow_write_under_validation(vals)
         if rec.partner_id.name = "My best friend":
             return True
         return res

```

So you can add any logic extending this method that adjust to your business case. If the problem is with something that is common can be even added here in the original method in the base module.

This being said, I don't want to be blocking, it is not a harmful change and if you really find this context useful, we can merge it as it. Maybe I just miss a specific example in which is really needed.

ilyasProgrammer: @LoisRForgeFlow _check_allow_write_under_validation can be helpfull. But then need to create compatibility modules.
In my case - delivery_auto_refresh does not depend on base_tier_validation. That means need to create glue module base_tier_validation_delivery_auto_refresh and inherit _check_allow_write_under_validation.
Similar way all other modules, having compatibility issues with base_tier_validation, should have glue modules for base_tier_validation.
Would it be correct solution ?

LoisRForgeFlow: Yeah, I think you have a good point in that, with the context this glue modules could be avoided, let's merge then!

LoisRForgeFlow (approved):

PR #616, opened 2023-03-16
[FIX][14.0] base_export_manager: manage

petrus-v:
while exporting data, (could be before installing this module), if user request for postgresql id, we can get some entries likes ,  in ir.exports.line.

We should be able to handle such existing lines


pedrobaeza:
This is harmless and it's good to be covered, so merging:

/ocabot merge patch

Please check if applicable in upper versions.

PR #618, opened 2023-03-20
[16.0][MIG] base_binary_url_import: Migration to 16.0

lehoangan:
Dependency on old `rfc6266-parser` removed, and replaced by a simple function relying on `werkzeug`


base_binary_url_import/wizard/base_binary_url_import.py
```diff
@@ -0,0 +1,343 @@
...
+        self.ensure_one()
+        target_id = self.target_record_identifier
+        if not target_id:
+            return
+        if target_id.isdigit():
+            return "db_id"
+        elif "." in target_id:
+            return "xml_id"
+        else:
+            raise UserError(_("Identifier %s is not an Integer or XMLID") % target_id)
```
StephaneMangin, line 247:
```suggestion
        raise UserError(_("Identifier %s is not an Integer or XMLID") % target_id)
```

StephaneMangin (approved):

base_binary_url_import/wizard/base_binary_url_import.py
lehoangan: I think we don't need to modify as your suggestion

StephaneMangin: I approved it anyway ;)

base_binary_url_import/tests/__init__.py
```diff
@@ -0,0 +1,4 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```
leemannd: In the OCA we don't consider usefull to add license in `__init__.py`

base_binary_url_import/tests/__init__.py
```diff
@@ -0,0 +1,4 @@
...
+
+# from . import tes_binary_url_import
```
leemannd: Why is the test deactivated?

base_binary_url_import/wizard/base_binary_url_import.py
```diff
@@ -256,17 +264,30 @@ def import_binary_from_url(self, request_session=None):
...
+        def get_filename(respone):
+            content_disposition = respone.headers.get("Content-Disposition")
+            if content_disposition:
+                param, options = werkzeug.http.parse_options_header(content_disposition)
+                if param == "attachment" and options.get("filename"):
+                    return options.get("filename")
+
+            path = urlparse(respone.url).path
+            name = path[path.rfind("/") + 1 :]
+            return name
```
grindtildeath: I don't understand why we would like to maintain such code when an external library is doing it for us :confused:

grindtildeath (commented):
Thanks for the mig. But what's the reason behind not keeping the test module and rewriting everything?

base_binary_url_import/wizard/base_binary_url_import.py
nilshamerlinck: Hello @grindtildeath

- `rfc6266-parser` depends on [`LEPL`](https://github.com/SWW13/python-rfc6266-parser/blob/master/setup.py#L15), which is not compatible with python 3.6+:

```
    import rfc6266_parser
  File "/home/nils/.local/share/virtualenvs/venv-odoo16/lib/python3.10/site-packages/rfc6266_parser.py", line 14, in <module>
    from lepl import (
  File "/home/nils/.local/share/virtualenvs/venv-odoo16/lib/python3.10/site-packages/lepl/__init__.py", line 113, in <module>
    from lepl.contrib.matchers import SmartSeparator2
  File "/home/nils/.local/share/virtualenvs/venv-odoo16/lib/python3.10/site-packages/lepl/contrib/matchers.py", line 41, in <module>
    from lepl.matchers.derived import Optional
  File "/home/nils/.local/share/virtualenvs/venv-odoo16/lib/python3.10/site-packages/lepl/matchers/derived.py", line 38, in <module>
    from lepl.matchers.combine import And, DepthFirst, BreadthFirst, \
  File "/home/nils/.local/share/virtualenvs/venv-odoo16/lib/python3.10/site-packages/lepl/matchers/combine.py", line 45, in <module>
    from lepl.matchers.core import Literal
  File "/home/nils/.local/share/virtualenvs/venv-odoo16/lib/python3.10/site-packages/lepl/matchers/core.py", line 45, in <module>
    from lepl.matchers.support import OperatorMatcher, coerce_, \
  File "/home/nils/.local/share/virtualenvs/venv-odoo16/lib/python3.10/site-packages/lepl/matchers/support.py", line 37, in <module>
    from lepl.core.config import ParserMixin
  File "/home/nils/.local/share/virtualenvs/venv-odoo16/lib/python3.10/site-packages/lepl/core/config.py", line 40, in <module>
    from lepl.stream.factory import DEFAULT_STREAM_FACTORY
  File "/home/nils/.local/share/virtualenvs/venv-odoo16/lib/python3.10/site-packages/lepl/stream/factory.py", line 31, in <module>
    from collections import Iterable
ImportError: cannot import name 'Iterable' from 'collections' (/home/nils/.pythonz/pythons/CPython-3.10.10/lib/python3.10/collections/__init__.py)
```

- `rfc6266-parser` was [forked](https://github.com/genme/python-rfc6266-parser) to make it compatible with python 3.6+ and published as `uc-rfc6266-parser`;
- BUT the way to make it compatible with python 3.6+ was to drop `LEPL` in favor of `werkzeug`, pinned to [1.0.1](https://github.com/genme/python-rfc6266-parser/blob/master/setup.py#L16), which makes it incompatible with odoo, that also needs a specific `werkzeug` [version](https://github.com/odoo/odoo/blob/16.0/requirements.txt#L58)
- a [PR](https://github.com/genme/python-rfc6266-parser/pull/2) to unpin `werkzeug` exists, but the format `git+https://github.com/genme/python-rfc6266-parser.git@4af2f324be019bdd6ca6543bc36d1d031ca609d5#egg=rfc6266-parser` is not compatible with odoo's [`external_dependencies` manifest key](https://github.com/odoo/odoo/blob/e0d5b92542f9272616c27cd2c9bf96d37e9a5532/odoo/addons/base/models/ir_module.py#L337);
- as an alternative, we could have published our own fork on pypi, BUT at that point we just decided to drop the dependency in favor of a simple function based on `werkzeug`
- NOW that we're having a second look, we just noticed that there is [another quite recent fork](https://github.com/JohnDoee/pyrfc6266) from the original `rfc6266` that took the option of using `pyparsing` instead of `LEPL`: it could do the job, we'll look into it

grindtildeath: @nilshamerlinck Thank you very much for the explanation. :+1: 

IMO we should better use a lib as long as it does the job, but if we cannot let's just keep the function in the module, but move it to a dedicated `rfc6266_parser.py` file outside of the models folder.

base_binary_url_import/tests/__init__.py
lehoangan: I will remove it. I don't have this test in the module

base_binary_url_import/wizard/base_binary_url_import.py
nilshamerlinck: `pyrfc6266` does the job, so custom function has been removed; test module has been included also

gurneyalex (approved):

PR #623, opened 2023-03-22
[14.0][ADD] web_tree_customized_field_list

ilyasProgrammer:


elvise:
@ilyasProgrammer can you rebase ?

ilyasProgrammer:
@elvise done

elvise:
@ilyasProgrammer please rebase :)

francesco-ooops (approved):
Functional ok!

Often a customer asks to add a field to tree view, and this requires dev time. This module allows functionals to be independent on adding fields to tree view.

customized_list_view/models/custom_list.py
```diff
@@ -0,0 +1,139 @@
...
+from odoo import _, api, exceptions, fields, models
+
+
+class CustomListView(models.Model):
+    _name = "custom.list.view"
+    _description = "Custom List View"
+
+    name = fields.Char(required=True)
+    model_id = fields.Many2one("ir.model", required=True, ondelete="cascade")
+    model_name = fields.Char(compute="_compute_model_name", store=True)
```
geomer198: For what this field have compute attribute? Could you make this field with related attribute?
```py
model_name = fields.Char(related="model_id.model")
```
I think this solution is easier.

customized_list_view/models/custom_list.py
```diff
@@ -0,0 +1,139 @@
...
+    list_view_id = fields.Many2one("ir.ui.view", required=True)
+    line_ids = fields.One2many("custom.list.view.line", "custom_list_view_id")
+    original_arch = fields.Text(readonly=True)
+
+    @api.constrains("model_id")
+    def _check_unique_model(self):
+        for rec in self:
+            same_model_recs = self.env["custom.list.view"].search(
```
geomer198: I think better use `search_count` method. Method `search` with `count=True` argument return clean data, without running `browse` method.

customized_list_view/models/custom_list.py
```diff
@@ -0,0 +1,139 @@
...
+
+    @api.depends("model_id")
+    def _compute_model_name(self):
+        for rec in self:
+            rec.model_name = rec.model_id.model
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        res = super().create(vals_list)
+        res.original_arch = res.list_view_id.arch
```
geomer198: Please rename `res` variable, because `create` method return record or record set. Better name will be `records` here.
I think is wrong, `res` variable is record set then you need make like this:
```py
for record in records:
    record.original_arch = record.list_view_id.arch
```

customized_list_view/models/custom_list.py
```diff
@@ -0,0 +1,139 @@
...
+        return res
+
+    def write(self, vals):
```
geomer198: For what this method here? Please remove unused code.

customized_list_view/models/custom_list.py
```diff
@@ -0,0 +1,139 @@
...
+        res = super().write(vals)
+        return res
+
+    def button_apply_changes(self):
+        self.ensure_one()
+        doc = etree.XML(self.original_arch)
+        for mod_line in self.line_ids:
+            target = mod_line.before.name if mod_line.before else mod_line.after.name
```
geomer198: This line is correct but will make better:
```py
mod_line.before and mod_line.before.name or mod_line_after.name
```

customized_list_view/models/custom_list.py
```diff
@@ -0,0 +1,139 @@
...
+        self.list_view_id.arch = new_arch
+
+    def button_roll_back(self):
+        self.ensure_one()
+        doc = etree.XML(self.original_arch)
+        new_arch = etree.tostring(doc, encoding="unicode").replace("\t", "")
+        self.list_view_id.arch = new_arch
+
+
+class CustomListViewLine(models.Model):
```
geomer198: I think better this class get out of separate file with name `customer_list_view_line.py`.

customized_list_view/models/custom_list.py
```diff
@@ -0,0 +1,139 @@
...
+    _description = "Custom List View Line"
+
+    custom_list_view_id = fields.Many2one("custom.list.view")
+    model_id = fields.Many2one(related="custom_list_view_id.model_id")
+    model_name = fields.Char(related="custom_list_view_id.model_name")
+    field_id = fields.Many2one("ir.model.fields", "New Field")
+    after = fields.Many2one("ir.model.fields", "Place After")
+    before = fields.Many2one("ir.model.fields", "Place Before")
+    optional = fields.Selection(
+        [("show", "show"), ("hide", "hide"), ("", "")], default=""
```
geomer198: Why do you added empty option?
If you remove `default` argument and empty option then by default in selection empty option is auto added.

customized_list_view/models/custom_list.py
```diff
@@ -0,0 +1,139 @@
...
+    )
+    label = fields.Char()
+    widget = fields.Selection(
+        [
+            ("many2many_tags", "many2many_tags"),
```
geomer198: I think use technical name for name of selection element  is not good idea.

customized_list_view/models/custom_list.py
```diff
@@ -0,0 +1,139 @@
...
+            field_ids = self.env["ir.model.fields"].search(
+                [
+                    ("model", "=", rec.custom_list_view_id.model_name),
+                    ("name", "in", field_names),
+                ]
+            )
+            rec.fields_domain = json.dumps([("id", "in", field_ids.ids)])
+
+
+class IrUiView(models.Model):
```
geomer198: I think better this class get out of separate file with name `ir_ui_view.py`.

customized_list_view/models/custom_list.py
```diff
@@ -0,0 +1,139 @@
...
+    _inherit = "ir.ui.view"
+
+    def name_get(self):
+        res = []
```
geomer198: It's correct but you can do in one line:
```py
return [(rec.id, rec.xml_id) for rec in self]
```

customized_list_view/tests/test_customized_list_view.py
```diff
@@ -0,0 +1,68 @@
...
+
+class TestCustomizedListView(common.SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super(TestCustomizedListView, cls).setUpClass()
+        cls.view_model = cls.env["ir.ui.view"].sudo()
+        cls.users_model = cls.env["ir.model"]._get("res.users")
+        cls.view_users_tree = cls.env.ref("base.view_users_tree")
+
+    def test_all_customized_list_view(self):
```
geomer198: Could you add docstring where described what you do in this test. 
Please add message for each assert case.

geomer198 (changes requested):
Could you please fix code by my comments.

ilyasProgrammer:
@geomer198 Thanks for the review. Updated according to your suggestions.

customized_list_view/models/custom_list.py
```diff
@@ -0,0 +1,120 @@
...
+                    ("id", "!=", rec.id),
+                ]
+            )
+            if same_model_recs:
+                raise exceptions.ValidationError(
+                    _(
+                        "Only one record per model is allowed. "
+                        "Please modify existing record instead."
+                    )
+                )
```
HekkiMelody: **suggestion (non-blocking)**: I think you could implement a simpler SQL `unique` constraint here.

**question**: Shouldn't the uniqueness constraint apply to `list_view_id` rather than `model_id`?

customized_list_view/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+clw1,clw1,model_custom_list_view,base.group_system,1,1,1,1
+clw2,clw2,model_custom_list_view,base.group_user,1,0,0,0
+clwl1,clwl1,model_custom_list_view_line,base.group_system,1,1,1,1
+clwl2,clwl2,model_custom_list_view_line,base.group_user,1,0,0,0
```
HekkiMelody: **question**: does `base.group_user` actually need access to this?

HekkiMelody (commented):
Code review, only a few minor remarks.

However, I should point out that the customizations do not survive an upgrade. In other words, when any module is upgraded, the views of that module are reverted back to what is written in the XML file, and it would be necessary to press "Apply Changes" again to make them appear again (Tested on runboat)

ilyasProgrammer:
> Code review, only a few minor remarks.
> 
> However, I should point out that the customizations do not survive an upgrade. In other words, when any module is upgraded, the views of that module are reverted back to what is written in the XML file, and it would be necessary to press "Apply Changes" again to make them appear again (Tested on runboat)

Thats right. Updated description and added roadmap.

customized_list_view/models/custom_list.py
ilyasProgrammer: Agree

customized_list_view/security/ir.model.access.csv
ilyasProgrammer: No. Removed.

customized_list_view/models/custom_list.py
```diff
@@ -0,0 +1,112 @@
...
+    line_ids = fields.One2many("custom.list.view.line", "custom_list_view_id")
+    original_arch = fields.Text(readonly=True)
+
+    _sql_constraints = [
+        (
+            "unique_model_list_view_rec",
+            "UNIQUE(model_id, list_view_id)",
+            "Only one record per model is allowed. "
+            "Please modify existing record instead.",
+        )
```
HekkiMelody: This is really "Only one record per view is allowed", and the uniqueness constraint could be applied to `list_view_id` only, IMO.

HekkiMelody (approved):
Approved with the above caveat that it may be very counterintuitive to use.

HekkiMelody (approved):

customized_list_view/models/ir_module.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2024 ooops404
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import models
+
+
+class Module(models.Model):
+    _inherit = "ir.module.module"
+
+    def _button_immediate_function(self, function):
```
HekkiMelody: **praise** this is an elegant way of making the changes persistent

**thought**: not all changes may be applied. at any given time. I'd say we could either add the `active` field to the `custom.list.view` model, or another boolean field to keep track of which changes need to be applied and which ones should not.

HekkiMelody (approved):

HekkiMelody:
Code review LGTM

customized_list_view/models/custom_list.py
geomer198: This comment is not resolved.

customized_list_view/models/ir_module.py
```diff
@@ -0,0 +1,18 @@
...
+        res = super(Module, self)._button_immediate_function(function)
+        views_mods = self.env["custom.list.view"].search([])
+        for vm in views_mods:
```
geomer198: This solution is not optimal.
Please change this implementation without 2 cycles. If you want remove all 'custom.list.view.line' records without `field_id` value, you can use search for this. I think it's faster then you solution.

customized_list_view/tests/test_customized_list_view.py
geomer198: This comment is not resolved.

geomer198 (changes requested):
Could you check my review?

customized_list_view/models/custom_list.py
geomer198: > I think better this class get out of separate file with name `customer_list_view_line.py`.

@ilyasProgrammer What do you think about it?

customized_list_view/models/custom_list.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class CustomListView(models.Model):
+    _name = "custom.list.view"
+    _description = "Custom List View"
+
+    active = fields.Boolean(default=True)
+    name = fields.Char(required=True)
+    model_id = fields.Many2one("ir.model", required=True, ondelete="cascade")
+    model_name = fields.Char(related="model_id.model", store=True)
```
geomer198: For what you use `store` here?

customized_list_view/tests/test_customized_list_view.py
```diff
@@ -0,0 +1,83 @@
...
```
geomer198: Could you add doc string for this test method, which describe what this method do?

geomer198 (changes requested):
Please, add little fix.

customized_list_view/tests/test_customized_list_view.py
ilyasProgrammer: The method is for tests. There is comment for each subtest and message for each assert. It is sufficient.

customized_list_view/models/custom_list.py
ilyasProgrammer: To avoid recompute.

geomer198 (approved):
LGTM!

francesco-ooops:
@pedrobaeza could you merge this one? thanks!

pedrobaeza:
Being a modification of the web client, it should be prefixed with `web_` IMO. A name can be `web_tree_customized_field_list`.

francesco-ooops:
@pedrobaeza should it be in this repo or in OCA/web in your opinion?

pedrobaeza:
Good question... Let's keep it here as is less crowded.

francesco-ooops (approved):
LGTM

@pedrobaeza shall we go?

PR #625, opened 2023-03-30
[13.0][ADD] server_action_merge

hbrunn:
This module allows users to merge records of arbitrary models.

Configuration
=============

To configure this module, you need to:

- Go to Settings/Technical/Server actions
- Create a server action of type `Merge` (field `Action To Do`)
- Click `Create Contextual Action` to make the action accessible for users

Usage
=====

To use this module, you need to:

- Go to a list view of a model for which you have created a merge server action
- Mark at least two records and start your merge action
- Select the record to merge into
- Click `Merge`

Technically, this is a UI for https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_merge_records.py#L916


server_action_merge/models/ir_actions_server.py
```diff
@@ -0,0 +1,34 @@
...
+
+class IrActionsServer(models.Model):
+    _inherit = "ir.actions.server"
+
+    state = fields.Selection(selection_add=[("merge", "Merge records")])
+    merge_method = fields.Selection(
+        [("orm", "ORM"), ("sql", "SQL")], default="orm", string="Merge method"
+    )
+    merge_handling = fields.Selection(
+        [("delete", "Delete"), ("deactivate", "Deactivate"), ("none", "Do nothing")],
```
legalsylvain, line 16: what will happen if the "none" option is selected?

hbrunn: well, nothing ;-)

legalsylvain: Does it make sense ? 

we could : 
- remove that option
- remove the default (default=False)
- by view, set merge_handling required, if state==merge
- by constrains, forbid state==merge and merge_handling==False

hbrunn: I just added a help text for this. I think 'delete' is a good default, that's how I'd expect a merge operation to work in general. But there might be cases where you'd like to keep the merged records, and if there's no 'active' field, doing nothing is the only option, right?

PR #630, opened 2023-04-12
[FIX] base_tier_validation: Use domain for filtering values

etobella:
Using active_test might arise lots of problems when executing functions. For example, I found the issue when validating Stock request orders, as it was trying to use archived routes. It was a problem, because I had a solution when coming from the widget and another one when using the standard menu.

The right approach is to show them using domain.

@rousseldenis @kittiu @LoisRForgeFlow

### DIFF
--- a/base_tier_validation/static/src/js/systray.js
+++ b/base_tier_validation/static/src/js/systray.js
@@ -134,7 +134,7 @@ odoo.define("tier_validation.systray", function (require) {
                 $(event.currentTarget).data(),
                 $(event.target).data()
             );
-            var context = {active_test: false};
+            var context = {};
             this.do_action({
                 type: "ir.actions.act_window",
                 name: data.model_name,
--- a/base_tier_validation/static/src/js/systray.js
+++ b/base_tier_validation/static/src/js/systray.js
@@ -134,7 +134,7 @@ odoo.define("tier_validation.systray", function (require) {
                 $(event.currentTarget).data(),
                 $(event.target).data()
             );
-            var context = {from_review_systray: true, active_test: false};
+            var context = {active_test: false};
             this.do_action({
                 type: "ir.actions.act_window",
                 name: data.model_name,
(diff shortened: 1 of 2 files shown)


base_tier_validation/static/src/js/systray.js
```diff
@@ -144,7 +145,10 @@ odoo.define("tier_validation.systray", function (require) {
                     [false, "form"],
                 ],
                 search_view_id: [false],
-                domain: [["can_review", "=", true]],
+                domain: [
+                    ["can_review", "=", true],
+                    ["active", "in", [true, false]],
```
rousseldenis, line 149: Mmmmh, indeed that one seems better !

rousseldenis: I'll try to test that soon

base_tier_validation/static/src/js/systray.js
```diff
@@ -134,7 +134,8 @@ odoo.define("tier_validation.systray", function (require) {
                 $(event.currentTarget).data(),
                 $(event.target).data()
             );
-            var context = {from_review_systray: true, active_test: false};
+            var context = {};
+            console.log("HI!!!");
```
LoisRForgeFlow: :wave:   I think this can be removed

LoisRForgeFlow (approved):
LGTM

bosd (approved):
LGTM!

bosd:
@etobella  Can you forward port this?

hbrunn:
@tobella @LoisRForgeFlow this breaks tier validation for any model that doesn't have the active flag. I suggest to rework this in order to look at the model's fields and and only add the domain if the active field exists.

LoisRForgeFlow:
@hbrunn Indeed, I just merged the fix proposed last friday https://github.com/OCA/server-ux/pull/642. You can find the fwport to v15 and 16 linked to that pr which aren't merged yet (needing review)

PR #638, opened 2023-04-19
[15.0][MIG] base_tier_validation_waiting

bosd:
Depends on: 
#732 Do not sent double messages on the first validation sequence


base_tier_validation_waiting/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+This module extends base_tier_validation to add an initial **"waiting"** status to tier reviews.
+It also auto moves next review to pending status when needed and handles notification if set.
```
simahawk, line 2: can you drop empty readme files pls?

base_tier_validation_waiting/tests/test_tier_validation.py
```diff
@@ -0,0 +1,101 @@
...
+
+from odoo.tests.common import SavepointCase, tagged
+
+
+@tagged("post_install", "-at_install")
+class TierTierValidation(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super(TierTierValidation, cls).setUpClass()
+
```
simahawk: can you set tracking_disable + no_reset_password context keys pls?

simahawk (commented):
LG minor comments

bosd:
@cwjoalder Can you please review?

bosd:
@simahawk Can you review the changes? :pray:

base_tier_validation_waiting/models/tier_validation.py
```diff
@@ -0,0 +1,50 @@
...
+    def _validate_tier(self, tiers=False):
+        """extend base tier validation method to allow bypass of waiting
+        sequences as pending state is hardcoded"""
+        self.ensure_one()
+        res = super(TierValidation, self)._validate_tier(tiers)
+        tier_reviews = tiers or self.review_ids
+        waiting_reviews = tier_reviews.filtered(
+            lambda r: r.status == "waiting"
+            and (self.env.user in r.reviewer_ids)
+            and r.approve_sequence_bypass
```
michaelslade54, line 33: It appears this migration relies on PR #656 or am I missing something?

bosd: Great Spot!! Seems it was missing in the migration.

bosd: That pr is merged, so we can continue here.

bosd:
Force pushed to recreate runboat

bosd:
@Dranyel-Bosd Can you please review?

Dranyel-Bosd (approved):
looks so good to me :+1: :sparkler:

base_tier_validation_waiting/tests/test_tier_validation.py
simahawk: I meant class wise ;)
```suggestion
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```

base_tier_validation_waiting/tests/test_tier_validation.py
```diff
@@ -0,0 +1,103 @@
...
+        cls.loader = FakeModelLoader(cls.env, cls.__module__)
+        cls.loader.backup_registry()
+        from odoo.addons.base_tier_validation.tests.tier_validation_tester import (
+            TierValidationTester,
+        )
+
+        cls.loader.update_registry((TierValidationTester,))
+        cls.test_model = cls.env[TierValidationTester._name].with_context(
+            tracking_disable=True, no_reset_password=True
```
simahawk: see above

base_tier_validation_waiting/models/tier_review.py
```diff
@@ -0,0 +1,45 @@
...
+
+    status = fields.Selection(
+        selection_add=[("waiting", "Waiting")],
+        default="waiting",
+        ondelete={"waiting": "set default"},
+    )
+
+    def _notify_pending_status(self, review_ids):
+        """Method to call and reuse abstract notification method"""
+        resource = self.env[self.model].browse(self.res_id)
```
simahawk, line 29: this is not tested, any chance?

base_tier_validation_waiting/models/tier_review.py
```diff
@@ -0,0 +1,45 @@
...
+        )
+        if reviews:
+            # get minimum sequence of all to prevent jumps
+            next_seq = min(reviews.mapped("sequence"))
+            for record in reviews:
+                # if not in waiting, nothing to do
+                if record.status != "waiting":
+                    continue
+                # if approve by sequence, check sequence has been reached
+                if record.approve_sequence:
```
simahawk, line 46: I'm not sure I get this IF block: in any case you set "pending" -> can't you have 1 single check?

base_tier_validation_waiting/models/tier_review.py
```diff
@@ -0,0 +1,45 @@
...
+                    if record.sequence == next_seq:
+                        record.status = "pending"
+                # else, move forward any waiting review
+                else:
+                    record.status = "pending"
+                # notify if state has changed
+                if record.status == "pending":
```
simahawk, line 53: IMO this can should be done on inverse method for `status` field

base_tier_validation_waiting/models/tier_review.py
```diff
@@ -0,0 +1,45 @@
...
+                    if record.definition_id.notify_on_pending:
+                        record._notify_pending_status(record)
+        return super(TierReview, self)._compute_can_review()
```
simahawk, line 56: can't we call super w/ a subset of the records?

base_tier_validation_waiting/models/tier_validation.py
```diff
@@ -0,0 +1,50 @@
...
+        """method to notify when reaching pending"""
+        subscribe = "message_subscribe"
+        post = "message_post"
+        if hasattr(self, post) and hasattr(self, subscribe):
+            for rec in self:
+                users_to_notify = tier_reviews.filtered(
+                    lambda r: r.definition_id.notify_on_pending and r.res_id == rec.id
+                ).mapped("reviewer_ids")
+                # Subscribe reviewers and notify
+                getattr(rec.sudo(), subscribe)(
```
simahawk: why calling sudo every time if you can do `for rec in self.sudo()`?

simahawk: Also, is sudo really needed?

base_tier_validation_waiting/models/tier_validation.py
```diff
@@ -0,0 +1,50 @@
...
+                    partner_ids=users_to_notify.mapped("partner_id").ids
+                )
+                getattr(rec.sudo(), post)(
+                    subtype_xmlid=self._get_requested_notification_subtype(),
+                    body=rec._notify_requested_review_body(),
+                )
+
+    def _validate_tier(self, tiers=False):
+        """extend base tier validation method to allow bypass of waiting
+        sequences as pending state is hardcoded"""
```
simahawk: I'd drop this docstring and move the custom logic into `_validate_tier_waiting_reviews`

simahawk (changes requested):

base_tier_validation_waiting/models/tier_validation.py
bosd: Apparantly it is: https://github.com/OCA/server-ux/pull/359#issue-936237493

base_tier_validation_waiting/models/tier_review.py
bosd: I've added tests on the notifications

bosd:
@simahawk Thanks for your review!

I'm working my way towards more coverage and green tests :grin: 
Adding coverage by testing the notifications.
But it's breaking/affecting tests of other modules.
Added the fixes for those in this PR. Is that allowed/correct OCA procedure?

I've watched your talks about the unittests. 
Still I'm struggling with getting how to get the tier_validation_server_action fixed.

This module is altering the behaviour by adding the state waiting, which results in breaking the tests of server action.
When this module is installed, an alternative version of tests should be run by that module. Is that correct?
Can you explain how to do that?  :blush: :pray: 

```

===================================================================== 
2023-10-23 07:04:27,754 1037 ERROR odoo odoo.addons.base_tier_validation_server_action.tests.test_tier_validation: FAIL: TierTierValidation.test_1_auto_validation
  File "/__w/server-ux/server-ux/base_tier_validation_server_action/tests/test_tier_validation.py", line 100, in test_1_auto_validation
    self.assertEqual(
AssertionError: Lists differ: ['waiting', 'waiting', 'waiting'] != ['pending', 'pending', 'pending']

First differing element 0:
'waiting'
'pending'

- ['waiting', 'waiting', 'waiting']
+ ['pending', 'pending', 'pending']
 
 ```

**Comment by DavidJForgeFlow** [comment]
```
from odoo import api, fields, models
from odoo.tools import config


    _inherit = "tier.review"


    def _default_status(self):
        if not config["test_enable"] or config["test_enable"] and self.env.context.get("_testing_base_tier_validationwaiting"):
            self.status = "waiting"
        else:
            self.status = "pending"
        
    status = fields.Selection(
        selection_add=[("waiting", "Waiting")],
        default="_default_status",
        ondelete={"waiting": "set default"},
    )
```
Maybe this helps to work with the other module

**Review by Dranyel-Bosd — APPROVED** [review]
LGTM :+1:

**Comment by bosd** [comment]
Dropped the dependency on 732. Now we finally have green tests  :heavy_check_mark: 
:sparkles: :tada:

File: base_tier_validation_waiting/models/tier_review.py [hunk 11]
```diff
@@ -2,14 +2,25 @@
...
 from odoo import api, fields, models
+from odoo.tools import config
 
 
 class TierReview(models.Model):
     _inherit = "tier.review"
 
+    def _default_status(self):
+        if (
+            not config["test_enable"]
```
simahawk, line 13: do we really need this specific check for tests?

base_tier_validation_waiting/models/tier_review.py
```diff
@@ -2,14 +2,25 @@
...
+            and self.env.context.get("testing_base_tier_validation_waiting")
+        ):
+            return "waiting"
+        else:
+            return "pending"
+
     status = fields.Selection(
         selection_add=[("waiting", "Waiting")],
-        default="waiting",
+        default=_default_status,
```
simahawk, line 23: use lambda to not reference the method object directly

base_tier_validation_waiting/models/tier_review.py
bosd: Yes, without this commit, it would mean that we need to patch a lot of depending modules, to fix their tests.

Another solution I've evaluated is using seperate tests. That might solve the problem with the CI tests here on github.
But it would simply transfer the problem to the persons running an custom CI like odoo.sh.

The method here is not really clean. But it's the least dirty solution.
(Discussed this test issue with a couple of oca contributors at the OXP)
Maybe you have another insight to solve this problem?

PR #640, opened 2023-04-19
[16.0][IMP] base_tier_validation: compute validation flag dynamically

hbrunn:
with this the validation button shows immediately on a form and not only after safe


astirpe:
@hbrunn thank you for this PR!
I'm testing it locally in combination with account_move_tier_validation. When I create a new invoice, before it's saved, the Confirm button is displayed and the Request Validation button is not. Not sure if I'm doing something wrong.

hbrunn:
is there still a problem with updating the dependencies in all registries? if you created your tier definition in the same session, can you restart the server and see if it behaves correctly then?

astirpe:
@hbrunn Restarting the server doesn't change anything:

- method `_compute_need_validation()` is called only after saving
- method `_compute_need_validation_dependencies()` is not called at all

hbrunn:
I think I see the issue: Probably your domain only includes `move_type` (or otherwise exclusively fields that are marked as readonly on the form)? Does it work for you if you add something like `('name', 'like', '%')`?

If that works, get_view should add the fields used in the domain as invisible + not readonly, this should trigger the computation.

astirpe:
I tried with `('name', 'like', '%')` but didn't work.

PR #641, opened 2023-04-20
[15.0][FIX] base_tier_validation: edit domain can_review

ps-tubtim:
I found the issue when validating the model without an active field such as Expense. So, I edit `search_can_review` function and I change the domain in `_onReviewFilterClick` to the old code.

NOTE: I don't know how to check if the field exists in .js file (like .py file). You can suggest it to me.

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -81,19 +81,15 @@ class TierValidation(models.AbstractModel):
 
     @api.model
     def _search_can_review(self, operator, value):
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow:
@ps-tubtim This problem has been fixed with https://github.com/OCA/server-ux/pull/642 in v14.

LoisRForgeFlow:
I think we can close this one in favor of the fwport to 15 - https://github.com/OCA/server-ux/pull/646

PR #642, opened 2023-04-20
[14.0][FIX] base_tier_validation: Detect if model has active field

etobella:
@ps-tubtim I think this fixes your problem. If we port the changes to 15 all should be fixed now (#641)

Tested with product_tier_validation, purchase_tier_validation and stock_request_tier_validation

@LoisRForgeFlow

### DIFF
--- a/base_tier_validation/static/src/js/systray.js
+++ b/base_tier_validation/static/src/js/systray.js
@@ -135,6 +135,11 @@ odoo.define("tier_validation.systray", function (require) {
                 $(event.target).data()
             );
             var context = {};
+            var domain = [["can_review", "=", true]];
+            console.log(data);
+            if (data.active_field) {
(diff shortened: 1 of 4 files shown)


base_tier_validation/static/src/js/systray.js
```diff
@@ -135,6 +135,12 @@ odoo.define("tier_validation.systray", function (require) {
                 $(event.target).data()
             );
             var context = {};
+            var domain = [["can_review", "=", true]];
+            console.log(data);
+            if (data.active_field) {
+                domain.push(["active", "in", [true, false]]);
+            }
+            console.log(domain);
```
LoisRForgeFlow: Can you remove the consele.log?

LoisRForgeFlow (approved):
LGTM

sanchonuria (approved):

PR #647, opened 2023-04-24
[16.0][FIX] base_tier_validation: Detect if model has active field

LoisRForgeFlow:
Forward port of #642

Adapted to owl implementation in v16.


hbrunn (approved):
thanks!

etobella (approved):

Issue #474, opened 2022-03-18
[14.0] base_export_manager disable export don't work


ygol:
unable to re-open #57 so copying it to a new issue

## Module
base_export_manager

## Describe the bug
disable export don't work

## To Reproduce
**Affected versions**: 14.0

## Steps to reproduce the behavior:
1.Activate the developer mode
2.Go to Settings > Users > Groups to select a user group
3.Edit the group and go to the Access Rights tab
4.Uncheck the "Export Access" box on the object of your choice and save
5.Connect as a user of this group
6.Go to the list view of the object you disabled the export but The "Export" is still there.

**Expected behavior**
no "Export" option shown

nicomacr:
@ygol I fix this problem here https://github.com/OCA/server-ux/pull/478 over v15, if you want, you can test it there in runbot build.

ygol:
hi @nicomacr thanks for your reply.
Here's what I tried:
1. I changed the configuration for the lead/opportunity accesses:
![image](https://user-images.githubusercontent.com/2945174/162214929-e5bd9bd2-e557-440b-a968-f535657cd56c.png)

2. I still can access the export option:
![image](https://user-images.githubusercontent.com/2945174/162215013-b7a899d2-d5ba-4c73-90f1-abfbd99f622f.png)

Did I miss something?
Best.

nicomacr:
> hi @nicomacr thanks for your reply. Here's what I tried:
> 
> 1. I changed the configuration for the lead/opportunity accesses:
>    ![image](https://user-images.githubusercontent.com/2945174/162214929-e5bd9bd2-e557-440b-a968-f535657cd56c.png)
> 2. I still can access the export option:
>    ![image](https://user-images.githubusercontent.com/2945174/162215013-b7a899d2-d5ba-4c73-90f1-abfbd99f622f.png)
> 
> Did I miss something? Best.

@ygol yes, the user in  this case "Admin" have the General settings "Manager" because of that you still see the option. Try with different user or change that right.

ygol:
hmm sorry if I don't understand. I removed the export flag for anyone regarding this model. 
What/where is the general settings "Manager"?

![image](https://user-images.githubusercontent.com/2945174/162223583-3c4e3627-dc06-4483-8f86-081d14838f0f.png)

nicomacr:
@ygol in the user settings, change for "Access Rights" instead of "Settings"
[...]

PR #649, opened 2023-04-25
[16.0] [MIG] base_export_manager: Migration to 16.0

rjpathan:


kikopeiro:
Here's what I tried:

- I changed the configuration (Accesses Rights) for the res.partner.category  (all accesses)
- I still can access the export option over this model
- Tested with not admin user

Did I miss something? V15 is working fine

What is reproduced is similar to what was discussed in this issue https://github.com/OCA/server-ux/issues/474.

rjpathan:
@kikopeiro sorry for late reply.

Export action button is disabled with a javascript function https://github.com/rjpathan/server-ux/blob/16.0-mig-base_export_manager/base_export_manager/static/src/js/base_export_manager.js
It seems this is not working in v16.

Sorry, I am not sure how to fix this as I don't have proficiency with the javascript.

hugosantosred:
@rjpathan I've opened a [PR](https://github.com/rjpathan/server-ux/pull/1) in your branch with the fix. In v16 the ListController is defined in OWL, so is needed to patch that controller in order to set the isExportEnable property

rjpathan:
@hugosantosred Thanks for the fix.

base_export_manager/models/ir_exports_line.py
```diff
@@ -142,8 +142,8 @@ def _inverse_name(self):
             if any(parts):
                 # invalidate_cache -> in order to get actual value of field 'label'
                 # in function '_check_name'
-                self.invalidate_cache(ids=one.ids)
-                one._check_name()
+                self.env.flush_all()
+                self.env.invalidate_all()
```
stefan-tecnativa, line 146: Passing `flush=True` parameter to invalidate_all() wouldn't be the same as both lines that you wrote?
https://github.com/odoo/odoo/blob/16.0/odoo/api.py#L714-L723

stefan-tecnativa:
You're pre-commit problem is related with the list_controller.js file. Two possible solutions could be:
1- Convert your code to ESM module (ECMAScript Modules)
2- Use requires to import objects from other files: https://github.com/OCA/OCB/blob/15.0/addons/account/static/src/js/account_payment_field.js#L5

Good Luck!

kikopeiro (approved):
:+1:  LGTM (without code review)

base_export_manager/models/ir_exports_line.py
```diff
@@ -142,8 +142,7 @@ def _inverse_name(self):
             if any(parts):
                 # invalidate_cache -> in order to get actual value of field 'label'
                 # in function '_check_name'
-                self.invalidate_cache(ids=one.ids)
-                one._check_name()
+                self.env.invalidate_all()
```
stefan-tecnativa, line 145: self.browse(one.ids).invalidate_recordset()
one._check_name()
```

stefan-tecnativa (changes requested):

pedrobaeza:
Superseded by #702

PR #657, opened 2023-05-16
[14.0][FIX] base_custom_filter: Prevent all records from being displayed in favorites

victoralmau:
Prevent all records from being displayed in favorites

Please @pedrobaeza and @pilarvargas-tecnativa  can you review it?

@Tecnativa TT43350

### DIFF
--- a/base_custom_filter/models/ir_filters.py
+++ b/base_custom_filter/models/ir_filters.py
@@ -2,7 +2,8 @@
 # Copyright 2021 ForgeFlow S.L. (https://www.forgeflow.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import fields, models
+from odoo import api, fields, models
+from odoo.osv import expression
 
 
 class IrFilters(models.Model):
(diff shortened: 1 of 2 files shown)


base_custom_filter/models/ir_filters.py
```diff
@@ -29,3 +30,19 @@ def _selection_type(self):
         ondelete="cascade",
     )
     group_id = fields.Many2one(comodel_name="ir.filters.group", string="Filter Group")
+
+    @api.model
+    def get_filters(self, model, action_id=None):
+        """We need to inject a context to obtain only the records of favorite type."""
+        _self = self.with_context(filter_type="favorite")
+        return super(IrFilters, _self).get_filters(model, action_id=action_id)
```
pedrobaeza: Remember that there's no need of this double variable as we learnt:

```suggestion
        self = self.with_context(filter_type="favorite")
        return super().get_filters(model, action_id=action_id)
```

pedrobaeza (approved):

pilarvargas-tecnativa (approved):
LGTM :+1:

PR #665, opened 2023-06-14
[15.0][FIX] base_import_security_group: fix test test_error

pilarvargas-tecnativa:
When the web_responsive module is installed, the tests fail when looking for "Settings" in the menu, because you have to access the menu first to get that option. Adding the extra step to display the menu is the right option so that the tests don't fail whether the web_responsive module is installed or not.

Example of the menu without web_responsive:

image

Example of the menu with web_responsive:

image

Error:

image

cc @Tecnativa TT42822

@chienandalu @CarlosRoca13 please review


CarlosRoca13 (approved):

chienandalu (approved):
:+1: great :)

PR #671, opened 2023-06-20
[16.0][mig] mass_operation_abstract

aktiv-heli-kantawala:


legalsylvain:
Hi @heliaktiv  what is the objective of the migration of your module ? 
AFAIK, this module is obsolete, and should not be ported.
thanks.

PR #672, opened 2023-06-20
[16.0][IMP] base_revision: create method in batch

CRogos:


auroraeosrose (approved):
This looks good to me, passes tests and no longer throws warnings

CRogos:
@pedrobaeza could you have a look in this, too.

base_revision/__manifest__.py
```diff
@@ -7,7 +7,7 @@
 {
     "name": "Base Revision (abstract)",
     "summary": "Keep track of revised document",
-    "version": "16.0.1.0.0",
+    "version": "16.0.1.0.1",
```
pedrobaeza: No need to change manually the version number.

base_revision/models/base_revision.py
```diff
@@ -96,13 +96,14 @@ def copy_revision_with_context(self):
...
-            name_field = self._context.get("revision_name_field", "name")
-            rec.write({"unrevisioned_name": rec[name_field]})
-        return rec
+    @api.model_create_multi
+    def create(self, vals_list):
+        records = super().create(vals_list)
+        name_field = self._context.get("revision_name_field", "name")
+        for rec, values in zip(records, vals_list):
+            if "unrevisioned_name" not in values:
+                rec.write({"unrevisioned_name": rec[name_field]})
```
pedrobaeza: As you are here, why not doing this before calling super for not having 2 DB operations per record.

base_revision/tests/test_base_revision.py
```diff
@@ -27,7 +27,7 @@ def tearDownClass(cls):
         return super(TestBaseRevision, cls).tearDownClass()
 
     def _create_tester(self):
-        return self.revision_model.create({"name": "TEST0001"})
+        return self.revision_model.create([{"name": "TEST0001"}])
```
pedrobaeza: No need of this change, having already a specific test.

base_revision/models/base_revision.py
CRogos: Can we be sure that the "name" value is set in the vals_list? Also in case of e.g. invoices or sales orders, where the name is generated with a continues number?

CRogos: @pedrobaeza I've debugged it with sale order and the name attribute is already set. I've changed it as you suggested. ready for review/merge. 

I also like to mention that I really apricate your input during the last days and years!

pedrobaeza: You're welcome, and I also appreciate your will to improve existing code and to receive positively the comments.

pedrobaeza (approved):
/ocabot merge patch

CRogos:
Unfortunately the new unit test breaks testing in sale_order_revision.

Please have a look on test improvements:
https://github.com/OCA/server-ux/pull/698
https://github.com/OCA/sale-workflow/pull/2611

PR #676, opened 2023-06-26
[13.0][ADD]base_archive_date & web_archive_date

GuillemCForgeFlow:
Adds an archived timestamp and user archiving whenever a record is archived.

Two modules:

- Base module to set the archived date and user archiving the record.
- Glue module between first module and web to display the values on the _View Metadata_ section.

cc @ForgeFlow

### DIFF
--- a/base_archive_date/__init__.py
+++ b/base_archive_date/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/base_archive_date/__manifest__.py
+++ b/base_archive_date/__manifest__.py
@@ -1,0 +1,17 @@
+# Copyright 2023 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Base Archive Date",
+    "summary": """
+        Adds an archive timestamp and user doing the archiving to all models.
+    """,
+    "version": "13.0.1.0.0",
+    "category": "Usability",
+    "license": "AGPL-3",
+    "website": "https://github.com/OCA/server-ux",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "depends": ["base"],
+    "maintainers": ["GuillemCForgeFlow"],
+    "installable": True,
+    "application": False,
+}
--- a/base_archive_date/models/__init__.py
+++ b/base_archive_date/models/__init__.py
@@ -1,0 +1,1 @@
+from . import base
--- a/base_archive_date/models/base.py
+++ b/base_archive_date/models/base.py
@@ -1,0 +1,74 @@
+# Copyright 2023 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import api, fields, models
+
+LOG_ARCHIVE_DATE = "archive_date"
+LOG_ARCHIVE_UID = "archive_uid"
+
+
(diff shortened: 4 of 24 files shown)


legalsylvain (commented):
Hi @GuillemCForgeFlow. Thanks a lot for your contribution. looks very interesting feature. 

A general question regarding the name. Why do you use "timestamp". AFAIK, it's the technical name of the field in the database. however, all OCA module use the technical name of the fields in odoo. like : 
- char : https://github.com/OCA/web/tree/14.0/web_widget_char_size
- text : https://github.com/OCA/web/tree/14.0/web_widget_text_markdown
- many2one : https://github.com/OCA/web/tree/14.0/web_tree_many2one_clickable

So I was expecting something like ``base_archive_datetime``. (or ``base_archive_date`` if you want to mention the name of the field).

GuillemCForgeFlow:
Hi @legalsylvain , thank you for the feedback, I really appreciate it.

Actually, I agree with you, I will rename to `base_archive_date`, which is also more convenient to understand the meaning of the module. I had doubts when naming at first.

base_archive_timestamp/models/base.py
```diff
@@ -0,0 +1,56 @@
...
+        now = fields.Datetime.context_timestamp(self, fields.Datetime.now())
+        return now.replace(tzinfo=None)
+
+    @api.model
+    def _add_magic_fields(self):
+        result = super(Base, self)._add_magic_fields()
+        if self._log_access and "active" in self._fields:
+            self._add_field(
+                LOG_ARCHIVE_DATE,
+                fields.Datetime(string="Archived on", automatic=True, readonly=True),
```
legalsylvain:
```suggestion
                fields.Datetime(string="Last Archived on", automatic=True, readonly=True),
```

legalsylvain (commented):
And Functional review.

Work great ! 
Filter is possible :+1: 

On the metadata view : 
image

- The format of the date of "archive field" is not the same as the other date fields. not a blocking point, can be added in the roadmap.rst file.

- I wonder if the archive_uid field should be added also ? naive question not a clear point of view).

GuillemCForgeFlow:
Hi @legalsylvain thank you again for the feedback.
I've applied you suggestions, now I log the user doing the archiving as well.

For the format of the date, I've added in the roadmap for now.

Perhaps you know how we could solve it. I was able to check that the formating is done in javascript (https://github.com/odoo/odoo/blob/13.0/addons/web/static/src/js/tools/debug_manager_backend.js#L248). I've tried using the `t-options="{'widget': 'date', 'format': ...}"` but didn't work out.

JordiMForgeFlow (approved):
Very nice module! 👍🏼

legalsylvain:
> Perhaps you know how we could solve it.

not very expert in JS, and it will change a lot with OWL, I guess in recent version.

**Side question :** 
what about to rename the module base_archive_date_metadata to web_archive_date to comply with OCA conventions regarding glue modules. Just a suggestion.

GuillemCForgeFlow:
@legalsylvain Just changed it! 👍🏽

legalsylvain (approved):

JordiMForgeFlow (approved):
Functional review 👍🏼

JaumeBforgeFlow (approved):
LGTM!

AaronHForgeFlow (approved):
Nice feature!

mariadforgeflow (approved):
This module is very interesting and useful, thanks a lot for your contribution @GuillemCForgeFlow!!  :1st_place_medal:

PR #678, opened 2023-06-26
[14.0][WIP] base_tier_validation: Change FakeModelLoader to setup_test_model()

victoralmau:
Change `FakeModelLoader` to `setup_test_model()` (similar to tests of `account_move_tier_validation`: https://github.com/OCA/account-invoicing/blob/14.0/account_move_tier_validation/tests/test_tier_validation.py#L17)

@Tecnativa TT43351

### DIFF
--- a/base_tier_validation/tests/__init__.py
+++ b/base_tier_validation/tests/__init__.py
@@ -1,4 +1,3 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from . import common
 from . import test_tier_validation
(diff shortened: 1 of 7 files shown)


LoisRForgeFlow:
@victoralmau Could you provide details on the motivation behind this change?

victoralmau:
This change is not ready yet, but it was intended to solve the error that happened when `base_changeset_tier_validation` (https://github.com/OCA/server-tools/pull/2623) was installed and then `base_tier_validation` tests were launched.

With the current approach the error I indicate does not appear but now there is another side effect that I would have to try to solve.

Can you think of another possible solution (currently it is the same approach in the `account_move_tier_validation` tests)?

LoisRForgeFlow:
@victoralmau thanks for the insights.

Unfortunately, I don't have a better idea, at some point using FakeModelLoader was good because it covered exactly the testing case we had. With the time we also faced some issues when doing tests a bit differently to the specific cases "officially" supported by odoo_test_helper, without digging in the details of your issues, in an ideal world your fixes would be in the odoo_test_helper but I assume is not that easy...

victoralmau:
This error is quite complex to resolve and affects only the tests with a "false" error.

PR #679, opened 2023-06-28
[15.0][IMP] announcement: add general_anouncement option to announcement

pilarvargas-tecnativa:
A new field "is_general_announcement" is added which when checked sends the announcement to users belonging to the "Internal User" group.

cc @Tecnativa TT42581

@chienandalu @victoralmau please review :)

### DIFF
--- a/announcement/models/announcement.py
+++ b/announcement/models/announcement.py
@@ -20,6 +20,7 @@ class Announcement(models.Model):
     sequence = fields.Integer()
     name = fields.Char(string="Title", required=True)
     content = fields.Html()
+    is_general_announcement = fields.Boolean("General Announcement")
     announcement_type = fields.Selection(
         selection=[
             ("specific_users", "Specific users"),
@@ -33,7 +34,12 @@ class Announcement(models.Model):
         domain=[("share", "=", False)],
         inverse="_inverse_specific_user_ids",
     )
-    user_group_ids = fields.Many2many(comodel_name="res.groups")
+    user_group_ids = fields.Many2many(
+        comodel_name="res.groups",
+        compute="_compute_user_group_ids",
+        store=True,
+        readonly=False,
+    )
     allowed_user_ids = fields.Many2many(
         comodel_name="res.users",
         relation="announcement_res_users_allowed_rel",
@@ -83,6 +89,16 @@ class Announcement(models.Model):
         for announcement in self - specific_user_announcements:
             announcement.allowed_user_ids = announcement.user_group_ids.users
             announcement.allowed_users_count = len(announcement.user_group_ids.users)
+
+    @api.depends("is_general_announcement")
+    def _compute_user_group_ids(self):
+        for announcement in self:
+            if announcement.is_general_announcement:
+                announcement.announcement_type = "user_group"
+                announcement.user_group_ids = self.env.ref("base.group_user")
+            else:
+                announcement.specific_user_ids = False
+                announcement.user_group_ids = False
 
     @api.depends("announcement_log_ids")
     def _compute_read_announcement_count(self):
--- a/announcement/views/announcement_views.xml
+++ b/announcement/views/announcement_views.xml
@@ -67,6 +67,9 @@
                         </h1>
                     </div>
                     <group>
+                        <field name="is_general_announcement" />
+                    </group>
+                    <group>
                         <field name="active" invisible="1" />
                         <group name="announcement_validity">
                             <field name="notification_date" string="Announce at" />
@@ -115,7 +118,10 @@
                 </button>
             </xpath>
             <group name="announcement_validity" position="before">
-                <group name="announcement_scope">
+                <group
+                    name="announcement_scope"
+                    attrs="{'invisible': [('is_general_announcement', '=', True)]}"
+                >
                     <field name="announcement_type" />
                     <field
                         name="specific_user_ids"


announcement/models/announcement.py
```diff
@@ -70,10 +71,17 @@ def _inverse_specific_user_ids(self):
...
 
-    @api.depends("specific_user_ids", "user_group_ids")
+    @api.depends("specific_user_ids", "user_group_ids", "is_general_announcement")
     def _compute_allowed_user_ids(self):
         self.allowed_user_ids = False
         self.allowed_users_count = False
+        # If the option "General announcement" is checked, we assign the group
+        # "Internal users" to the user_group_ids field.
+        if self.filtered(lambda x: x.is_general_announcement):
+            for announcement in self:
```
chienandalu: Here you're assuming that `is_general_announcement` will be always truthy and the same for every record.

chienandalu (changes requested):

announcement/models/announcement.py
pilarvargas-tecnativa: Sorry, I don't know what I was thinking... corrected, thank you very much!

announcement/models/announcement.py
```diff
@@ -70,10 +71,17 @@ def _inverse_specific_user_ids(self):
...
     def _compute_allowed_user_ids(self):
         self.allowed_user_ids = False
         self.allowed_users_count = False
+        # If the option "General announcement" is checked, we assign the group
+        # "Internal users" to the user_group_ids field.
+        general_announcement = self.filtered(lambda x: x.is_general_announcement)
+        for announcement in general_announcement:
+            announcement.announcement_type = "user_group"
+            group_user = self.env.ref("base.group_user")
+            announcement.user_group_ids = [(6, 0, [group_user.id])]
```
chienandalu: Do this UX feature need to be inside the core logic? Wouldn't it be enough with a simple onchange that writes such group into the field on edit time?

pilarvargas-tecnativa: Thank you for the suggestion, this makes it even clearer.

announcement/models/announcement.py
```diff
@@ -84,6 +85,16 @@ def _compute_allowed_user_ids(self):
             announcement.allowed_user_ids = announcement.user_group_ids.users
             announcement.allowed_users_count = len(announcement.user_group_ids.users)
 
+    @api.onchange("is_general_announcement")
+    def _onchange_is_general_announcement(self):
+        # If the option "General announcement" is checked, we assign the group
+        # "Internal users" to the user_group_ids field.
+        general_announcement = self.filtered(lambda x: x.is_general_announcement)
```
chienandalu: You don't need to loop over self. The onchange decorator is always triggered over a singleton

announcement/models/announcement.py
```diff
@@ -84,6 +85,16 @@ def _compute_allowed_user_ids(self):
...
+        for announcement in general_announcement:
+            announcement.announcement_type = "user_group"
+            group_user = self.env.ref("base.group_user")
+            announcement.user_group_ids = [(6, 0, [group_user.id])]
```
chienandalu: Why not to assing the record directly? This syntax is mandatory only for create/update/copy methods.

announcement/models/announcement.py
```diff
@@ -123,13 +134,16 @@ def _search_in_date(self, operator, value):
...
+    @api.onchange("announcement_type", "is_general_anouncement")
     def _onchange_announcement_type(self):
         """We want to reset the values on screen"""
         if self.announcement_type == "specific_users":
             self.user_group_ids = False
         elif self.announcement_type == "user_group":
             self.specific_user_ids = False
+        elif self.is_general_announcement:
+            self.specific_user_ids = False
+            self.user_group_ids = False
```
chienandalu: You already have an onchange for that feature. Don't mixup those logics

announcement/models/announcement.py
chienandalu: What if the user unchecks the checkbox?

announcement/views/announcement_views.xml
```diff
@@ -115,20 +118,23 @@
...
+                    name="announcement_scope"
+                    attrs="{'invisible': [('is_general_announcement', '=', True)]}"
+                >
                     <field name="announcement_type" />
                     <field
                         name="specific_user_ids"
                         string="Users"
                         options="{'no_create': True}"
-                        attrs="{'invisible': [('announcement_type', '!=', 'specific_users')], 'required': [('announcement_type', '=', 'specific_users')]}"
+                        attrs="{'invisible': [('announcement_type', '!=', 'specific_users')], 'required': [('announcement_type', '=', 'specific_users'), ('is_general_announcement', '=', False)]}"
```
chienandalu: Is this new visibility rule needed at all when you're setting it up on the parent node?

chienandalu (changes requested):

announcement/views/announcement_views.xml
pilarvargas-tecnativa: That rule is not for visibility, it is for required.

pilarvargas-tecnativa:
@chienandalu I have made the relevant changes, can you check that everything is correct? thanks :)

announcement/views/announcement_views.xml
chienandalu: Oh, yeah. In any case, It doesn't make sense to me. Can you explain it?

announcement/models/announcement.py
```diff
@@ -123,7 +134,7 @@ def _search_in_date(self, operator, value):
             ("notification_expiry_date", ">=", now),
         ]
 
-    @api.onchange("announcement_type")
+    @api.onchange("announcement_type", "is_general_anouncement")
```
chienandalu:
```suggestion
    @api.onchange("announcement_type")
```

announcement/models/announcement.py
```diff
@@ -84,6 +85,16 @@ def _compute_allowed_user_ids(self):
             announcement.allowed_user_ids = announcement.user_group_ids.users
             announcement.allowed_users_count = len(announcement.user_group_ids.users)
 
+    @api.onchange("is_general_announcement")
+    def _onchange_is_general_announcement(self):
+        self.specific_user_ids = False
+        self.user_group_ids = False
```
chienandalu: This would be better in the else to the next if

chienandalu (changes requested):

announcement/views/announcement_views.xml
pilarvargas-tecnativa: Going back to analyse everything, in the logic I was considering at the beginning if it was necessary to take into account this condition to override that the field was required and let the operation for a general announcement. With the improvements and changes made it is not necessary. Thanks

chienandalu: These attributes are for user interaction. If you're hiding the fields when you check that checkbox there's no possible user interaction. So no behaviour change is need at this level IMO

announcement/models/announcement.py
pilarvargas-tecnativa: thanks

announcement/models/announcement.py
```diff
@@ -84,6 +85,17 @@ def _compute_allowed_user_ids(self):
             announcement.allowed_user_ids = announcement.user_group_ids.users
             announcement.allowed_users_count = len(announcement.user_group_ids.users)
 
+    @api.onchange("is_general_announcement")
+    def _onchange_is_general_announcement(self):
+        # If the option "General announcement" is checked, we assign the group
+        # "Internal users" to the user_group_ids field.
+        if self.is_general_announcement:
+            self.announcement_type = "user_group"
+            self.user_group_ids = [(4, self.env.ref("base.group_user").id)]
```
chienandalu:
```suggestion
            self.user_group_ids = self.env.ref("base.group_user")
```

announcement/models/announcement.py
```diff
@@ -84,6 +85,17 @@ def _compute_allowed_user_ids(self):
             announcement.allowed_user_ids = announcement.user_group_ids.users
             announcement.allowed_users_count = len(announcement.user_group_ids.users)
 
+    @api.onchange("is_general_announcement")
+    def _onchange_is_general_announcement(self):
+        # If the option "General announcement" is checked, we assign the group
+        # "Internal users" to the user_group_ids field.
```
chienandalu:
```suggestion
        """If the option "General announcement" is checked, we assign the group
        "Internal users" to the user_group_ids field."""
```

chienandalu (changes requested):

announcement/models/announcement.py
pilarvargas-tecnativa: thanks!

chienandalu (approved):
Let's go :) (please review @victoralmau )

announcement/models/announcement.py
```diff
@@ -84,6 +85,17 @@ def _compute_allowed_user_ids(self):
             announcement.allowed_user_ids = announcement.user_group_ids.users
             announcement.allowed_users_count = len(announcement.user_group_ids.users)
 
+    @api.onchange("is_general_announcement")
+    def _onchange_is_general_announcement(self):
```
pedrobaeza: Instead of an onchange, you should convert `user_group_ids` to a computed writable field for not depending on the UI for making the "internal" change.

pilarvargas-tecnativa: But I have a doubt about this, in the compute method I must limit myself to calculate the value of the user_group_ids field, but to set the rest of the values, for example "self.specific_user_ids = False" I must keep the onchange, right? or would it be correct to include everything in the compute method?

pedrobaeza: It's not mandatory to assign the value for all the records:

```python
@api.depends("is_general_announcement")
def _compute_group_user_ids:
    for record in self:
        if record.is_general_announcement:
            record.group_user_ids = [(6, 0, self.env.ref("base.group_user").ids)]
```

Now, if you uncheck the mark, do you want to keep it the same group or not. I don't have an strong opinion on that.

pilarvargas-tecnativa: The behaviour should be as follows. If you tick the option "General announcement" you assign the group "Internal user" to the field "user_groups_id" but hide the field to select the announcement type which by default is selected the announcement type for specific user. If the "General announcement" option is unchecked, the field for selecting the announcement type reappears and both the "user_groups_id" and "specific_user_ids" fields must be clear.

pedrobaeza: Ok, then do that second part on the `else` of the compute.

pedrobaeza (approved):
/ocabot merge minor

PR #682, opened 2023-06-30
[14.0][IMP] base_tier_validation : add write possibility for reviewers

Kev-Roche:
Add write option for reviewers on the tier definition, allowing to directly change some values before validation. 
Option is false by default in order to keep the previous behavior.


sebastienbeau (commented):
Ok with the concept, can you check why test are red ?

Kev-Roche:
> Ok with the concept, can you check why test are red ?

Hi @sebastienbeau , I fix it.

sebastienbeau (approved):
LGTM (code review)

LoisRForgeFlow (approved):
LGTM

PR #687, opened 2023-07-05
[13.0][FIX]web_archive_date: ensure new archive fields in model

GuillemCForgeFlow:
Fixing for those models not having the newly added fields in the `base_archive_date` module, where it will fail when reading.

cc @ForgeFlow


JordiMForgeFlow (approved):

LoisRForgeFlow (approved):
Functional review!

Thanks for the test case

PR #690, opened 2023-07-11
[14.0][FIX] mass_editing ambigous sql column

sergiocorato:
This PR solves this bug in migration (translated from italian):
_psycopg2.ProgrammingError: <ERROR: reference to column "create_uid" is ambiguos>
LINE 20: create_uid,_


legalsylvain (approved):
/ocabot merge patch

PR #694, opened 2023-07-12
[16.0][ADD] base_tier_validation: compute validation flag dynamically

hbrunn:
with this the validation button shows immediately on a form and not only after save

@astirpe finally I find time to work on this again. I'm a bit at a loss as to why this doesn't work for you, it does [on runboat](https://github.com/OCA/account-invoicing/pull/1507) though, even with just domain `[('move_type', '=', 'in_invoice')]`:

image

image

### DIFF
--- a/base_tier_validation/models/tier_definition.py
+++ b/base_tier_validation/models/tier_definition.py
@@ -85,3 +85,32 @@ class TierDefinition(models.Model):
             rec.valid_reviewer_field_ids = self.env["ir.model.fields"].search(
                 [("model", "=", rec.model), ("relation", "=", "res.users")]
             )
+
+    @api.model_create_multi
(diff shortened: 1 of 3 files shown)


astirpe:
Tested in Runboat and it works perfectly!
Locally I still have the same problem, so I'm probably doing something wrong.

hbrunn:
do you reload your webclient after you've edited the domain? the onchange triggers are part of the form caching.
and i'm still not 100% convinced i do the registry updating right, so safest thing to do would be restart odoo + f5

astirpe:
Yes, I tried many times with restart reload, refresh, etc.. but this is the behavior:

Before saving
image

After saving
image

hbrunn:
do you possibly have a very old checkout of odoo?

PR #695, opened 2023-07-13
[15.0][IMP] announcement: Prevent False specific_user_ids from when not appropriate

victoralmau:
Prevent False `specific_user_ids` from when not appropriate.

Example use case:
- We create an announcement with the `create()` method and specify the users (`specific_user_ids`).
- The users are not defined.

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT43046

### DIFF
--- a/announcement/models/announcement.py
+++ b/announcement/models/announcement.py
@@ -103,7 +103,6 @@ class Announcement(models.Model):
                 announcement.announcement_type = "user_group"
                 announcement.user_group_ids = self.env.ref("base.group_user")
             else:
-                announcement.specific_user_ids = False
                 announcement.user_group_ids = False
 
     @api.depends("announcement_log_ids")


announcement/models/announcement.py
```diff
@@ -103,7 +103,8 @@ def _compute_user_group_ids(self):
                 announcement.announcement_type = "user_group"
                 announcement.user_group_ids = self.env.ref("base.group_user")
             else:
-                announcement.specific_user_ids = False
```
pedrobaeza, line 106: I think the patch may be:

```suggestion
                announcement.user_group_ids = False
```

pedrobaeza: @pilarvargas-tecnativa thumb rule for the next time: you can't update a field in the compute of another field.

pedrobaeza (approved):

announcement/models/announcement.py
pilarvargas-tecnativa: > @pilarvargas-tecnativa thumb rule for the next time: you can't update a field in the compute of another field.

I take note, thank you

pilarvargas-tecnativa (approved):

PR #696, opened 2023-07-13
[15.0][IMP] announcement: add attachments field

pilarvargas-tecnativa:
A new field has been added to be able to attach files to the announcement, and these attachments will be displayed in the announcement notification modal.

image

Modal: 
image

cc @Tecnativa TT42581

@CarlosRoca13 @victoralmau please review :)

### DIFF
--- a/announcement/models/announcement.py
+++ b/announcement/models/announcement.py
@@ -1,4 +1,5 @@
 # Copyright 2022 Tecnativa - David Vidal
+# Copyright 2022 Tecnativa - Pilar Vargas
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo import _, api, fields, models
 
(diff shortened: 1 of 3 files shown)


CarlosRoca13:
Doing a test in runbot I found the following error:

1- We create a new Announcement
2- We assign it to Marc Demo and we add an attachment
3- Log in as Marc Demo
4- We can see an error on the screen

If you try to access the Announcement created via Discuss > Announcements you will see the same error.

In the following gif you can see an example
anuncio error

pedrobaeza:
Pilar, try linking the attachment to the announcement (res_id and res_model fields).

pilarvargas-tecnativa:
> try linking the attachment to the announcement (res_id and res_model fields)

The relationship of the attachments is m2m, the announcement is linked in the res_id and res_model fields of the attachments in addition to a table where the ids of the announcements and the ids of the attachments are stored. I don't understand why a user allowed in that announcement can't see it in case there are attachments. I have tried to rewrite the res_id field of attachments to add the ad model and this way they can be displayed, although it was a test since it is not correct.
I don't know how exactly I could link the attachments to the announcement with the fields you mention as it is an m2m.

chienandalu (approved):
Works like charm :) Thank you both :+1:

pedrobaeza (approved):
/ocabot merge major

PR #697, opened 2023-07-18
[15.0][MIG] sequence_check_digit: Migration to 15.0

kaynnan:
cc @marcelsavegnago 

#390


marcelsavegnago:
@augustodinizl @aug1804 Can you do the functional review please?

sequence_check_digit/models/ir_sequence.py
```diff
@@ -37,17 +37,17 @@ def check_check_digit_formula(self):
         try:
             self.get_next_char(0)
         except Exception:
-            raise ValidationError(_("Format is not accepted"))
+            raise ValidationError(_("Format is not accepted")) from None
```
etobella, line 40: Why did you add the from None?

kaynnan: Due to the suggestion made by the pre-commit

PR #698, opened 2023-07-20
[16.0][FIX] unit test in sale_order_revision fails

CRogos:
Unfortunately this https://github.com/OCA/server-ux/pull/672 breaks one unit test in sale_order_revision.


pedrobaeza (approved):
Let's fast-track it to fix the tests:

/ocabot merge patch

PR #706, opened 2023-08-11
[16.0][MIG] base_tier_validation_report: Migration to 16.0

michaelslade54:
Standard migration.


dreispt (approved):

bosd (changes requested):
Can you re-trigger tests / runboat?

base_tier_validation_report/reports/tier_review_report.xml
```diff
@@ -0,0 +1,151 @@
...
+
+    <record id="tier_review_all_action" model="ir.actions.act_window">
+        <field name="name">All Reviews</field>
+        <field name="type">ir.actions.act_window</field>
+        <field name="res_model">tier.review</field>
+        <field name="view_mode">tree,form</field>
+        <field name="context">{}</field>
+        <field name="view_id" ref="tier_review_report_tree" />
+        <field name="search_view_id" ref="tier_review_report_search" />
+        <field
```
bosd, line 125: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0

Any view with groups_id on it now has to move such groups to the elements of the view. But now you can put groups attribute in a view field instead of isolating it on a view without fear of an access error.

PR #708, opened 2023-08-11
[16.0][FIX] Fix model read permissions for request_validation

apatyk:
This PR fixes the model read permissions for the `request_validation()` method in the `tier_validation` module. After the changes introduced by [#69120](https://github.com/odoo/odoo/pull/69120) in [Odoo 15.0](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0) the method most users interact with requires elevated permissions. These changes make it so any user can request validation (a typical user flow).

16 0-request-validation-user-error


apatyk:
Upon further review, this is not needed.

PR #712, opened 2023-08-29
[16.0][ADD] base_tier_validation: allow comments being optional

hbrunn:
this allows comments to be configured to be optional depending on the record being validated/rejected (the defaults are chosen such that existing definitions won't behave differently after the update)


bosd:
Looks good!
I know it is not by this commit.
But it is a bit confusing that the field `has_comment` actually means `can_comment`

Stephanvandenberg:
Hi,

This PR has been tested and works like a charm.

PR #714, opened 2023-08-31
[14.0][IMP] base_tier_validation: Add method for deciding whether the elimination of revisions is necessary.

victoralmau:
Add method for deciding whether the elimination of revisions is necessary.

Example of use case:
- Install `purchase_tier_validation` + `base_changeset_tier_validation` (https://github.com/OCA/server-tools/pull/2623).
- Create a `tier.validation` for `purchase.order`
- Create a rule (`changeset.rule`) related t a Tier validation so that when the `partner_id` field is changed a revision is created.
- Change `partner_id` field of a draft purchase order with a "basic user".

Previous behavior:
- When we change the status to "sent", revisions are removed (change pending validation, lost).

Current behavior:
- When we change the status to "sent" deletes revisions that do NOT have `changeset_id` linked (changes pending validation have not been deleted and are still displayed).

**Clarification**:
This problem does not exist only with `purchase_tier_validation` because no revisions are added in "draft" status (for example).

@Tecnativa TT43351

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -270,9 +270,24 @@ class TierValidation(models.AbstractModel):
                 and not rec._context.get("skip_validation_check")
             ):
                 raise ValidationError(_("The operation is under validation."))
-        if vals.get(self._state_field) in (self._state_from + [self._cancel_state]):
-            self.mapped("review_ids").unlink()
+            if rec._allow_to_remove_reviews(vals):
+                rec.mapped("review_ids").unlink()
         return super(TierValidation, self).write(vals)
+
+    def _allow_to_remove_reviews(self, values):
+        """Method for deciding whether the elimination of revisions is necessary."""
+        self.ensure_one()
+        state_to = values.get(self._state_field)
+        if not state_to:
+            return False
+        state_from = self[self._state_field]
+        # If you change to _cancel_state
+        if state_to in (self._cancel_state):
+            return True
+        # If it is changed to _state_from and it was not in _state_from
+        if state_to in self._state_from and state_from not in self._state_from:
+            return True
+        return False
 
     def _check_state_from_condition(self):
         return self.env.context.get("skip_check_state_condition") or (


pedrobaeza:
Shouldn't the patch to be that `_state_from` should allow more than one state?

victoralmau:
I have updated the description with the use case to explain this change.

pedrobaeza:
Reading the use case, I think it applies the same with or without `base_changeset_tier_validation`: printing the draft PO changes the state, and the revision is gone, so this should be handled IMO allowing multiple states.

What do you think, @LoisRForgeFlow ?

base_tier_validation/models/tier_validation.py
```diff
@@ -271,9 +271,13 @@ def write(self, vals):
             ):
                 raise ValidationError(_("The operation is under validation."))
         if vals.get(self._state_field) in (self._state_from + [self._cancel_state]):
```
LoisRForgeFlow: `_state_from` already supports more than one state, see [in purchases](https://github.com/OCA/purchase-workflow/blob/14.0/purchase_tier_validation/models/purchase_order.py#L10) 

I think the problem is in this condition, which needs to be a bit more complex an check that the origin state is not in self._state_from. In other words if you go from one of the _state_from to the same or other of the _state_from, the reviews shouln't be unlinked.

base_tier_validation/models/tier_validation.py
```diff
@@ -270,10 +270,24 @@ def write(self, vals):
...
             ):
                 raise ValidationError(_("The operation is under validation."))
-        if vals.get(self._state_field) in (self._state_from + [self._cancel_state]):
+        if rec._allow_to_remove_reviews(vals):
             self.mapped("review_ids").unlink()
         return super(TierValidation, self).write(vals)
 
+    def _allow_to_remove_reviews(self, values):
+        """Method for deciding whether the elimination of revisions is necessary."""
+        state_to = values.get(self._state_field)
```
pedrobaeza, line 280:
```suggestion
        self.ensure_one()
        state_to = values.get(self._state_field)
```

base_tier_validation/models/tier_validation.py
```diff
@@ -270,10 +270,24 @@ def write(self, vals):
...
+        if not state_to:
+            return False
+        state_from = self[self._state_field]
+        # If you change to _cancel_state
+        if state_to in (self._cancel_state):
+            return True
+        # If it is changed to _state_from and it was not in _state_from
+        if state_to in (self._state_from) and state_from not in (self._state_from):
```
pedrobaeza: Incorrect parenthesis:

```suggestion
        if state_to in self._state_from and state_from not in self._state_from:
```

pedrobaeza (approved):

sergio-teruel (approved):

Issue #597, opened 2023-02-14
[15.0] date_range: Ranges not displayed on date/datetime type filters


ao-landoo:
Since module date_range was migrated to version 15.0 date.range model registers are not displayed as filters for fields that are of type Date/Datetime.

In order to reproduce the behavior:
1- Go to Technical -> Date Range Types and create a register, for instance name = "MONTH"
2- Go to Technical -> Date Range and create a register, for instance type "MONTH" and from 1-feb to 28-feb
3- Go to sales order list view, and try to filter by create date
4- It's not allowing to filter by MONTH date range, it's not appearing

It's working fine on runboat 14.0

![Captura de pantalla de 2023-02-14 17-02-48](https://user-images.githubusercontent.com/74420754/218791645-e84c29b3-7c42-4905-a759-8fdaa46c11c0.png)

@JoanMForgeFlow I'm mentioning you as you're who migrated it. Any idea what could be wrong?

pedrobaeza:
@JoanMForgeFlow can you please check that missing bits from the migration?

alan196:
I have the same issue, I have tested in my local with Odoo and OCB, and also tested in this [runboat](http://oca-server-ux-15-0-21ec89416d3b.runboat.odoo-community.org/)

RadjaCyb:
> Since module date_range was migrated to version 15.0 date.range model registers are not displayed as filters for fields that are of type Date/Datetime.
> 
> In order to reproduce the behavior: 1- Go to Technical -> Date Range Types and create a register, for instance name = "MONTH" 2- Go to Technical -> Date Range and create a register, for instance type "MONTH" and from 1-feb to 28-feb 3- Go to sales order list view, and try to filter by create date 4- It's not allowing to filter by MONTH date range, it's not appearing
> 
> It's working fine on runboat 14.0
> 
> ![Captura de pantalla de 2023-02-14 17-02-48](https://user-images.githubusercontent.com/74420754/218791645-e84c29b3-7c42-4905-a759-8fdaa46c11c0.png)
> 
> @JoanMForgeFlow I'm mentioning you as you're who migrated it. Any idea what could be wrong?

Hello,

Did you find any solution please ?

RadjaCyb:
> I have the same issue, I have tested in my local with Odoo and OCB, and also tested in this [runboat](http://oca-server-ux-15-0-21ec89416d3b.runboat.odoo-community.org/)

Did you find any solution please ?

PR #715, opened 2023-08-31
[15.0][FIX] date_range: Allow select date_ranges on custom filters

CarlosRoca13:
This commit solves the problems reported on Issue https://github.com/OCA/server-ux/issues/597

cc @Tecnativa TT43442

ping @pedrobaeza @sergio-teruel 

I tried to set the description provided on the date_range but it needs to overwrite the onApply method with not a simply way to do it. As the description shows the range you are filtering by, I don't think this is a very important feature to keep.

### DIFF
--- a/date_range/static/src/js/date_range.esm.js
+++ b/date_range/static/src/js/date_range.esm.js
@@ -1,76 +1,77 @@
 /** @odoo-module **/
-import {patch} from "web.utils";
+import {patch} from "@web/core/utils/patch";
 import {_lt} from "@web/core/l10n/translation";
 import {session} from "@web/session";
 import {FIELD_OPERATORS, FIELD_TYPES} from "web.searchUtils";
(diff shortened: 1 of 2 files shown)


sergio-teruel (approved):
Great!! @CarlosRoca13 

Thanks!!!

lmignon (approved):
Thank you for the fix @CarlosRoca13 LGTM (Code + functional)

pedrobaeza:
@CarlosRoca13 one thing: I did the range from 2023-01-01 to 2023-12-31. Why it puts the next day on the filter?

Selección_019

pedrobaeza:
And please fw-port it to v16.

CarlosRoca13:
> 2023-01-01 to 2023-12-31

image

I can't reproduce it...

pedrobaeza:
OK, it seems my fault.

PR #716, opened 2023-09-04
[14.0][IMP] base_tier_validation: Change reviews table + Add Validate / Reject buttons

victoralmau:
Change reviews table (remove review templates + javascript).

This change makes the tables to be defined the same as the rest of Odoo and allows for easier inheritance.

Extra improvement: If we have installed `web_tree_many2one_clickable` the many2one fields are now clickable keeping the breadcrumb: https://github.com/OCA/server-tools/pull/2623#issuecomment-1706019703

**Before**:
antes

**After**:
despues

Add Validate + Reject buttons from Reviews table.
ejemplo

@Tecnativa TT43351


base_tier_validation/models/res_users.py
```diff
@@ -41,6 +41,7 @@ def review_user_count(self):
                     }
         return list(user_reviews.values())
 
+    # TODO: No longer used
```
pedrobaeza:
```suggestion
    # TODO: No longer used in ... To be removed in vXX
```

pedrobaeza (commented):
Looking at the changes, I think the initial intention was to avoid to inherit the view everywhere, and instead, to be injected automatically. And I would say the new view is uglier, isn't it?

What do you think, @LoisRForgeFlow ?

victoralmau:
Updated the initial description with images, other improvements and problem solved with this change.

victoralmau:
Finally, the existing approach for the revision table seems more correct and flexible.

PR #727, opened 2023-09-25
Added logic to extend base_tier_validation module with hr department

gcaceres93:
Added logic to extend base_tier_validation module with logic to select a department from the HR module as an option for the reviewers criteria


bosd:
@gcaceres93 
Can you fix tests and pre-commit?
Please make comments in the code in English.

Issue #728, opened 2023-10-10
base_tier_validation: wrong counter for pending review


OpenCode:
## Module
base_tier_validation
- [ ] 11.0
- [ ] 12.0
- [ ] 13.0 
- [ ] 14.0 
- [ ] 15.0 
- [ ] 16.0 - https://github.com/OCA/server-ux/pull/729

## Describe the bug
In some cases, the pending review counter reports a higher value than in reality. In fact, clicking on the relevant item shows fewer records than the number indicated. This happens because records in deleted status are also counted

## To Reproduce
- Install base_tier_validation and another  tier validation (for example sale).
- Create a tier validation rule and create 2 orders in the validation domain
- Cancel one of the sales
The pending count will show `2 Pending Reviews`

PR #729, opened 2023-10-10
[FIX] base_tier_validation: Exclude cancel records from pending count

OpenCode:
Fix https://github.com/OCA/server-ux/issues/728


LoisRForgeFlow (approved):
I've been able to reproduce in local and test your fix. LGTM

PR #730, opened 2023-10-17
[14.0][IMP] base_tier_validation: Add validation_status field to tier.validation

victoralmau:
Add `validation_status` field to `tier.validation`

@Tecnativa TT33369

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -31,15 +31,28 @@ class TierValidation(models.AbstractModel):
         auto_join=True,
     )
     to_validate_message = fields.Html(compute="_compute_validated_rejected")
+    # TODO: Delete in v17 in favor of validation_status field
     validated = fields.Boolean(
         compute="_compute_validated_rejected", search="_search_validated"
     )
     validated_message = fields.Html(compute="_compute_validated_rejected")
     need_validation = fields.Boolean(compute="_compute_need_validation")
+    # TODO: Delete in v17 in favor of validation_status field
     rejected = fields.Boolean(
         compute="_compute_validated_rejected", search="_search_rejected"
     )
     rejected_message = fields.Html(compute="_compute_validated_rejected")
+    # Informative field (used in purchase_tier_validation), will be reliable as of v17
+    validation_status = fields.Selection(
+        selection=[
+            ("no", "Without validation"),
+            ("pending", "Pending"),
+            ("rejected", "Rejected"),
+            ("validated", "Validated"),
+        ],
+        default="no",
+        compute="_compute_validation_status",
+    )
     reviewer_ids = fields.Many2many(
         string="Reviewers",
         comodel_name="res.users",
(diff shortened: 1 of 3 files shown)


base_tier_validation/models/tier_validation.py
```diff
@@ -171,6 +181,17 @@ def _compute_validated_rejected(self):
             rec.rejected_message = rec._get_rejected_message()
             rec.to_validate_message = rec._get_to_validate_message()
 
+    def _compute_validation_status(self):
```
LoisRForgeFlow, line 187: Instead of adding a new compute, can you rely on _compute_validated_rejected to not have duplicated logic or a possible deviation in the way of calculating what "validated" means.

validated True and rejected False -> status validated
validated False and rejected True -> status rejected
validated False and rejected False + review_ids exist -> status pending

LoisRForgeFlow (commented):
This opens a good debate, now that I see validation_status I like it better than the current `validated` and `rejected` booleans... however, I don't think it is worth to refactor in current stables versions. So let's try to make these coexist without creating a big mess. See my comment:

base_tier_validation/models/tier_validation.py
LoisRForgeFlow: You can add a comment for future version 17 in which we suggest to fully replace booleans by the validations_status.

pedrobaeza: Another option is to rely on this field, and let the other fields `rejected` and `validated` as computed for backwards compatibility.

LoisRForgeFlow: But the main driver of all flows, views, etc. continues to be the booleans. The transition can be done with backward compatibility in v17, but in current stable verions I wouldn't change how the booleans values are computed to avoid surprises.

pedrobaeza: OK, it's like that right now.

pedrobaeza (approved):

base_tier_validation/models/tier_validation.py
LoisRForgeFlow: No, it is not like that now. I prefer to do this https://github.com/OCA/server-ux/pull/730#discussion_r1363897379. Now the status is based on one logic and not on the booleans, that have another logic. That's exactly what I want to avoid.

LoisRForgeFlow (changes requested):
while open thread is not solved

base_tier_validation/models/tier_validation.py
victoralmau: Changes done.

LoisRForgeFlow (approved):
Thank you! LGTM

bosd:
@victoralmau 
Just asking, are these changes still compatible with `base_tier_validation_waiting`?

victoralmau:
> @victoralmau Just asking, are these changes still compatible with `base_tier_validation_waiting`?

This new field does not change any existing behavior (https://github.com/OCA/server-ux/pull/730#discussion_r1365312450), it is only something informative that can be used in other modules.

PR #731, opened 2023-10-20
[FIX] date_range: prevent filters to be added multiple times

espo-tony:
Module date_range automatically adds, to all date and datetime filtering options, a new item for each Date Range Type defined into the system.
Under some circumstances, when the method responsible for it is invoked, it works on a data structure ( `OPERATORS.date` / `OPERATORS.datetime` ) which wasn't preventively cleaned-up. If that happens, because this method is invoked each time the user clicks on the **Filters** button, the result is to have multiple times the same Date Range Type options.

https://github.com/OCA/server-ux/assets/8192093/3de50aa1-8813-4a1c-8142-c66862342754

This PR aims at making the module more robust to this kind of situations by adding a prior check when adding a Date Range Type option to filters.

### DIFF
--- a/date_range/static/src/js/date_range.esm.js
+++ b/date_range/static/src/js/date_range.esm.js
@@ -30,14 +30,16 @@ patch(CustomFilterItem.prototype, "date_range.CustomFilterItem", {
                     date_range: true,
                     date_range_type: range_type,
                 };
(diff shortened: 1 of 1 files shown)


CasVissers-360ERP (approved):

bosd (approved):
LGTM

PR #733, opened 2023-10-22
[FIX]base_tier_validation_report: attempt to fix ci

bosd:
Fixes #734

PR #735, opened 2023-10-24
[15.0][FIX] base_substate: track template on substate

Saran440:
Move function `_track_template()` on base_substate_mixin to mail_thread
Currently, Email Template on Base Substate is not working.

Step to error:
1. Create base substate and add Email Template
2. Change substate to substate there is email template, it will not send email template.

### DIFF
--- a/base_substate/tests/test_base_substate.py
+++ b/base_substate/tests/test_base_substate.py
@@ -17,6 +17,14 @@ class TestBaseSubstate(common.TransactionCase):
         cls.substate_test_sale = cls.env["base.substate.test.sale"]
         cls.substate_test_sale_line = cls.env["base.substate.test.sale.line"]
 
+        cls.mail_template = cls.env["mail.template"].create(
+            {
+                "name": "Waiting for legal documents",
+                "model_id": cls.substate_test_sale.id,
+                "subject": "Test Email Substate",
+            }
+        )
+
         cls.base_substate = cls.env["base.substate.mixin"]
         cls.substate_type = cls.env["base.substate.type"]
 
@@ -68,6 +76,7 @@ class TestBaseSubstate(common.TransactionCase):
                 "name": "Waiting for legal documents",
                 "sequence": 2,
                 "target_state_value_id": cls.substate_val_sale.id,
+                "mail_template_id": cls.mail_template.id,
             }
         )
 
@@ -106,12 +115,17 @@ class TestBaseSubstate(common.TransactionCase):
         )
         self.assertTrue(so_test1.state == "draft")
         self.assertTrue(so_test1.substate_id == self.substate_under_negotiation)
-
+        self.assertNotIn(
+            self.mail_template.subject, so_test1.message_ids.mapped("subject")
+        )
         # Test that validation of sale order change substate_id
         so_test1.button_confirm()
         self.assertTrue(so_test1.state == "sale")
         self.assertTrue(so_test1.substate_id == self.substate_wait_docs)
-
+        # Check some message_ids are created and sent email
+        self.assertIn(
+            self.mail_template.subject, so_test1.message_ids.mapped("subject")
+        )
         # Test that substate_id is set to false if
         # there is not substate corresponding to state
         so_test1.button_cancel()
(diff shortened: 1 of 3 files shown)


bosd:
I can confirm the issue.
Now everybode seems to be patching it in their `*_substate` module.

Test on this one are failling.
Unit tests on 15.0 branch are currently broken. Fix in: https://github.com/OCA/server-ux/pull/733
Can you review the fix?

bosd (changes requested):
See comments in: https://github.com/OCA/server-ux/pull/736#pullrequestreview-1709775298

bosd:
Just tested this locally. The error still exsists. 

```
  File "/opt/odoo/custom/src/odoo/odoo/service/server.py", line 1260, in preload_registries
    registry = Registry.new(dbname, update_module=update_module)
  File "/opt/odoo/custom/src/odoo/odoo/modules/registry.py", line 87, in new
    odoo.modules.load_modules(registry, force_demo, status, update_module)
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 470, in load_modules
    processed_modules += load_marked_modules(cr, graph,
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 363, in load_marked_modules
    loaded, processed = load_module_graph(
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 198, in load_module_graph
    registry.setup_models(cr)
  File "/opt/odoo/custom/src/odoo/odoo/modules/registry.py", line 279, in setup_models
    model._prepare_setup()
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 2975, in _prepare_setup
    cls.__bases__ = cls.__base_classes
TypeError: Cannot create a consistent method resolution
order (MRO) for bases BaseModel, base, mail.thread, mail.activity.mixin, base.substate.mixin

```

Saran440:
@bosd Thank you for checked.
Error because this model `base.substate.mixin` use `mail.thread` inherit and other module such as hr_expense_substate inherit `mail.thread` already
So, i deleted inherit and someone must inherit `mail.thread` itself if model not support

base_substate/tests/test_base_substate.py
```diff
@@ -106,12 +115,10 @@ def test_sale_order_substate(self):
         )
         self.assertTrue(so_test1.state == "draft")
         self.assertTrue(so_test1.substate_id == self.substate_under_negotiation)
-
-        # Test that validation of sale order change substate_id
+        # Test that validation of sale order change substate_id and send email
```
bosd: Thanks for all your work on this. ❤️  Can you actually test the mail message?

Saran440: OK, I added test mail message in test script please check.

bosd (approved):

bosd:
@Saran440 Can you please force push again. To see if tests pass this time?

Dranyel-Bosd (approved):
LGTM :100:

bosd:
@oca/server-ux-maintainers Can we get a merge here?

StefanRijnhart (commented):
The issue seems to be that _track_template of the mixin is bypassed in the ORM inheritance MRO. The solution put forward earlier today in https://github.com/OCA/sale-workflow/pull/3483/files is to change the inheritance order.
The solution in this PR is to refactor out the method.
I would like the solutions to converge, but I don't know which approach is better. Do you have any arguments for either? @Saran440 @bosd @len-foss

len-foss:
Well, you could argue "my" solution is a bit more finicky because you need to make sure that inherits are in the right order, but this is something that is true in all the rest of Odoo's ORM (and arguably all of OOP).

But I think the reason this solution is worse is that this bypass the mail being sent in the post commit hook. This feature has been done to avoid a flow that would crash after it has sent an email (you can imagine the spam as the backend user will never see the message in DB). It's better to let mail.thread handle everything.

bosd:
> Well, you could argue "my" solution is a bit more finicky

So should we proceed with this one?

StefanRijnhart:
I think the usage of the post commit hook is a strong argument. We can therefore go with @len-foss's solution or refactor the solution proposed here such that the outgoing mail is only triggered in the post commit hook. That is probably best of both worlds, but it requires additional development.

pedrobaeza (changes requested):
Same as https://github.com/OCA/server-ux/pull/736#issuecomment-2732021000

pedrobaeza (approved):
Clarified in the other PR.

/ocabot merge patch

PR #737, opened 2023-10-24
[15.0][IMP] base_tier_validation: Add validation_status field to tier.validation

victoralmau:
FWP from 14.0: https://github.com/OCA/server-ux/pull/730

Add `validation_status` field to `tier.validation`

@Tecnativa TT33369

### DIFF
--- a/base_tier_validation/i18n/base_tier_validation.pot
+++ b/base_tier_validation/i18n/base_tier_validation.pot
@@ -6,6 +6,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 15.0\n"
 "Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2023-10-24 06:38+0000\n"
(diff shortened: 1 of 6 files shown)


pedrobaeza:
Add here also the extra column with decoration for doing it in one shot.

victoralmau:
> Add here also the extra column with decoration for doing it in one shot.

It is not possible to add the extra column here, the column is added to an abstract model that is inherited in others (`purchase.order` for example) and it is in those views specifically (if desired) where it has to be added.

pedrobaeza:
I don't get the sense of the new commits.

victoralmau:
> I don't get the sense of the new commits.

The latest commits are related to resolving the test error: https://github.com/OCA/server-ux/actions/runs/6623289456/job/17990083714#step:8:1520

pedrobaeza (approved):

chienandalu (approved):

PR #740, opened 2023-10-27
[PERF] base_tier_validation: filter records with reviews before compute fields

celm1990:
Before this commit, in a database with many records candidates for applying reviews, fields with a `function search` defined were very slow because the search was performed over all records with these states. After this commit, first filter only records with `review_ids` set, which significantly enhancement performance by more than 10X!!!

Database with more than 10K Sale Orders on draft, sent(states defined for review). Less than 1K records has started a flow review

image

### Before change
**search_rejected**
image

**search_validated**

image

### After this commit
**search_rejected**
image

**search_validated**
image

fixed https://github.com/OCA/server-ux/issues/496

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -106,9 +106,9 @@ class TierValidation(models.AbstractModel):
     def _search_validated(self, operator, value):
         assert operator in ("=", "!="), "Invalid domain operator"
         assert value in (True, False), "Invalid domain value"
-        pos = self.search([(self._state_field, "in", self._state_from)]).filtered(
-            lambda r: r.validated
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (approved):
Thanks for contributing, it looks correct to me :+1:

jfialloss (approved):
Great job!!!

pedrobaeza:
Please fw-port it to upper versions.

celm1990:
> Please fw-port it to upper versions.

@pedrobaeza sorry for question but this fw-port is done automatically or I must be create 1 PR for each branch?

pedrobaeza:
You should create them, thanks.

PR #744, opened 2023-11-01
[15.0][PERF] base_tier_validation: filter records with reviews before compu…

celm1990:
…te fields

Before this commit, in a database with many records candidates for applying reviews, fields with a defined search function were very slow because the search was performed over all records with these states. After this commit, first filter only records with review_ids set, which significantly reduces performance by more than 10X!!!

Forward port of https://github.com/OCA/server-ux/pull/740


pedrobaeza:
I'm now wondering if the `filtered` condition can be done also as an extra element in the search domain.

bosd (approved):
LGTM

LoisRForgeFlow (approved):
Let's for now merge this versions to align versions.

Further refinements can still be done after this.

/ocabot merge patch

PR #751, opened 2023-11-07
[15.0][FIX] base_search_custom_field_filter: Force name on custom filters

CarlosRoca13:
Name is required to use the field for filters, but this one is not set on pivot and graph views, by doing this change we are making sure that this field is filled.

cc @Tecnativa TT45828

ping @carlosdauden @pedrobaeza

### DIFF
--- a/base_search_custom_field_filter/models/base.py
+++ b/base_search_custom_field_filter/models/base.py
@@ -61,6 +61,9 @@ class Base(models.AbstractModel):
             field = custom_filter._get_related_field()
             field_name = custom_filter.expression
             res["fields"][field_name] = field.get_description(self.env)
+            # Force name for avoiding error accesing to this field from pivot and graph
+            # (needed for owl framework)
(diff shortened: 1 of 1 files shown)


base_search_custom_field_filter/models/base.py
```diff
@@ -61,6 +61,9 @@ def load_views(self, views, options=None):
             field = custom_filter._get_related_field()
             field_name = custom_filter.expression
             res["fields"][field_name] = field.get_description(self.env)
+            # Force name for avoiding error accesing to this field from pivot and graph
+            # (needed for owl framework)
+            res["fields"][field_name]["name"] = field_name
```
chienandalu, line 66: Is it the technical one?

CarlosRoca13: yes, it corresponds to the expression used in the definition of the custom filter field

chienandalu (approved):

carlosdauden (approved):

PR #753, opened 2023-11-09
[15.0][FIX]date_range: Fixed portal access denied

ChristophAbenthungCibex:
The module "date_range" is not usable for portal users.
This fix adds the needed permissions to access the date_range for portal users. The added permissions are the same as a regular user.

Base Odoo module "portal" is required to login to the portal.
Another module which uses the date_range module is needed. For example the base Odoo module "project".

Current behavior:
The portal user accesses a project he got shared editable. When he tries to open the filter menu he gets an access denied error. 

Steps to reproduce:
1. Login as portal user
2. open a project that got shared "editable" (the portal user can edit/ write on that project)
3. Click on button "filter"
4. The access denied error appears

[Date_Range_Project_access_denied.webm](https://github.com/OCA/server-ux/assets/132348854/a62d6ad7-f121-45b5-a37e-d0ff5ad1ef09)

### DIFF
--- a/date_range/security/ir.model.access.csv
+++ b/date_range/security/ir.model.access.csv
@@ -1,6 +1,8 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 access_date_range_date_range,date_range.date_range,model_date_range,base.group_user,1,0,0,0
 access_date_range_date_range_type,date_range.date_range_type,model_date_range_type,base.group_user,1,0,0,0
+access_date_range_date_range_portal,date_range.date_range,model_date_range,base.group_portal,1,0,0,0
+access_date_range_date_range_type_portal,date_range.date_range_type,model_date_range_type,base.group_portal,1,0,0,0
 access_date_range_date_range_config,date_range.date_range.config,model_date_range,base.group_system,1,1,1,1
 access_date_range_date_range_type_config,date_range.date_range_type.config,model_date_range_type,base.group_system,1,1,1,1
 access_date_range_generator,access_date_range_generator,model_date_range_generator,base.group_system,1,1,1,1


pedrobaeza:
I don't get why the portal requires this permission. I have just tried on runboat before this PR (http://oca-server-ux-15-0-e81102ffd3d5.runboat.odoo-community.org/my/tasks?) using portal/portal, and no problem found. This can be something on your side.

ChristophAbenthungCibex:
The portal user only needs the access rights to date_range when the portal user has write rights on the project. See http://oca-server-ux-15-0-e81102ffd3d5.runboat.odoo-community.org/my/project/1?
Is it possible that date range is not active for the read-only view?
access_denied_date_range

pedrobaeza:
Tested and no problem on my side:

Peek 2023-11-09 13-34

If you don't put exact steps to reproduce the problem, it's hard to reproduce it. It's true that the full project sharing is a special case of "backend" view in portal.

ChristophAbenthungCibex:
Sorry if I described the steps to little. I thought it was clear enough.
updated the pull request description.

Steps to reproduce:
1. Login as portal user
2. open a project that got shared "editable" (the portal user can edit/ write on that project)
3. Click on button "filter"
4. The access denied error appears

pedrobaeza:
It was clear for you, but not for the rest. Thanks for telling and editing the main comment including them.

Indeed that case is not covered. As said, this only happens because the project sharing is adding a backend view in portal. Not sure if adding such read permission is "too much". The only other path I see is to not show ranges on that special case (can it be detected?).

I let the usual collaborators of this module to give their opinion.

lmignon (approved):
As it's a read access and date ranges are not sensitive data, I've no objection to grant the read access to the protal users.

ChristophAbenthungCibex:
Any news?

PR #754, opened 2023-11-11
[13.0][IMP] base_tier_validation: filter tier definition for the current company

celm1990:
In a multicompany environment as follows:
User Request. Allowed Companies: C1, C2. Selected Companies: `C1`, `C2`, Current Company `C1` 
User Approver: Allowed companies: C1, C2, C3. Selected Companies: `C1` Current Company `C1` 

Before this commit, when a user requested validation with multicompany enabled (`C1, C2`), definitions of other companies could be applied. If the user approver had access to fewer companies (`C1`) and confirmed the document, then the remaining reviews (`C2`) were ignored in the pending state. These reviews were shown on the sidebar but could never be approved because the document had already been approved. 

After this commit, tier definitions are filtered only by the current company of the user request.

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -195,7 +195,12 @@ class TierValidation(models.AbstractModel):
 
     def _compute_need_validation(self):
         for rec in self:
-            tiers = self.env["tier.definition"].search([("model", "=", self._name)])
+            tiers = self.env["tier.definition"].search(
+                [
+                    ("model", "=", self._name),
+                    ("company_id", "in", [False] + self.env.company.ids),
+                ]
+            )
             valid_tiers = any([rec.evaluate_tier(tier) for tier in tiers])
             rec.need_validation = (
                 not rec.review_ids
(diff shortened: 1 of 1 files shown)


celm1990:
@LoisRForgeFlow @pedrobaeza, could you please provide feedback on whether these changes can be merged or if they might violate any business rules? I believe it's more common for reviews to be applied by the current company rather than all selected companies. What are your thoughts on this?

base_tier_validation/models/tier_validation.py
```diff
@@ -195,7 +195,14 @@ def _calc_reviews_rejected(self, reviews):
 
     def _compute_need_validation(self):
         for rec in self:
-            tiers = self.env["tier.definition"].search([("model", "=", self._name)])
+            tiers = self.env["tier.definition"].search(
+                [
+                    ("model", "=", self._name),
+                    "|",
+                    ("company_id", "=", self.env.company.id),
+                    ("company_id", "=", False),
```
pedrobaeza:
```suggestion
                    ("company_id", "in", [False] + self.env.company.ids),
```

And the same for the other.

pedrobaeza (approved):

LoisRForgeFlow (approved):
Reviewed locally, LGTM

Thanks for the fix!

LoisRForgeFlow:
@celm1990 Could you forward port this to newer versions?

PR #756, opened 2023-11-14
[14.0][IMP] base_tier_validation: Add `_prepare_tier_review_vals()` methot to allow inheritance.

victoralmau:
Add `_prepare_tier_review_vals()` methot to allow inheritance.

@Tecnativa TT45804


chienandalu (commented):
I was trying a functional test but I'm not ble to create a tier definition manually as the referenced model field doesn't seem to respond in the form. Is it a know issue or something?

victoralmau:
You are right, to test it will be necessary to install `purchase_tier_validation` for example in order to test the use case.

pedrobaeza:
Isn't possible to define demo data attached to an existing model (res.partner or res.user?) to test it?

victoralmau:
> Isn't possible to define demo data attached to an existing model (res.partner or res.user?) to test it?

I think that in this case, demo data would not be enough.

We could do it with `res.users` model (because it has a `state` field) but as we will need to add https://github.com/OCA/purchase-workflow/blob/14.0/purchase_tier_validation/models/purchase_order.py and https://github.com/OCA/purchase-workflow/blob/14.0/purchase_tier_validation/models/tier_definition.py I understand that it is not possible and for that we will need to create a module `test_base_tier_validation` (or similar) that will only be used for that. That change could also cause the currently existing `base_tier_validation` tests to be moved to that module (probably making `FakeModelLoader` unnecessary).

pedrobaeza:
OK, create that module, but don't move tests. We still want to unit tests this module in a isolated way.

victoralmau:
Ok, the module `test_base_tier_validation` has been added, you can test now with the partners.

LoisRForgeFlow (commented):
Can you expand a bit on the use case?

Comment is more thought as a user input rather than an extra description generated by the system.

pedrobaeza:
Can you show visually the goal?

victoralmau:
Example of configuration and result:
ejemplo-1
ejemplo-2

pedrobaeza:
Isn't this something for base_changeset_tier_validation?

LoisRForgeFlow:
> Isn't this something for base_changeset_tier_validation?

Indeed, I think so. In the base module that `comment` is dedicated for reviews from definitions with option "Comment":
image

So then when you click validate or reject a wizard ask for you comment:
image

image

victoralmau:
> Isn't this something for base_changeset_tier_validation?

With `base_changeset_tier_validation` installed there are 2 possible routes:
1. Direct validation request by status change: The comment to be defined in the revision will be the one related to this PR.
2. Revision created by a changeset: The comment to be defined will be similar to this PR, except that the before and after value will be shown. Untaxed amount: 600 > 800

victoralmau:
According to https://github.com/OCA/server-ux/pull/756#issuecomment-1811970679, it will then be necessary to create a new `summary` field in the revisions, add it to the revisions table and populate that field with the indicated behavior. Do you agree?  The commented change should be in `base_tier_validation` or in `base_changeset_tier_validation`?

LoisRForgeFlow:
> According to [#756 (comment)](https://github.com/OCA/server-ux/pull/756#issuecomment-1811970679), it will then be necessary to create a new `summary` field in the revisions, add it to the revisions table and populate that field with the indicated behavior. Do you agree? The commented change should be in `base_tier_validation` or in `base_changeset_tier_validation`?

For me your use case for this "summary" is very tight to specific changeset reviews, so I would put it `base_changeset_tier_validation` to keep base module clean.

I would also suggest that instead of adding a new field, you could play with the existing description of the review and extend it so it would look like this: "You definition name - Vendor: partner A -> partner B". Wouldn't that cover your use case?

LoisRForgeFlow (approved):
LGTM

chienandalu (approved):
Thanks :)

pedrobaeza (approved):
/ocabot merge patch

PR #757, opened 2023-11-14
[14.0][IMP] base_tier_validation: filter tier definition for the current co…

celm1990:
…mpany

In a multicompany environment as follows:
User Request. Selected Companies: C1, C2, Current Company C1 User Approver: Selected Companies: C1 Current Company C1 Before this commit, when a user requested validation with multicompany enabled (C1, C2), definitions of other companies could be applied. If the user approver had access to fewer companies (C1) and confirmed the document, then the remaining reviews (C2) were ignored in the pending state. These reviews were shown on the sidebar but could never be approved because the document had already been approved. After this commit, tier definitions are filtered only by the current company of the user request.

Forward port of https://github.com/OCA/server-ux/pull/754

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -223,7 +223,12 @@ class TierValidation(models.AbstractModel):
             if isinstance(rec.id, models.NewId):
                 rec.need_validation = False
                 continue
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

LoisRForgeFlow (approved):
Thanks!

/ocabot merge patch

PR #764, opened 2023-11-20
[14.0][FIX] base_action_visibility_restriction : fix access error for ir.actions.server 

GSLabIt:
# Description of the problem or feature or link to the task

Current behavior prior to this PR:

an error is raised when looping on server actions to get exclude groups:

error

Desired behavior after this PR:

error is not raised


francesco-ooops (approved):
Functional ok!

francesco-ooops:
@JordiMForgeFlow good for you?

JordiMForgeFlow:
Hi @francesco-ooops 

Could you describe a bit when do you get the error?

francesco-ooops:
Hi @JordiMForgeFlow, 

I could not reproduce the issue on runboat but after installing the module in two different environments (without adding any data to employ the module's features) some users without "Administration/Settings" group would receive the access error from the screenshot when trying to access website and CRM app, respectively

We tested the fix and it's solving the issue.

JordiMForgeFlow:
Hi @francesco-ooops 

I think you should try to check why the 'run()' is being executed when entering the app as maybe there is another solution instead of using sudo(), which should be a last resort

francesco-ooops:
> Hi @francesco-ooops
> 
> I think you should try to check why the 'run()' is being executed when entering the app as maybe there is another solution instead of using sudo(), which should be a last resort

@JordiMForgeFlow Sorry, just for clarification as I'm not technical: isn't the module designed to execute the 'run()' when entering the app?

Or are you suggesting to investigate whether it could be run at another time?

JordiMForgeFlow:
@francesco-ooops I have been reviewing the code from Odoo and the suggested fix is correct. What I was saying before is that I was expecting some justification on why the proposed fix was to simply put sudo(). Turns out Odoo uses also the sudo() in this method so LGTM! :)

JordiMForgeFlow (approved):

ilyasProgrammer (approved):
LGTM

francesco-ooops:
@JordiMForgeFlow shall we merge?

PR #762, opened 2023-11-17
[15.0][IMP] base_tier_validation: Add _prepare_tier_review_vals() methot to allow inheritance

victoralmau:
FWP from 14.0: https://github.com/OCA/server-ux/pull/756

Add `_prepare_tier_review_vals()` methot to allow inheritance.

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT45804

PR #776, opened 2023-11-28
[16.0][IMP] base_tier_validation: Add _prepare_tier_review_vals() methot to allow inheritance

victoralmau:
FWP from 15.0: https://github.com/OCA/server-ux/pull/762

Add `_prepare_tier_review_vals()` methot to allow inheritance.

@Tecnativa TT45804

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -455,33 +455,35 @@ class TierValidation(models.AbstractModel):
                     body=rec._notify_requested_review_body(),
                 )
 
+    def _prepare_tier_review_vals(self, definition, sequence):
+        return {
+            "model": self._name,
+            "res_id": self.id,
+            "definition_id": definition.id,
+            "requested_by": self.env.uid,
+            "sequence": sequence,
+        }
+
     def request_validation(self):
         td_obj = self.env["tier.definition"]
-        tr_obj = created_trs = self.env["tier.review"]
+        tr_obj = self.env["tier.review"]
+        vals_list = []
         for rec in self:
-            if rec._check_state_from_condition():
-                if rec.need_validation:
-                    tier_definitions = td_obj.search(
-                        [
-                            ("model", "=", self._name),
-                            ("company_id", "in", [False] + self.env.company.ids),
-                        ],
-                        order="sequence desc",
-                    )
-                    sequence = 0
-                    for td in tier_definitions:
-                        if rec.evaluate_tier(td):
-                            sequence += 1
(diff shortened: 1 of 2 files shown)


base_tier_validation/models/tier_validation.py
```diff
@@ -472,15 +481,8 @@ def request_validation(self):
...
-                                {
-                                    "model": self._name,
-                                    "res_id": rec.id,
-                                    "definition_id": td.id,
-                                    "sequence": sequence,
-                                    "requested_by": self.env.uid,
-                                }
-                            )
+                            vals = rec._prepare_tier_review_vals(td, sequence)
+                            created_trs += tr_obj.create(vals)
```
pedrobaeza: Please take the occasion to accumulate the dictionaries and create all the records at once.

base_tier_validation/models/tier_validation.py
```diff
@@ -469,18 +478,12 @@ def request_validation(self):
                         order="sequence desc",
                     )
                     sequence = 0
+                    vals = []
```
pedrobaeza: For following naming conventions:

```suggestion
                    vals_list = []
```

pedrobaeza: And declare it outside the loop and do only one create.

pedrobaeza: This last comment is unattended.

victoralmau: It's changed

pedrobaeza: Outside the other loop, and you can join the 2 `if` lines into one.

base_tier_validation/models/tier_validation.py
```diff
@@ -472,16 +482,11 @@ def request_validation(self):
...
-                                    "res_id": rec.id,
-                                    "definition_id": td.id,
-                                    "sequence": sequence,
-                                    "requested_by": self.env.uid,
-                                }
+                            vals_list.append(
+                                rec._prepare_tier_review_vals(td, sequence)
                             )
                     self._update_counter({"review_created": True})
+        created_trs += tr_obj.create(vals_list)
```
pedrobaeza: No need now of pre-setting the variable:

```suggestion
        created_trs = tr_obj.create(vals_list)
```

(and remove its initialization).

And as said, you can also reduce one indentation level joining the 2 if.

pedrobaeza (approved):
Thanks for attending all the nitpickings.

LoisRForgeFlow (approved):
Thanks!

/ocabot merge patch

PR #471, opened 2022-03-15
[15.0][MIG] base_user_locale

ChrisOForgeFlow:
Standard Migration
@ForgeFlow

PR #780, opened 2023-12-08
[16.0][MIG] base_user_locale

JordiBForgeFlow:
Standard migration, using https://github.com/OCA/server-ux/pull/471 as a reference

How was tested:

1. Make sure that you test with incognito mode in your browser, because it seems that the locale settings are stored in the browser's cache

2. Introduce the new locale settings in the config:
Using English (US) as base language, I override the locale into EU standard format:

image

3. Test
When creating a sales order I can see that the new format is used:

image


JordiBForgeFlow:
@LoisRForgeFlow @JordiMForgeFlow FYI

JordiMForgeFlow (approved):
functional review 👍🏼

LoisRForgeFlow (approved):
Tested locally, LGTM

PR #782, opened 2023-12-12
[ADD] base_action_visibility_restriction_domain

ilyasProgrammer:
Extends idea of **base_action_visibility_restriction**. Instead of excluded groups here we have special model - action restriction. Using restriction records you can create domains for different groups. Domain will allow/deny action run on sepcific target model records.

If all restrictions of an action are without domain - hide action based on group (**base_action_visibility_restriction** behaviour).
If at least one with domain - action is shown. Access will be defined once user run action.
If restriction record without domain - user can't run action.  If with domain - domain defines for which records user (users group) can run action.


base_action_visibility_restriction_domain/models/ir_actions.py
```diff
@@ -0,0 +1,165 @@
...
+
+    restriction_ids = fields.One2many("ir.actions.restriction", "server_action_id")
+
+    def run(self):
+        """
+        Prevent action running based on groups and domain
+        """
+        for action in self.sudo():
+            restrictions = action.restriction_ids.filtered(
+                lambda x: x.group_id.id in self.env.user.groups_id.ids
```
HekkiMelody: **suggestion**: like in `get_bindings`, we can move `self.env.user.groups_id.ids` before the `for` loop so it's only accessed once.

Also in this case, we could keep the recordset objects instead of the ids, but that's just preference.
`lambda x: x.group_id in self.env.user.groups_id`

base_action_visibility_restriction_domain/models/ir_actions.py
```diff
@@ -0,0 +1,165 @@
...
+                    raise AccessError(
+                        _(
+                            "You don't have enough access rights to run "
+                            "this action on records: %s." % blocked_records
+                        )
+                    )
+                else:
+                    raise AccessError(
+                        _("You don't have enough access rights to run this action.")
+                    )
```
HekkiMelody: **suggestion (non-blocking)**: this could be slightly more readable. Since you set `deny` and immediately `break` to check `deny`, we could do without `deny` entirely IMO

```suggestion
            rec_ids = self.env.context.get("active_id") or self.env.context.get(
                "active_ids"
            )
            records = self.env[self.model_name].browse(rec_ids)
            for restr in restrictions:
                if restr.condition_domain in ["[]", False, ""]:
                    # no domain - block
                    raise AccessError(
                        _("You don't have enough access rights to run this action.")
                    )
                else:
                    # domain allows to run action
                    filtered_records = records.filtered_domain(
                        safe_eval(restr.condition_domain)
                    )
                    blocked_records = records - filtered_records
                    if blocked_records:
                        raise AccessError(
                            _(
                                "You don't have enough access rights to run "
                                "this action on records: %s." % blocked_records
                            )
                        )
```
At that point you could even drop the `else` clause entirely.

HekkiMelody (commented):
Code review, looks good, just some minor suggestions for cleaner code.

With how it's implemented, I agree that the dependency on `base_action_visibility_restriction` feels unnecessary.

This could either be improvement to that module, with the proper data migration, or a new module that's incompatible with `base_action_visibility_restriction`. I think the former option would be my preference.

Also tests are missing.

ilyasProgrammer:
@aleuffre Thanks for the review. Updated according to your suggestions.
Would be good to get feedback from @pedrobaeza or repo maintainers regarding dependency and incompatibility, since, as far as i know, incompatible modules considered as bad practice.

pedrobaeza:
Why is it incompatible? Indeed by default modules shouldn't be incompatible. Or you incorporate a method to inhibit the behavior, or do glue modules for avoiding the incompatibility.

HekkiMelody (approved):
Code review, ok

I still have some doubts about this being its own module rather than a direct improvement to `base_action_visibility_restriction` since half the code is about migration to and from the base module, hiding views from that module and replacing them etc. 

But approving nonetheless.

francesco-ooops (approved):
Functional ok!

francesco-ooops:
@pedrobaeza can we merge this one?

pedrobaeza:
Merging blindly:

/ocabot merge nobump

PR #785, opened 2023-12-20
[16.0] [ADD] base_archive_security

ivantodorovich:
Usage
=====

If you want to restrict the ability to archive records to only certain users you can do so by removing the new group to the users.


chienandalu (approved):
Tested :+1:

ivantodorovich:
Thanks for the review, @chienandalu 

I've just pushed the following changes:

https://github.com/OCA/server-ux/compare/1d2fcf35d60386886cc916e4f75f0ceb48de1f16..3d943c90e412538ff67ce16a28a5545d76100307

- post_init_hook: consider also archived users
- security group: make the addition to the default user in a noupdate tag, so that it can be disabled

jcoux (approved):

chienandalu (approved):

PR #692, opened 2023-07-12
[16.0][MIG] base_tier_validation_server_action

hbrunn:
this is based on not yet merged https://github.com/OCA/server-ux/pull/584 so better review that first

PR #788, opened 2023-12-28
[16.0][MIG] base_tier_validation_server_action

Saran440:
This module continue from https://github.com/OCA/server-ux/pull/692
and fix bug can't use server action rejected with button

To test the changes with the 'purchase_order_tier_validation' module:

1. Create a tier validation and add a server action for rejection using 'button_cancel()'.
2. Proceed to the purchase order and submit it for tier validation.
3. Upon rejection, an error occurred previously as the record would be deleted.

This PR resolves this issue, allowing for rejection while ensuring the document status changes to 'Cancelled' instead of resulting in an error.

cc @hbrunn


hbrunn (approved):
thanks, this looks much cleaner. @bosd if you approve too, we can merge this

bosd (changes requested):
Can you please attend to these errors/warnings?
image

Saran440:
@bosd This warning is correct because we use logger.warning in code. Auto validate will not working when reviewer more than 1

Selection_006

bosd:
Thanks for the explanation.
How about the error "missing 1 positional argument"?

Saran440:
error "missing 1 positional argument" for cover test script only.

bosd (approved):

PR #789, opened 2023-12-29
[14.0][ADD] base_module_default_filter

HekkiMelody:
Simple module to remove the (often useless) default filter on the apps view.


francesco-ooops (approved):
Functional ok!

@pedrobaeza good for you? :)

pedrobaeza:
Well, this can be done easily putting a favorite filter as default, so not sure about the need of this module.

francesco-ooops:
@pedrobaeza how would you suggest to proceed in order to have such a filter automatically enabled for any db ever created?

pedrobaeza:
Ansible or other automation tool. The effort to install the module is the same than setting the filter IMO (and both are scriptable through the same automation tool), so I would go with the one not adding code to maintain.

francesco-ooops:
Ok, so far in our environments every new db has a set of modules installed by default so we thought of adding this one to the flow, but we'll check out how to do this with Ansible. Thanks!

PR #790, opened 2024-01-02
[MIG] date_range: Migrate to version 17.0

sonhd91:
Based on: 
* https://github.com/ramiadavid/server-ux/tree/17.0-mig-date_range

Note:
* fixed issue `active` fields feature doesn't worked on `date.range.types` (doesn't hide it on search view when `active` set `False`)


gurneyalex (approved):

gurneyalex (approved):

pedrobaeza:
Wrong readme fragment defining 2 times an image with the same name it seems. Not sure if something of the automatic conversion or copy/paste thing.

sonhd91:
> Wrong readme fragment defining 2 times an image with the same name it seems. Not sure if something of the automatic conversion or copy/paste thing.

I just updated image name to avoid duplicate subname error

PR #792, opened 2024-01-02
[FIX] base_archive_security: skip check on empty recordset Also, allow to bypass through a special context key

RaoufJabriCamptocamp:
This error disappeared with the uninstallation of the module base_archive_security.

FIX : add condition to toggle_active method to check if user is public user or not.

### DIFF
--- a/base_archive_security/models/base.py
+++ b/base_archive_security/models/base.py
@@ -9,8 +9,14 @@ class Base(models.AbstractModel):
     _inherit = "base"
 
     def toggle_active(self):
-        # check if the user is in the group that can archive/unarchive the record
-        if not self.env.user.has_group("base_archive_security.group_can_archive"):
+        # skip check if special bypass context key exist in env
+        if self.env.context.get("bypass_archive_security"):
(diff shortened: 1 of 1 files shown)


base_archive_security/models/base.py
```diff
@@ -9,8 +9,13 @@ class Base(models.AbstractModel):
     _inherit = "base"
 
     def toggle_active(self):
+        # check if user is public user or not.
```
ivantodorovich:
```suggestion
```

ivantodorovich (changes requested):
Could you change the commit message to something that makes more sense in this context?

ie:

```
[FIX] base_archive_security: skip check on empty recordset

Also, allow to bypass through a special context key
```

RaoufJabriCamptocamp:
@ivantodorovich Done

ivantodorovich (approved):

ivantodorovich (approved):
Thanks @RaoufJabriCamptocamp !

/ocabot merge patch

PR #793, opened 2024-01-03
[16.0][FIX] base_tier_validation: error on loading view from m2o field

RLeeOSI:
Fixes an issue when loading the form view of any model that inherits tier.validation from a m2o field (Create and Edit...). Computed fields must be loaded in before the OWL component renders the form view. 

Previous behavior: 
- Load the form view of a model that inherits tier.validation from a m2o field using "Create and Edit..." option
- Get `Missing field string information for the field '${fieldName}' from the '${resModel}' model` error from processArch()
- Caused by fields missing from "models" key returned by get_view()
- Previous code ignored fields needed if the model was already present in the "models" key

Changes:
- No longer ignores necessary fields if the model is already present in the "model" key

Possible future improvements:
- Reduce repeated code
- Add fields after adding the validation buttons node

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -577,16 +577,21 @@ class TierValidation(models.AbstractModel):
                 new_node = etree.fromstring(new_arch)
                 for new_element in new_node:
                     node.addprevious(new_element)
+                for model in new_models:
+                    if model in all_models:
+                        all_models[model] = set(all_models[model])
+                        all_models[model] = all_models[model].union(new_models[model])
+                    else:
+                        all_models[model] = new_models[model]
                 # _add_tier_validation_reviews process
                 new_node = self._add_tier_validation_reviews(node, params)
                 new_arch, new_models = View.postprocess_and_fields(new_node, self._name)
                 for model in new_models:
                     if model in all_models:
-                        continue
-                    if model not in res["models"]:
-                        all_models[model] = new_models[model]
+                        all_models[model] = set(all_models[model])
+                        all_models[model] = all_models[model].union(new_models[model])
                     else:
-                        all_models[model] = res["models"][model]
+                        all_models[model] = new_models[model]
                 new_node = etree.fromstring(new_arch)
                 node.append(new_node)
             res["arch"] = etree.tostring(doc)


LoisRForgeFlow (approved):
Thanks for the fix, tested and LGTM :+1:

bosd:
I've noticed the regression.
Also from v16 the translation of the slection field does not seem to be loaded correctly.
Easiest way to check is even in the `en_Us` language.
The `pending` state should show as `Pending`. (Capital P)

image

LoisRForgeFlow:
> I've noticed the regression. Also from v16 the translation of the slection field does not seem to be loaded correctly. Easiest way to check is even in the `en_Us` language. The `pending` state should show as `Pending`. (Capital P)
> 
> image

Is this a comment for this PR, or something to be improved in general? In other words, could you clarify if you are approving or rejecting this specific PR.

Thanks!

bosd:
> Is this a comment for this PR, or something to be improved in general? In other words, could you clarify if you are approving or rejecting this specific PR.

Sorry, The mistake was on my end.
I had the assumption this PR was supposed to fix the "translation" of the selection field.
Which it does not. So, my comment is still valid and a separate issue should be created for it.
(It is occurs in V16 + V17.)

I'm neutral on this PR. I did not encouter the problem, that this PR is supposed to fix.

LoisRForgeFlow:
> > Is this a comment for this PR, or something to be improved in general? In other words, could you clarify if you are approving or rejecting this specific PR.
> 
> Sorry, The mistake was on my end. I had the assumption this PR was supposed to fix the "translation" of the selection field. Which it does not. So, my comment is still valid and a separate issue should be created for it. (It is occurs in V16 + V17.)
> 
> I'm neutral on this PR. I did not encouter the problem, that this PR is supposed to fix.

Ok, thanks for clarifying. You could create the issues then if you think they are relevant.


Regarding this PR, let's wait one more review.

houzefa-abba:
Hi I'm only noticing this PR now, I did #846 to fix this same part of the code, for another issue (form-embedded lists). Apparently according to a comment there, this also fixes workflow kinks.

I'm not closing #846 as duplicate though as it handles the "Reduce repeated code" you mention & adds a unit test.

I'll let you decide how we merge these (either only #846, or #846 rebased on top of this PR, to have both commits).

PR #795, opened 2024-01-04
[16.0][FIX] Fix tier validation error when not changing state

agyamuta:
(cherry picked from commit 887a9ba6654ea3c67b12934396d3aea70faf9b92)


LoisRForgeFlow (changes requested):
I think we have already discussed this in other PRs, that is incorrect, it would allow to write any other field when a document in under validation.

PR #787, opened 2023-12-27
[17.0][MIG] base_tier_validation: Migration to 17.0

celm1990:
superseded by https://github.com/OCA/server-ux/pull/755

PR #799, opened 2024-01-10
[MIG] base_tier_validation_formula: Migrate to version 17.0

sonhd91:
Ref: BSRD-706

Depended on:
* https://github.com/OCA/server-ux/pull/787


base_tier_validation_formula/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2019 Creu Blanca
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Base Tier Validation Formula",
+    "summary": """
+        Formulas for Base tier validation""",
+    "version": "17.0.1.0.0",
```
bosd, line 8:
```suggestion
    "version": "17.0.1.0.0",
    "category": "Tools",
```

base_tier_validation_formula/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    "license": "AGPL-3",
```
bosd, line 10:
```suggestion
    "license": "AGPL-3",
    "development_status": "Mature",
```

base_tier_validation_formula/tests/test_tier_validation.py
```diff
@@ -0,0 +1,169 @@
...
+                "name": "access.tester",
+                "model_id": cls.tester_model.id,
+                "perm_read": 1,
+                "perm_write": 1,
+                "perm_create": 1,
+                "perm_unlink": 1,
+            }
+        )
+
+        # Create users:
```
bosd, line 46: Is there a reason new users are created here?
(It comes at a cost, can the users from the base module be re-used?)

bosd (commented):
Thanks for this pr! Looks good, Some non-blocking comments.

base_tier_validation_formula/__manifest__.py
bosd: Promotion should'nt have to be in migration commit. But I noticed the missing tag, and it likely satisfies mature status.

base_tier_validation_formula/tests/test_tier_validation.py
sonhd91: Yeah, that will do, but we still need create one more user for tests case

sonhd91:
Branch updated, please check again

Since `base_tier_validation` was merged, I uncomment some related code for this module

gurneyalex (approved):

PR #800, opened 2024-01-10
[17.0][IMP] base_tier_validation: Merge module with base_tier_validation_waiting

bosd:
As proposed as an idea in:
https://github.com/OCA/server-ux/pull/787#issuecomment-1871461565
This is the PR for the migration of base_tier_validation_waiting into the base module `base_tier_validation_waiting`.

Please fasstrack, as v17 is not deployed yet in production, this is the best time to introduce this refactor.

To the contributors/authers/reviewers of the migration pr..

@celm1990, @nguyenminhchien, @bofiltd, @moitabenfdz, @sonhd91, @Dranyel-Bosd, @BT-jcolmeiro
Can you please review?

### DIFF
--- a/base_tier_validation/models/tier_review.py
+++ b/base_tier_validation/models/tier_review.py
@@ -13,12 +13,13 @@ class TierReview(models.Model):
 
     name = fields.Char(related="definition_id.name", readonly=True)
     status = fields.Selection(
-        selection=[
+        [
+            ("waiting", "Waiting"),
             ("pending", "Pending"),
             ("rejected", "Rejected"),
             ("approved", "Approved"),
         ],
-        default="pending",
+        default="waiting",
     )
     model = fields.Char(string="Related Document Model", index=True)
     res_id = fields.Integer(string="Related Document ID", index=True)
@@ -77,11 +78,26 @@ class TierReview(models.Model):
 
     @api.depends("definition_id.approve_sequence")
     def _compute_can_review(self):
+        reviews = self.filtered(lambda rev: rev.status in ["waiting", "pending"])
+        if reviews:
+            # get minimum sequence of all to prevent jumps
+            next_seq = min(reviews.mapped("sequence"))
+            for record in reviews:
+                # if approve by sequence, check sequence has been reached
+                if record.approve_sequence:
+                    if record.sequence == next_seq:
+                        record.status = "pending"
+                # if there is no approval sequence go directly to pending state
+                elif not record.approve_sequence:
+                    record.status = "pending"
+                if record.status == "pending":
+                    if record.definition_id.notify_on_pending:
+                        record._notify_pending_status(record)
         for record in self:
             record.can_review = record._can_review_value()
 
     def _can_review_value(self):
-        if self.status != "pending":
+        if self.status not in ("pending", "waiting"):
             return False
         if not self.approve_sequence:
             return True
@@ -126,3 +142,8 @@ class TierReview(models.Model):
             if not reviewer_field or not reviewer_field._name == "res.users":
                 raise ValidationError(_("There are no res.users in the selected field"))
         return reviewer_field
+
+    def _notify_pending_status(self, review_ids):
+        """Method to call and reuse abstract notification method"""
+        resource = self.env[self.model].browse(self.res_id)
+        resource._notify_review_available(review_ids)
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -22,8 +22,6 @@ class TierValidation(models.AbstractModel):
     _state_to = ["confirmed"]
     _cancel_state = "cancel"
 
-    # TODO: step by step validation?
-
     review_ids = fields.One2many(
         comodel_name="tier.review",
         inverse_name="res_id",
@@ -47,6 +45,7 @@ class TierValidation(models.AbstractModel):
     validation_status = fields.Selection(
         selection=[
             ("no", "Without validation"),
+            ("waiting", "Waiting"),
             ("pending", "Pending"),
             ("rejected", "Rejected"),
             ("validated", "Validated"),
@@ -70,12 +69,15 @@ class TierValidation(models.AbstractModel):
     def _compute_has_comment(self):
         for rec in self:
             has_comment = rec.review_ids.filtered(
-                lambda r: r.status == "pending" and (self.env.user in r.reviewer_ids)
+                lambda r: r.status in ("waiting", "pending")
+                and (self.env.user in r.reviewer_ids)
             ).mapped("has_comment")
             rec.has_comment = True in has_comment
 
     def _get_sequences_to_approve(self, user):
-        all_reviews = self.review_ids.filtered(lambda r: r.status == "pending")
+        all_reviews = self.review_ids.filtered(
+            lambda r: r.status in ("waiting", "pending")
+        )
         my_reviews = all_reviews.filtered(lambda r: user in r.reviewer_ids)
         # Include all my_reviews with approve_sequence = False
         sequences = my_reviews.filtered(lambda r: not r.approve_sequence).mapped(
@@ -98,7 +100,7 @@ class TierValidation(models.AbstractModel):
     def _search_can_review(self, operator, value):
         domain = [
             ("review_ids.reviewer_ids", "=", self.env.user.id),
-            ("review_ids.status", "=", "pending"),
+            ("review_ids.status", "in", ["pending", "waiting"]),
             ("review_ids.can_review", "=", True),
             ("rejected", "=", False),
         ]
@@ -111,7 +113,7 @@ class TierValidation(models.AbstractModel):
     def _compute_reviewer_ids(self):
         for rec in self:
             rec.reviewer_ids = rec.review_ids.filtered(
-                lambda r: r.status == "pending"
+                lambda r: r.status in ("waiting", "pending")
             ).mapped("reviewer_ids")
 
     @api.model
@@ -145,7 +147,6 @@ class TierValidation(models.AbstractModel):
                 ("model", "=", self._name),
                 ("reviewer_ids", operator, value),
                 ("can_review", "=", True),
-                ("status", "=", "pending"),
             ]
         )
         return [("id", model_operator, list(set(reviews.mapped("res_id"))))]
@@ -192,6 +193,12 @@ class TierValidation(models.AbstractModel):
                 and any(item.review_ids.filtered(lambda x: x.status == "pending"))
             ):
                 item.validation_status = "pending"
+            elif (
+                not item.validated
+                and not item.rejected
+                and any(item.review_ids.filtered(lambda x: x.status == "waiting"))
+            ):
+                item.validation_status = "waiting"
             else:
                 item.validation_status = "no"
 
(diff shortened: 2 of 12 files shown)


bosd:
Some background info...

Since sequential validations are supported in the base module. It makes sense from an user perspective to only sent out notifications to the validator if it is thier turn in the sequence to review.
Merging of these two modules accomplish that.

Before this pr (and without the waiting module) it was only possible to sent out and notification when a record was created. A reviewer in the end of the sequence got the e-mail notification before it's turn.
If the user decided to go to the document and make an attempt to do so. He was unable to do it.. Because it was not their turn yet..
Since there was no capability to get a notification. The user- had to visit the record now and then to see if it was their turn to validate.
This was a poor flow from usabulity point of view.
Easily fixed by the waiting module.

The merge of into the base module. Should make it more clear to users/functional consultants. As well as easier to maintain.

bofiltd (approved):
Makes sense.
Functional test locally. LGTM :champagne:

base_tier_validation/models/tier_review.py
```diff
@@ -77,11 +78,26 @@ def _compute_reviewed_formated_date(self):
...
+                # if approve by sequence, check sequence has been reached
+                if record.approve_sequence:
+                    if record.sequence == next_seq:
+                        record.status = "pending"
+                # if there is no approval sequence go directly to pending state
+                elif not record.approve_sequence:
+                    record.status = "pending"
+                if record.status == "pending":
+                    if record.definition_id.notify_on_pending:
+                        record._notify_pending_status(record)
```
celm1990, line 95: I'm uncertain whether it's advisable to update fields within a `computed function` if those fields are not explicitly defined as computed (field `status`).

base_tier_validation/models/tier_validation.py
```diff
@@ -70,12 +69,15 @@ class TierValidation(models.AbstractModel):
     def _compute_has_comment(self):
         for rec in self:
             has_comment = rec.review_ids.filtered(
-                lambda r: r.status == "pending" and (self.env.user in r.reviewer_ids)
+                lambda r: r.status in ("waiting", "pending")
+                and (self.env.user in r.reviewer_ids)
```
celm1990:
```suggestion
                and self.env.user in r.reviewer_ids
```

base_tier_validation/models/tier_validation.py
```diff
@@ -320,6 +327,20 @@ def _check_state_conditions(self, vals):
     def _validate_tier(self, tiers=False):
         self.ensure_one()
         tier_reviews = tiers or self.review_ids
+        waiting_reviews = tier_reviews.filtered(
+            lambda r: r.status == "waiting"
+            or r.approve_sequence_bypass
+            and (self.env.user in r.reviewer_ids)
+        )
```
celm1990: Are parentheses valid?
```suggestion
        waiting_reviews = tier_reviews.filtered(
            lambda r: r.status == "waiting"
            or r.approve_sequence_bypass
            and self.env.user in r.reviewer_ids
        )
```

base_tier_validation/models/tier_validation.py
```diff
@@ -427,7 +448,8 @@ def _rejected_tier(self, tiers=False):
         self.ensure_one()
         tier_reviews = tiers or self.review_ids
         user_reviews = tier_reviews.filtered(
-            lambda r: r.status == "pending" and (self.env.user in r.reviewer_ids)
+            lambda r: r.status in ("waiting", "pending")
+            and (self.env.user in r.reviewer_ids)
```
celm1990:
```suggestion
            and self.env.user in r.reviewer_ids
```

base_tier_validation/models/tier_validation.py
```diff
@@ -594,3 +624,22 @@ def get_view(self, view_id=None, view_type="form", **options):
...
+                    and r.res_id == x.id
+                ).mapped("reviewer_ids")
+                # Subscribe reviewers and notify
+                getattr(rec, subscribe)(
+                    partner_ids=users_to_notify.mapped("partner_id").ids
+                )
+                getattr(rec, post)(
+                    subtype_xmlid=self._get_requested_notification_subtype(),
+                    body=rec._notify_requested_review_body(),
+                )
```
celm1990: If at this point in the code, it is certain that this method exists because it is not called directly?, it enhances readability.
```suggestion
                rec.message_subscribe(
                    partner_ids=users_to_notify.mapped("partner_id").ids
                )
                rec.message_post(
                    subtype_xmlid=self._get_requested_notification_subtype(),
                    body=rec._notify_requested_review_body(),
                )
```

celm1990 (commented):
first code review; I will test it for functionality.

Dranyel-Bosd (approved):
Functional test locally, LGTM 🍾

celm1990:
I tested these changes, but I have some doubts. For example, I have three reviews for three specific users:

R1 -> Admin
R2 -> U2
R3 -> U3
image


**Before this commit:**
When Admin approved R1, an email is sent to U2 and U3 notifying them that R1 is approved.

**After this commit:**
When Admin approves R1, an email is sent to both U2 and U3, notifying them that R1 is approved(as previous). Additionally, a new email is sent to U2 and **U3** informing them that a new review (R2) has been created. However, for the purpose of this changes, if the intention is to notify only the next user (**U2**), I hope that only **U2** is notified and that no email is sent to **U3**. This is because if U3 receives an email, they might attempt to review and approve, causing confusion as it is not their turn yet. we would revert back to the initial problem.

image

Please let me know if I understood the functionality of the module correctly.

bosd:
> Additionally, a new email is sent to U2 and **U3** informing them that a new review (R2) has been created. However, for the purpose of this changes, if the intention is to notify only the next user (**U2**), I hope that only **U2** is notified and that no email is sent to **U3**. This is because if U3 receives an email, they might attempt to review and approve, causing confusion as it is not their turn yet. we would revert back to the initial problem.

Yes, I think you understand the functionality correctly.
The main problem before, was that U3 got an invitation email to review. 
But he coul'dnt review because it was not his turn yet.
Before this PR, he was unable to get a notification when it actually was his turn.

Another use case is that U3 could be upper management. And only wants to be bothered with messges which made it past the previous tiers. 

I'm under the impression you have the functionality and use case correct.
Yet, I'm unable to determine from your feedback if the module passed the functional test.
Dunno if it is the case, but the notification settings might be a bit "misleading"/unclear.
In a sequential flow, I would reccomend to only check the `Notify Reviewers on reaching Pending` box.
Ignoring all the other notifications, this should happen.


```mermaid
flowchart TD
    A[U0 validation request] -->|Notification message: request to review sent to U1| B[U1 Approves SO]
    B -->|Notification message: request to review sent to  U2| C[U2 approves SO]
    C -->|Notification message: request to review sent to U3| D[U3 approves SO]
    D--> E(The End\n Record = validated)
```

bosd:
While out of scope of this pr, we might want to look into who is receiving the "approved messages".
To have it go to the original document owner makes sense. In case U3 is upper management, we don't want to bother him with all the back and forth happening in the tiers before.

celm1990 (approved):
Thanks for clarifying. I unchecked "Notify on Creation" as per your suggestion to reduce emails. However, I noticed that other users are still being notified when reviews are requested. I suspect this is happening because of followers. So, LGTM .

image

By the way, in the other PR(Improve UI), you can add these fields as `optional="show"`  😅

NOTE: Please let me know your comments regarding my comment on computed fields not being defined as computed explicitly. Could this be causing cache inconsistency?

base_tier_validation/models/tier_review.py
bosd: >NOTE: Please let me know your comments regarding my comment on computed fields not being defined as computed explicitly. Could this be causing cache inconsistency?

tbh, I'm a bit overasked by this.. 
Maybe it is not the ideal situation. (yet, I'm unaware of any alternative method to implement this).
Well, we've deployed this code succesfully in production. W/o any problems. 
Judging from the experience it is okay..
But happy to learn from the more experienced dev's.

bosd:
Oopsie, I answered on the comment above, but it does'nt show up down here.
If more improvements are suggested I'm happy to include them.
But as other mirgrations depend on this base module, ideally we would move forward.

@OCA/tools-maintainers I think this one is ready for merge..

celm1990:
@LoisRForgeFlow @pedrobaeza, could you take a look at this PR? What do you think, is it possible to merge?

bosd:
@kittiu Can you please review as well?

bosd:
@gurneyalex Can you take a look as well? 🙏

bosd:
@oca/edi-maintainers Can we please fasttrack this one? :pray:

bosd:
Oops, used the wrong handle to fasttrack this one.. [@OCA/server-ux-maintainers](https://github.com/orgs/OCA/teams/server-ux-maintainers) seems to be the correct one?

Or @rousseldenis Do you have super powers here as well?

rousseldenis:
> Oops, used the wrong handle to fasttrack this one.. [@OCA/server-ux-maintainers](https://github.com/orgs/OCA/teams/server-ux-maintainers) seems to be the correct one?
> 
> Or @rousseldenis Do you have super powers here as well?

Nope.

LoisRForgeFlow (commented):
Functional review, in general I think this is a great proposal, thanks! Some comments below:

1. This field seems to be a bit misplaced:

image

2. Time to add the sequence handler widget?

image

3. There are multiple notifications with the same message. This one is not a blocking comment as I know this was an issue before this change too, but with this sequencial approval it becomes more noisy.

image

bosd:
> Functional review, in general I think this is a great proposal, thanks! Some comments below:
> 
> 1. This field seems to be a bit misplaced:
> 
> image
>

The UI Improvements are made in follow up pr.
https://github.com/OCA/server-ux/pull/804


> 2. Time to add the sequence handler widget?
> 
> image
> 
Great Suggestion.. 
Will add it, In this PR or the UI PR. In case this one gets merged first.

> 3. There are multiple notifications with the same message. This one is not a blocking comment as I know this was an issue before this change too, but with this sequencial approval it becomes more noisy.
> 
> image

I agree, The messages and the content could be improved.
Have been thinking about rewording the message to.. But decided to not do big chances for now.
Adding the reviewer name to which the review is requested from could be usefull.
But will also lead to confusion by request from a group or server action.


@LoisRForgeFlow Are you ok to merge this one as is?
So the other modules which depend on it can be refactored during the migration.

LoisRForgeFlow:
@bosd Yes, I think we can merge as is, and continue the improvements on followup PRs. I don't want to hold this because of my lack of time to review

base_tier_validation/models/tier_review.py
```diff
@@ -77,11 +78,26 @@ def _compute_reviewed_formated_date(self):
 
     @api.depends("definition_id.approve_sequence")
     def _compute_can_review(self):
```
StefanRijnhart, line 80: @bosd Can you help me understand `can_review` vs. `waiting` and `pending`? Is it always the case that `can_review` is true if and only if its status is `pending`? Would it be safe for me to propose a simplification based on this? Or am I missing something?

bosd: Oof, thats been a while.
Doing this without code check, out of memory:
can_review_ everyone who can review the record.
Waiting = when it is not yet the users turn to approve
Pending = his turn to approve in sequential flow?

StefanRijnhart: Thanks for checking in on this! It seems that can_review on tier.validation is indeed user aware, but can_review on tier.review is not. That's why I'm thinking it *is* simply the case that can_review equals 'pending' so I think the related code can be simplified, or maybe `can_review` on the tier.review level can be dropped entirely.

Issue #809, opened 2024-01-24
[16.0] [base_tier_validation] Incorrect timezone for Validation Date in Tier Review


evanomnisoft:
Validation Date for tier reviews shows in UTC instead of user timezone.

This seems to do the trick:
~~requested_date_utc = pytz.timezone(timezone).localize(review.reviewed_date)~~
requested_date_utc = pytz.timezone("UTC").localize(review.reviewed_date)

https://github.com/OCA/server-ux/blob/a16356f3ba6945c0225388db9dfe474b3d9298be/base_tier_validation/models/tier_review.py#L74

evanomnisoft:
#810

PR #810, opened 2024-01-24
[16.0][FIX][base_tier_validation] fix validation date timezone

evanomnisoft:
Validation Date for tier reviews shows in UTC instead of user timezone.

requested_date_utc should be localized to UTC instead of user timezone

### DIFF
--- a/base_tier_validation/models/tier_review.py
+++ b/base_tier_validation/models/tier_review.py
@@ -71,7 +71,7 @@ class TierReview(models.Model):
             if not review.reviewed_date:
                 review.reviewed_formated_date = False
                 continue
-            requested_date_utc = pytz.timezone(timezone).localize(review.reviewed_date)
+            requested_date_utc = pytz.timezone("UTC").localize(review.reviewed_date)
             requested_date = requested_date_utc.astimezone(pytz.timezone(timezone))
             review.reviewed_formated_date = requested_date.replace(tzinfo=None)
 
--- a/base_tier_validation/models/tier_review.py
+++ b/base_tier_validation/models/tier_review.py
@@ -71,9 +71,9 @@ class TierReview(models.Model):
             if not review.reviewed_date:
                 review.reviewed_formated_date = False
                 continue
-            requested_date_utc = pytz.timezone("UTC").localize(review.reviewed_date)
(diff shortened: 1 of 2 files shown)


evanomnisoft:
#809

LoisRForgeFlow (commented):
Hi,

Thanks for contributing. I have a couple of questions/requests:

- Could you provide a step by step guide to reproduce the issue you are trying to fix?
- Also the original line you are changing is a bit weird `requested_date_utc = (...utc stuff...)review.reviewed_date)`  I don't see how the requested date could be filled with the validation date, thoughts?

evanomnisoft:
Hi,

These are the steps to reproduce the issue

1. Setup Odoo and User with a timezone other than UTC (eg. GMT+8)
2. Install tier validation + another support tier validation such as sale_tier_validation
3. Configure at least 1 tier definition
4. Trigger the tier definition by creating a sale order
5. Validate using the validation user set in step 2
6. In the sale order, expand the Review section which shows the list of validators
7. See the validation date (it will be in UTC or GMT+0 no matter what)

You are absolutely right on the variables used though. The field in question here is the datetime when the tier review has happened (field name: reviewed_date)

1. It starts of as reviewed_date (which is essentially UTC but naive) but gets assigned to requested_date_utc then back to review_formatted_date
2. A more accurately named alternative could be something like:

```
reviewed_date_utc = pytz.timezone("UTC").localize(review.reviewed_date)
            
reviewed_date_tz = reviewed_date_utc.astimezone(pytz.timezone(timezone))

review.reviewed_formated_date = reviewed_date_tz.replace(tzinfo=None)
```

Do you think that makes better sense?

LoisRForgeFlow:
@evanomnisoft Thanks for the details, I could confirm now the issue :+1: 

Regarding the names of the variables, yes, I would appreciate if you can do the rename.  It will make the code easier to understand.

evanomnisoft:
@LoisRForgeFlow Done. Thanks!

LoisRForgeFlow (approved):
Thanks :+1:

bosd (approved):

LoisRForgeFlow:
@evanomnisoft could you port this to 17.0?

PR #820, opened 2024-02-06
[14.0][ADD] base_rule_override: new module

HekkiMelody:
New module to add record rules that are in AND with other rules

To do:

- [x] Add tests


francesco-ooops (approved):
Functional ok!

PR #824, opened 2024-02-11
[17.0][RFC] base_tier_validation: remove unused code

bosd:
remove the code, which was needed for the review widget pre V16.

### DIFF
--- a/base_tier_validation/models/res_users.py
+++ b/base_tier_validation/models/res_users.py
@@ -45,19 +45,3 @@ class Users(models.Model):
                         "pending_count": len(records),
                     }
         return list(user_reviews.values())
-
-    @api.model
-    def get_reviews(self, data):
-        review_obj = self.env["tier.review"].with_context(lang=self.env.user.lang)
-        res = review_obj.search_read([("id", "in", data.get("res_ids"))])
-        for r in res:
-            # Get the translated status value.
-            r["display_status"] = dict(
-                review_obj.fields_get("status")["status"]["selection"]
-            ).get(r.get("status"))
-            # Convert to datetime timezone
-            if r["reviewed_date"]:
-                r["reviewed_date"] = fields.Datetime.context_timestamp(
-                    self, r["reviewed_date"]
-                )
-        return res
--- a/base_tier_validation/tests/test_tier_validation.py
+++ b/base_tier_validation/tests/test_tier_validation.py
@@ -417,9 +417,7 @@ class TierTierValidation(CommonTierValidation):
         # Request validation
         review = test_record.with_user(self.test_user_2).request_validation()
         self.assertTrue(review)
-        self.assertTrue(self.test_user_1.get_reviews({"res_ids": review.ids}))
-        self.assertTrue(self.test_user_1.review_ids)
-        test_record.invalidate_model()
+        self.env.invalidate_all()
         self.assertTrue(test_record.review_ids)
         # Used by front-end
         count = self.test_user_1.with_user(self.test_user_1).review_user_count()
@@ -475,8 +473,6 @@ class TierTierValidation(CommonTierValidation):
         self.assertEqual(len(records), 1)
         review = self.test_record.with_user(self.test_user_2.id).request_validation()
         self.assertTrue(review)
-        self.assertTrue(self.test_user_1.get_reviews({"res_ids": review.ids}))
-        self.assertTrue(self.test_user_1.review_ids)
         self.test_record.with_user(self.test_user_1.id).request_validation()
 
     def test_18_test_review_by_res_users_field(self):


bosd:
the function `get_reviews` was no longer called upon in the javascript. After this commit. 
https://github.com/OCA/server-ux/commit/145c5702fc51eaf6622ed38d64ce0dd26b7dc145#diff-5d679fcb388f892b31b8c390404d207ed9b96c806e57db499e13d04443d628e9L33
(or anywhere else afaik)
But it still was unit tested.

The remaining code might need some refactoring :thinking: 

@celm1990 @LoisRForgeFlow 
Can you please review / provide feedback?

base_tier_validation/models/res_users.py
```diff
@@ -15,7 +15,6 @@ def review_user_count(self):
         domain = [
             ("status", "=", "pending"),
             ("can_review", "=", True),
-            ("id", "in", self.env.user.review_ids.ids),
```
LoisRForgeFlow: this

base_tier_validation/models/res_users.py
```diff
@@ -24,7 +23,7 @@ def review_user_count(self):
             if reviews:
                 records = (
                     self.env[model]
-                    .with_user(self.env.user)
+                    .sudo()
```
LoisRForgeFlow: and this

LoisRForgeFlow (commented):
can you also explain the other changes?

base_tier_validation/models/res_users.py
bosd: Had to remove this, to make the tests pass.
I'm not seeing in the code where the values to review_ids are written. 
(expect for the no longer needed lines in the test)
`self.assertTrue(self.test_user_1.get_reviews({"res_ids": review.ids}))`
With no valid ID assigned it broke the tests..

(Filtering on the user ID is done, a couple llines below.)

base_tier_validation/models/res_users.py
bosd: Without this. There was an error in the tests. The user is not allowed to acces SO records.

base_tier_validation/models/res_users.py
bosd: Maybe some more refactoring is needed on the domain. (Or an alternative function to write `review_ids`) I'm not sure..

LoisRForgeFlow:
@bosd The two changes above that you claimed are needed to fix tests, doesn't look needed as many time has passed and tests of tier validation continue to pass. Can you just remove them and just keep the deletion of unused code? After that, it will be good to merge.

Thanks!

bosd:
@LoisRForgeFlow Done, I removed those fixes.

LoisRForgeFlow:
@bosd test are failing now, can you check? I mean there must be something wrong here, main branch is green

bosd:
@LoisRForgeFlow With the earlier mentiond fixes removed tests are indeed failling.
(Some time has passed since I started working on this. While quickly looking at this, I don't know why it fails now.)
Assuming it was something acces rights related and compute order.

LoisRForgeFlow:
@bosd is you branch rebased on top of the latest 17.0?

bosd:
@LoisRForgeFlow Yes it is.

LoisRForgeFlow:
@bosd  hmmm weird indeed. I guess you will need to have a deeper look when you have time, we cannot merge without the green check.

bosd:
@LoisRForgeFlow Ive included the fixes again. Is it ok to merge it as is?
As I don't expect to have some time available in the near future to dig very deep into this and find the reasoning.

base_tier_validation/models/res_users.py
StefanRijnhart: This needs to stay, otherwise the review count will include all users' reviews I fear. The test can be fixed with
```
--- a/base_tier_validation/tests/test_tier_validation.py
+++ b/base_tier_validation/tests/test_tier_validation.py
@@ -417,7 +417,7 @@ class TierTierValidation(CommonTierValidation):
         # Request validation
         review = test_record.with_user(self.test_user_2).request_validation()
         self.assertTrue(review)
-        test_record.invalidate_model()
+        self.env.invalidate_all()
         self.assertTrue(test_record.review_ids)
         # Used by front-end
         count = self.test_user_1.with_user(self.test_user_1).review_user_count()
```

bosd: Thanks, that fixed it!

bosd:
Included the fix, and rebased.

StefanRijnhart (approved):

base_tier_validation/models/res_users.py
celm1990: @bosd I think it is better to use `sudo` or `with_user` in the test. Did you try using these options instead of changing the code to make the test work?

celm1990 (commented):
LGTM but I have the same question as @LoisRForgeFlow regarding `sudo`

base_tier_validation/models/res_users.py
bosd: It was a long time ago I was working on this. I think I just changed it to make the test work.

bosd: Test pass as well now by using `with_user`. So reverted this change.

celm1990 (approved):
LGTM

LoisRForgeFlow (approved):
Thanks

PR #832, opened 2024-02-23
[FIX] base_tier_validation: prevent useless loop

fmdl:


LoisRForgeFlow (changes requested):
Why is it useless? Could you elaborate the problem you are fixing?

fmdl:
Because if you have 100 records, you overwrite 99 times the same value for the same key in user_reviews.

ByteMeAsap (approved):
Looks Good !!!

celm1990 (approved):
LGTM

bosd (approved):

LoisRForgeFlow (approved):
/ocabot merge patch

PR #835, opened 2024-02-29
[FIX] mass_editing : Prevent to crash if the module is installed, but if there are no data

legalsylvain:
For that reason, do not populate mass_editing_line, if mass_field_rel doesn't exist, that occures if we come from the latest version of the module in 12.0

### DIFF
--- a/mass_editing/migrations/13.0.1.0.0/post-migration.py
+++ b/mass_editing/migrations/13.0.1.0.0/post-migration.py
@@ -7,6 +7,10 @@ def migrate(cr, version):
     if not version:
         return
     # Don't execute if already migrated in v12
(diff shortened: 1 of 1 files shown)


mass_editing/migrations/13.0.1.0.0/post-migration.py
```diff
@@ -7,6 +7,10 @@ def migrate(cr, version):
     if not version:
         return
     # Don't execute if already migrated in v12
```
chienandalu, line 9: Doesn't this comment correspond to the other statement?

legalsylvain: This comment is for both

Issue #822, opened 2024-02-08
[16.0+] base_tier_validation: Selection Field not translated.


bosd:
## base_tier_validation
The name of the module that has a bug.

## Describe the bug
The selection field in the review table is no longer translated. It shows the variable name instead of the text value.

![image](https://github.com/OCA/server-ux/assets/11499387/4990cd3c-9f83-4d5a-8087-628bd10fd958)

It should show:
The `pending` state should show as `Pending`. (Capital P)
And thereby being able to be translated.
Screenshot From V15.
![image](https://github.com/OCA/server-ux/assets/11499387/41989eea-70b1-4825-ac59-dc663fd012fc)

## To Reproduce
**Affected versions**: 16, 17

1. Install the base module
2. Install a tier_validation module e.g. sale_tier_validation
3. Create a tier definition
4. Create a document to be validation with a tier validation
5. Look at the tier review table. See that the translation text is not resolved.

**Expected behavior**
The text to be translated as in Pre V16 versions

**Additional context**
As discovered in: https://github.com/OCA/server-ux/pull/793#issuecomment-1883808701

bosd:
cc [LoisRForgeFlow](https://github.com/LoisRForgeFlow)

bosd:
@bizzappdev Can you take a look at this? :pleading_face:

bizzappdev:
@bosd Fix is at #830

PR #839, opened 2024-03-09
[FIX][17.0]base_tier_validation: Selection Field translated

bizzappdev:
Fix: https://github.com/OCA/server-ux/issues/822 for v17


LoisRForgeFlow (approved):
Thanks!

/ocabot merge minor

PR #840, opened 2024-03-12
[15.0][ADD] web_filter_header_button: New module

chienandalu:
This module allows to add some selected filters as buttons in the header control panel.

Peek 12-03-2024 13-50

cc @Tecnativa TT47127

@pedrobaeza do you think the name is ok?


pedrobaeza:
It should be called web_filter_button IMO

pedrobaeza:
The commit message still indicates `filter_button`. Thinking twice, `web_filter_header_button` may be more specific.

pedrobaeza:
I think this should be in OCA/web.

chienandalu:
What are the nuances for a module to be in  `server-ux`?

pedrobaeza:
Well, this is a modification of the web client, so we usually put them in oca/web.

chienandalu:
Moved to https://github.com/OCA/web/pull/2768

PR #841, opened 2024-03-12
[DOC] date_range: Update broken image links

amh-mw:


lmignon (changes requested):
Hi @amh-mw 

Thank you for your contrib. The `README.rst` file is generated from fragments defined into several files located under the `readme` directory. The right fix is to modify the links into these files. Once done, the `README.rst` will be generated at merge time. Can you update your PR accordingly?

francesco-ooops:
@amh-mw you can take this as an example: https://github.com/OCA/server-ux/blob/16.0/template_content_swapper/readme/USAGE.rst?plain=1#L3

lmignon (approved):

francesco-ooops:
@lmignon I think it's better to use relative paths

date_range/readme/CONFIGURE.rst
```diff
@@ -12,7 +12,7 @@ inherit from `date.range.search.mixin`:
 
 This will make a `Period` field show up in the search view:
 
-  .. figure:: https://raw.githubusercontent.com/OCA/server-tools/12.0/date_range/static/description/date_range_many2one_search_field.png
+  .. figure:: https://raw.githubusercontent.com/OCA/server-ux/15.0/date_range/static/description/date_range_many2one_search_field.png
```
pedrobaeza: The correct way to put it and forget about this on each version is to put a relative path, and let the README generator to put the absolute one, but meanwhile, you see the correct image in local devel env.

```suggestion
  .. figure:: ../static/description/date_range_many2one_search_field.png
```

lmignon: @pedrobaeza you're right. I just checked the readme generator and relative paths are replaced by the right absolute path at generation.

lmignon (changes requested):
As stated by others reviewers, the path must be relative

francesco-ooops (approved):
LGTM

@amh-mw would you be so kind to open a PR for v16 as well? Thanks!

lmignon (approved):

PR #844, opened 2024-03-14
[16.0][DOC] date_range: Update broken image links.

lmignon:
forward port of #841


lmignon:
fast tracking since #841 is approved and merged and it's a minor change improving the generated documentation.

PR #849, opened 2024-03-30
[16.0][ADD] base_live_update

hbrunn:
This module provides a framework other modules can use to update the
client side UI when records of some model have changed.

To test this on runboat, allow notifications, go to a kanban view of partners, in another window/computer/user, change the name of some partner, and see this change reflected in the kanban view. Or group the view ie by company, and drag some partner into another company.

Or install whatever module you're interested in and add the model in question to Settings/Technical/Live update notifications.


base_live_update/views/live_update_notification.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2024 Hunki Enterprises BV
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0) -->
+<data>
+    <record id="view_live_update_notificatoin_tree" model="ir.ui.view">
```
StefanRijnhart:
```suggestion
    <record id="view_live_update_notification_tree" model="ir.ui.view">
```

StefanRijnhart:
Workes very nicely already on the partner kanban view! For my purposes the kanban card moving between groups is the killer purpose here. Maybe that bit does not need to be in a separate module?

hbrunn:
indeed, I'm more and more of a mind to do all of that in the base module, as we depend on web already anyways

hbrunn:
@StefanRijnhart killer purpose implemented. ~~Very inefficiently (see TODOs), which defeats the purpose of this module to some degree, but I think this is good enough for version 1 and can be iterated on.~~

antoniocanovas (approved):
Functional review OK !!

PR #853, opened 2024-04-11
[ADD] [15.0] base_warn_option

Shide:
This module extends the functionality of blocking messages to support more consistency
between messages and to allow you to set a blocking message quickly.

This is a PR of a series that depends on this one. Next PRs will add these modules on its respective repo:

[sale_warn_option](https://github.com/OCA/sale-workflow/pull/3076)
[purchase_warn_option](https://github.com/OCA/purchase-workflow/pull/2240)
[account_warn_option](https://github.com/OCA/account-invoicing/pull/1704)
[stock_warn_option](https://github.com/OCA/stock-logistics-workflow/pull/1571)

MT-5193 @moduon @rafaelbn @yajo @EmilioPascual please review if you want :)


base_warn_option/readme/CONTEXT.md
```diff
@@ -0,0 +1,12 @@
+This module has been developed to maintain consistency in the messages used when
+blocking some models.
+
+This module doesn't do anything by itself.
```
yajo, line 4: At least, it adds some models, views and permissions. You could include configuration instructions, so you don't need to repeat them on each of the upcoming submodules.

Shide: Permissions are not necessary because this field will go into each <group> on the models and the <group> has the related permission itself. And only can be created or modified by an administrator (should be a list of predefined reasons to block a customer or product)

Views are created but are not used (only form when opening the record on a pop-up).

Menus are not needed because there is no location to put it.

There is nothing to configure.

I'm going to add buttons to archive/restore records.

base_warn_option/models/warn_option.py
```diff
@@ -0,0 +1,27 @@
...
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
+
+from odoo import fields, models
+
+
+class WarnOption(models.Model):
+    _name = "warn.option"
+    _description = "Warn Option"
+
+    name = fields.Text(required=True, help="This text will be the warning message.")
```
EmilioPascual, line 11: IMHO: Maybe translate?

base_warn_option/models/warn_option.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2024 Moduon Team S.L.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
+
+from odoo import fields, models
+
+
+class WarnOption(models.Model):
```
edlopen, line 7: If this model is intended to be used as a base in other modules and does not have its own functionality, I think it would be better to make this class abstract so that it does not create a table in the database. What do you think?

base_warn_option/models/warn_option.py
Shide: No, because in te field that is going to be copied, is not translatable.

rafaelbn (approved):
I cannot test here.

In my side 👍🏼 base for the next

EmilioPascual (approved):
LGTM

edlopen:
LGTM

edlopen (approved):
LGTM

base_warn_option/models/warn_option.py
Shide: Must create the table because it's going to concentrate all the possible values (with `selection_add`) instead of creating additional models.

yajo (changes requested):
This module looks good, but since it does "nothing", I'd like to review at least one of the upcoming extensions to understand how both relate before approving.

Shide:
Added a naviable button in configuration to view all the warning options

base_warn_option/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+
+{
+    "name": "Base Warn Option",
+    "summary": "Add Options to Warn Messages",
+    "version": "15.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Tools",
+    "website": "https://github.com/OCA/server-ux",
+    "author": "Moduon, Odoo Community Association (OCA)",
+    "maintainers": ["Shide"],
```
yajo: issue:

```suggestion
    "maintainers": ["Shide", "rafaelbn"],
```

yajo (changes requested):
After reviewing dependant PRs https://github.com/OCA/sale-workflow/pull/3076 and https://github.com/OCA/purchase-workflow/pull/2240 I see clearly now how this module serves as a common base for those.

It looks OK but there's a missing requirement I just noticed before merging. Could you fix it please?

Thanks!

yajo (approved):

PR #854, opened 2024-04-11
[15.0][ENH] base_tier_validation, add Notify Reviewers by Approve Sequence

kittiu:
This enhancement add another option for Tier Definition.

**Notify reviewer by sequence**
 
Use with 'Approve by sequence' to ensure that, the next reviewer will be notified only when he/she is the next approver in sequence.

**Note:**
* This works independently from 'Notify reviewers on creation', so it shouldn't be used together.
* This method will not add subscription and will not notify any other followers except the reviewer.


kittiu:
cc @ps-tubtim

ps-tubtim (approved):
Functional Test: Ok

Saran440 (approved):
Tested :+1: . Run on production

bosd:
On first sight, this looks like an alternative implementation of https://github.com/OCA/server-ux/pull/638 `base_tier_validation_waiting`?

The use case you describe is available in V17. So maybe backport?

Saran440:
@bosd Hi, Thank you for suggest.

I tested on v17. it working :+1: but i test on your pr https://github.com/OCA/server-ux/pull/638 but still send email duplicate.

I think I can't backport from v17 to v15 because structure code `base_tier_validation` has changed.

However, this PR shouldn't merged to v15

Saran440 (changes requested):
In version17 use field name `notify_on_pending` can't merge this PR

PR #860, opened 2024-04-17
[16.0][ADD] base_tier_validation: set all form fields readonly when under validation

hbrunn:
this makes it obvious for users that they can't edit things when a record is under validation, and avoids doing so accidentally

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -1,12 +1,14 @@
 # Copyright 2017-19 ForgeFlow S.L. (https://www.forgeflow.com)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+import json
 from ast import literal_eval
 
 from lxml import etree
 
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
+from odoo.osv.expression import OR
 from odoo.tools.misc import frozendict
 
 
@@ -632,6 +634,21 @@ class TierValidation(models.AbstractModel):
                         all_models[model] = res["models"][model]
                 new_node = etree.fromstring(new_arch)
                 node.append(new_node)
+            excepted_fields = self._get_under_validation_exceptions()
+            for node in doc.xpath("//field[@name][not(ancestor::field)]"):
+                if node.attrib.get("name") in excepted_fields:
+                    continue
+                modifiers = json.loads(
+                    node.attrib.get("modifiers", '{"readonly": false}')
+                )
+                if modifiers.get("readonly") is not True:
+                    modifiers["readonly"] = OR(
+                        [
+                            modifiers.get("readonly", []) or [],
+                            [("review_ids", "!=", [])],
+                        ]
+                    )
+                    node.attrib["modifiers"] = json.dumps(modifiers)
             res["arch"] = etree.tostring(doc)
             res["models"] = frozendict(all_models)
         return res


base_tier_validation/models/tier_validation.py
```diff
@@ -632,6 +634,18 @@ def get_view(self, view_id=None, view_type="form", **options):
                         all_models[model] = res["models"][model]
                 new_node = etree.fromstring(new_arch)
                 node.append(new_node)
+            for node in doc.xpath("//field[@name][not(ancestor::field)]"):
+                modifiers = json.loads(
+                    node.attrib.get("modifiers", '{"readonly": false}')
+                )
+                if modifiers.get("readonly") is not True:
```
hbrunn, line 644: note to self: exclude the fields listed in `_get_under_validation_exceptions` here

Kev-Roche:
Hi, nice idea !

LoisRForgeFlow (commented):
EDIT ~There are possible exceptions though. You should get them with the method `_get_under_validation_exceptions`~

Sorry, I just saw your own comment https://github.com/OCA/server-ux/pull/860#discussion_r1569102735.

hbrunn:
@LoisRForgeFlow that's implemented now

LoisRForgeFlow (approved):
LGTM

Kev-Roche (approved):
LGTM

StefanRijnhart (approved):

base_tier_validation/models/tier_validation.py
```diff
@@ -632,6 +634,21 @@ def get_view(self, view_id=None, view_type="form", **options):
...
+                    modifiers["readonly"] = OR(
+                        [
+                            modifiers.get("readonly", []) or [],
+                            [("review_ids", "!=", [])],
+                        ]
+                    )
+                    node.attrib["modifiers"] = json.dumps(modifiers)
```
Kev-Roche, line 651: @hbrunn Am I missing something or the fields stay readonly, even if the reviews are validated or rejected ?

hbrunn: that's intentional. The goal is: Make it so you can't edit anything after requesting validation.
Where is that a problem?

Kev-Roche: Thank you for your answer. 
On my opinion, the review process is depending of the record state. After 1/ changing the record state and 2/validate or reject the review, the record need to be editable again (not readonly anymore) no ?
For exemple, if the fields are readonly only if a review_ids state is "pending" ?

hbrunn: I don't really see why you would want to allow editing for validated records. Doesn't that miss the point of validation? For validation rejected, review_ids is deleted once you restart the validation, so you can edit again. What's your use case?

If this is for some custom code of yours, maybe propose a refactoring of the for loop to call methods for its different aspects, and then you can override/customize this part in that method?

Kev-Roche: For exemple, for a product with 3 states : draft, normal, end of life. 
We validate a product in draft and put it in normal state. Then, we can not change the state as end of life. 
Sorry to insist, I see your point of view and I m agree but the module is based on the state of the record and between 2 states only. 
I will be happy to propose a refactor or a modification depending of your advices.

hbrunn: to me that would make the eol field an excepted field?

Otherwise, I see several options:
- don't look at review_ids here, but at validation_state == pending
- make it configurable for programmers with a flag on the class
- make it configurable for programmers by refactoring this into its own function

Kev-Roche: Thank you for your time, I like the last option, I made a PR : https://github.com/OCA/server-ux/pull/888

PR #873, opened 2024-05-06
[14.0][IMP] Add reminder to base_tier_validation

TDu:


base_tier_validation/models/tier_review.py
```diff
@@ -123,3 +124,20 @@ def _get_reviewers(self):
...
+
+    def _notify_review_reminder_body(self):
+        delay = (fields.Datetime.now() - self.create_date).days
+        return _("A review has been requested %s days ago.") % (delay)
+
+    def _notify_review_reminder(self):
+        record = self.env[self.model].browse(self.res_id)
+        post = "message_post"
+        if hasattr(record, post):
+            getattr(record.sudo(), post)(
```
simahawk: why posting a msg rather than creating an activity?

grindtildeath: Actually it can be done only if the model inheriting from `tier.validation` also inherits from `mail.activity.mixin` and not only `mail.thread` :grimacing: 

EDIT: There's another issue with activities if the tier definition requests validation to a group. As an activity can be assigned to only one user, we don't want to schedule an activity for each user in the group. I'd suggest to keep the version from @TDu or it would make things much more complicated than they need to be.

base_tier_validation/models/tier_definition.py
```diff
@@ -72,6 +72,10 @@ def _get_tier_validation_model_names(self):
         "this definition is triggered.",
     )
     has_comment = fields.Boolean(string="Comment", default=False)
+    notify_reminder_delay = fields.Integer(
+        string="Remind notifier",
```
grindtildeath:
```suggestion
        string="Send reminder message on pending reviews",
```

base_tier_validation/models/tier_definition.py
```diff
@@ -72,6 +72,10 @@ def _get_tier_validation_model_names(self):
...
+        help="Number of days after which the user should be reminded (0 = no reminder)",
```
grindtildeath:
```suggestion
        help="Number of days after which a message must be posted to remind about pending validation (0 = no reminder)",
```

grindtildeath:
@TDu can you please check https://github.com/camptocamp/server-ux/pull/1 ? :pray:

sebalix (approved):

base_tier_validation/models/tier_review.py
simahawk: worth mentioning in a comment?

base_tier_validation/models/tier_definition.py
```diff
@@ -92,3 +97,33 @@ def _compute_domain_reviewer_field(self):
             rec.valid_reviewer_field_ids = self.env["ir.model.fields"].search(
                 [("model", "=", rec.model), ("relation", "=", "res.users")]
             )
+
+    def _get_review_needing_reminder(self):
+        """Return all the reviews that have the reminder setup."""
```
simahawk, line 102: "all the reviews" but it returns just one record... who's right? :)

simahawk (approved):
LG, minor remark

simahawk:
Commits should be squashed in 1.
@LoisRForgeFlow gentle ping for a review :wink:

LoisRForgeFlow (commented):
LGTM :+1: 

Waiting the squash of commits and a commit msg that follows the convention `[TAG] module_name: title of changes` to approve

TDu:
I have squashed and reworded the commit

LoisRForgeFlow (approved):
:+1:

LoisRForgeFlow:
@TDu I would appreciate if you can take some time to fwport this to newer versions

TDu:
> @TDu I would appreciate if you can take some time to fwport this to newer versions

Please see #919 and #920

LoisRForgeFlow:
@TDu Thanks for the ports, I'm just missing v17, would you do that?

PR #876, opened 2024-05-07
[FIX] base_tier_validation: allow sudo writes

### LINKED ISSUE #875: base_tier_validation: blocking notifications cron, a.k.a. should it block sudo writes?
## Module
base_tier_validation

## Describe the bug
We have a database where `base_tier_validation` is installed along with other addons from its family.

The cron `mail.ir_cron_send_scheduled_message` is blocked in loop because of that addon.

## To Reproduce
**Affected versions**: 16.0 and probably others.

1. Have some record that:
  - Needs to notify a customer.
  - Has no portal token.
  - Is pending tier validation.
2. Run the cron (you can do it manually)

**Actual behavior**

Cron enters fail loop. Traceback:

    2024-05-06 12:30:04,883 42 INFO odoo odoo.addons.base.models.ir_cron: Starting job `Notification: Send scheduled message notifications`.
    2024-05-06 12:30:04,889 42 INFO odoo odoo.addons.mail.models.mail_message_schedule: Send 1 scheduled messages
    2024-05-06 12:30:04,921 42 ERROR odoo odoo.addons.base.models.ir_cron: Call from cron Notification: Send scheduled message notifications for server action #115 failed in Job #6
      File "/opt/odoo/custom/src/odoo/odoo/addons/base/models/ir_cron.py", line 391, in _callback
        self.env['ir.actions.server'].browse(server_action_id).run()
      File "/opt/odoo/custom/src/odoo/odoo/addons/base/models/ir_actions.py", line 674, in run
        res = runner(run_self, eval_context=eval_context)
      File "/opt/odoo/auto/addons/website/models/ir_actions_server.py", line 61, in _run_action_code_multi
        res = super(ServerAction, self)._run_action_code_multi(eval_context)
      File "/opt/odoo/custom/src/odoo/odoo/addons/base/models/ir_actions.py", line 544, in _run_action_code_multi
        safe_eval(self.code.strip(), eval_context, mode="exec", nocopy=True, filename=str(self))  # nocopy allows to return 'action'
      File "/opt/odoo/custom/src/odoo/odoo/tools/safe_eval.py", line 383, in safe_eval
        return unsafe_eval(c, globals_dict, locals_dict)
      File "ir.actions.server(115,)", line 1, in <module>
      File "/opt/odoo/auto/addons/mail/models/mail_message_schedule.py", line 51, in _send_notifications_cron
        messages_scheduled._send_notifications()
      File "/opt/odoo/auto/addons/mail/models/mail_message_schedule.py", line 80, in _send_notifications
        record._notify_thread(schedule.mail_message_id, msg_vals=False, **notify_kwargs)
      File "/opt/odoo/auto/addons/sms/models/mail_thread.py", line 240, in _notify_thread
        recipients_data = super(MailThread, self)._notify_thread(message, msg_vals=msg_vals, **kwargs)
      File "/opt/odoo/auto/addons/mail_mobile/models/mail_thread.py", line 25, in _notify_thread
        recipients_data = super(MailThread, self)._notify_thread(message, msg_vals=msg_vals, **kwargs)
      File "/opt/odoo/auto/addons/mail_post_defer/models/mail_thread.py", line 33, in _notify_thread
        return super()._notify_thread(message, msg_vals=msg_vals, **kwargs)
      File "/opt/odoo/auto/addons/mail/models/mail_thread.py", line 2471, in _notify_thread
        self._notify_thread_by_email(message, recipients_data, msg_vals=msg_vals, **kwargs)
      File "/opt/odoo/auto/addons/mail/models/mail_thread.py", line 2557, in _notify_thread_by_email
        recipients_groups_data = self._notify_get_recipients_classify(partners_data, model_name, msg_vals=msg_vals)
      File "/opt/odoo/auto/addons/mail/models/mail_thread.py", line 2957, in _notify_get_recipients_classify
        groups = self._notify_get_recipients_groups(msg_vals=local_msg_vals)
      File "/opt/odoo/auto/addons/sale/models/sale_order.py", line 1236, in _notify_get_recipients_groups
        groups = super()._notify_get_recipients_groups(msg_vals=msg_vals)
      File "/opt/odoo/auto/addons/portal/models/mail_thread.py", line 30, in _notify_get_recipients_groups
        access_token = self._portal_ensure_token()
      File "/opt/odoo/auto/addons/portal/models/portal_mixin.py", line 32, in _portal_ensure_token
        self.sudo().write({'access_token': str(uuid.uuid4())})
      File "/opt/odoo/auto/addons/delivery_auto_refresh/models/sale_order.py", line 92, in write
        )._auto_refresh_delivery()
      File "/opt/odoo/auto/addons/delivery_auto_refresh/models/sale_order.py", line 65, in _auto_refresh_delivery
        self.with_context(auto_refresh_delivery=True)._remove_delivery_line()
      File "/opt/odoo/auto/addons/delivery_auto_refresh/models/sale_order.py", line 112, in _remove_delivery_line
        res = super()._remove_delivery_line()
      File "/opt/odoo/auto/addons/delivery/models/sale_order.py", line 56, in _remove_delivery_line
        to_delete.unlink()
      File "/opt/odoo/auto/addons/delivery/models/sale_order.py", line 193, in unlink
        self.filtered('is_delivery').order_id.filtered('carrier_id').carrier_id = False
      File "/opt/odoo/custom/src/odoo/odoo/fields.py", line 1325, in __set__
        records.write({self.name: write_value})
      File "/opt/odoo/auto/addons/delivery_auto_refresh/models/sale_order.py", line 84, in write
        res = super().write(vals)
      File "/opt/odoo/auto/addons/sale_stock/models/sale_order.py", line 121, in write
        res = super(SaleOrder, self).write(values)
      File "/opt/odoo/auto/addons/mail/models/mail_thread.py", line 315, in write
        result = super(MailThread, self).write(values)
      File "/opt/odoo/auto/addons/mail/models/mail_activity_mixin.py", line 241, in write
        return super(MailActivityMixin, self).write(vals)
      File "/opt/odoo/auto/addons/base_tier_validation/models/tier_validation.py", line 282, in write
    odoo.exceptions.ValidationError: Esta operación está en proceso de validación.

**Expected behavior**
Cron should work. Notifications should be sent, no matter the validation status.

**Additional context**

A possibility to fix the problem would be:
1. Add `portal` to the dependencies of `base_tier_validation`
2. Override `_portal_ensure_token()`.
3. Call super `with_context(skip_validation_check=True)`

However, pay attention to this part of the traceback:

      File "/opt/odoo/auto/addons/portal/models/portal_mixin.py", line 32, in _portal_ensure_token
        self.sudo().write({'access_token': str(uuid.uuid4())})

As you can see, the token is being written with sudo. I think it's a bug to block writes in sudo.

@moduon MT-5997

yajo:
Fix https://github.com/OCA/server-ux/issues/875 by always allowing superuser to write.

@moduon MT-5997

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -251,6 +251,11 @@ class TierValidation(models.AbstractModel):
         return True
 
     def write(self, vals):
+        if (
+            self.env.is_superuser()
+            and self.env.context.get("skip_validation_check") is not False
+        ):
+            return super().write(vals)
         for rec in self:
             if rec._check_state_conditions(vals):
                 if rec.need_validation:
--- a/base_tier_validation/tests/test_tier_validation.py
+++ b/base_tier_validation/tests/test_tier_validation.py
@@ -55,6 +55,9 @@ class TierTierValidation(CommonTierValidation):
         record.invalidate_model()
         with self.assertRaises(ValidationError):
             record.write({"test_field": 0.5})
+        # Sudo writes always allowed
+        record.sudo().write({"test_field": 0.5})
+        self.assertEqual(record.test_field, 0.5)
 
     def test_06_validation_process_open(self):
         """Operation forbidden while a validation process is open."""


EmilioPascual (approved):

LoisRForgeFlow (commented):
Hi @yajo 

What is the process that needs to bypass the validation?

The problem I see with this change is that it is a bit to wide of a fix. For instance, it will allow procurements to extend POs under validation even when the procurement comes from a regular user. See https://github.com/odoo/odoo/blob/9fd48fcaa04e884871789abb7f19dacef49a2a8c/addons/purchase_stock/models/stock_rule.py#L139

yajo:
> What is the process that needs to bypass the validation?

You can see the full report in https://github.com/OCA/server-ux/issues/875. The process is the standard notifications creation cron, along with the need to auto-generate a portal token for some records. It's an Odoo standard process (you might see other modules in the traceback, but they don't affect this issue).



> For instance, it will allow procurements to extend POs under validation even when the procurement comes from a regular user. See https://github.com/odoo/odoo/blob/9fd48fcaa04e884871789abb7f19dacef49a2a8c/addons/purchase_stock/models/stock_rule.py#L139

AFAICS, [the `purchase_tier_validation` module validates only `purchase.order`, not `purchase.order.line`](https://github.com/OCA/purchase-workflow/blob/82bbd8dcebe4c131ad4b8ad6ae5a8e24ac70a951/purchase_tier_validation/models/purchase_order.py#L8), so IIUC this problem you say wouldn't happen.

Do you see other cases where this would be dangerous?

FWIW, reading this module code, it seems to me like it could be implemented as an extension to the rules system, where write permissions are extended by the validation state of a record. A sensible refactor would be to override `check_access_rule()` instead of overriding `write()`. And just like the rules system, IMHO it makes sense that `sudo()` overpasses them.

fcvalgar:
I'm testing in an EE client environment with the following modules installed and have not seen any problems with any of the flows used by these modules.

image

LoisRForgeFlow:
@yajo 

> You can see the full report in https://github.com/OCA/server-ux/issues/875.

Thanks, it was in the description, too quick reading... :man_facepalming: 

> AFAICS, [the purchase_tier_validation module validates only purchase.order, not purchase.order.line](https://github.com/OCA/purchase-workflow/blob/82bbd8dcebe4c131ad4b8ad6ae5a8e24ac70a951/purchase_tier_validation/models/purchase_order.py#L8), so IIUC this problem you say wouldn't happen.

EDIT: I did a wrong test, it appear to go throught, which was unexpected for me. I'll review.

EDIT2: So I tested v15 and the issue also happens, a procurement can extend a PO under validation, this was for sure not the intended behavior. So we have one more issue to fix... 

In any case, from my point of view since there are sudos in several places in odoo core that you cannot control, I'm not 100% sure that the patch you propose is the path to go.

yajo:
Yeah, it's a tricky case.

I could follow this other approach from https://github.com/OCA/server-ux/issues/875#issue-2282619751:

> 1. Add portal to the dependencies of base_tier_validation
> 2. Override _portal_ensure_token().
> 3. Call super with_context(skip_validation_check=True)

However I feel like it's the wrong approach. 

I mean: by default, when Odoo does some action with sudo, it usually means that it should happen always, no matter if you can or not write to the record. Thus, there are only a few sudos around, like the case at hand.

So I think it makes sense to assume that the tier validation modules should cover those cases explicitly if needed. I pushed a change to avoid skipping validations when `skip_validation_check` context is explicitly set to `False`.

But you're the maintainer. If you prefer the other solution, just tell me and I'll adapt the PR. It will involve depending on `portal`, though.

rafaelbn:
Hello @yajo ,

Then, after talking with @LoisRForgeFlow , the FIX https://github.com/OCA/delivery-carrier/pull/815 is enough ?

😄 ❤️ Thank all for you ideas and opinions with a great and sane contributions!

yajo:
> the FIX [OCA/delivery-carrier#815](https://github.com/OCA/delivery-carrier/pull/815) is enough ?

Yes, that fixes the problem without needing to merge this PR.

However, this PR still makes sense IMHO. I let you close it or merge it @LoisRForgeFlow 😃

LoisRForgeFlow:
> However, this PR still makes sense IMHO. I let you close it or merge it @LoisRForgeFlow 😃

Yes, I mean conceptually it makes sense. However, we uncovered the issue with PO lines that should be addressed and we also know that the procurements use sudo. I'll need to find some time to tinker a bit.

rafaelbn:
@yajo Review this please! 🙏🏼 I guess we can close it as you say in https://github.com/OCA/server-ux/issues/875 that it is solved in https://github.com/OCA/delivery-carrier/pull/799 so this PR is already solved in other module? Thank you! 😄 ❤️

yajo:
Nevertheless, this PR is probably the good thing to do, but I let @LoisRForgeFlow choose if merge or close.

For our particular issue, it's not needed because it was fixed in OCA/delivery-carrier#799.

PR #878, opened 2024-05-08
[ADD][16.0] developer_menu module

bealdav:
image


bealdav:
cc @dreispt @StefanRijnhart @Kev-Roche

Kev-Roche (approved):
LGTM

dreispt (approved):

developer_menu/views/menu_view.xml
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+
+<menuitem
+        id="conf_tech"
+        parent="base.menu_administration"
+        name="🧰"
```
legalsylvain: There is no images in menu item in odoo and OCA modules. Introducing it breaks that implicit rule. 

Also an image is less accessible and doens't brings any advantages.

legalsylvain (dismissed):
I think it's better to move menu entries into the new développer menu, instead of duplicate. 

Otherwise, all things will be duplicated. (In the search feature of web_responsive module )

developer_menu/views/menu_view.xml
bealdav: It's not an image, it's a unicode char like emoji. Odoo use emojis in some context.
We are no more in ascii or latin world ;-)

The purpose of OCA is not to always follow the Odoo way in all contexts : there are many examples. That's the key for innovation

The advantage is not to be more accessible, it's to be more compact (in settings apps which is not practical), it's intentional.

This module is not supposed to be installed on production instance as its name tells then no confusion for users.

> I think it's better to move menu entries into the new développer menu, instead of duplicate.

I don't think on my side, it's a matter of taste but if someone miss this shortcut , it expects to find orginal menus.

legalsylvain: > The advantage is not to be more accessible

Well, i'm not blind and i'm not very aware of such topic. But IFAIK, set an emoji as the name of an item is not accessible for blind people. Also it can not be searched...
Généraly, using image introduces a lot of matter of taste. For exemple, the current wallet image doesn't mean "configuration" for me, but for you it does.

bealdav: Ok, what do you suggest instead of this wallet aka tool box ?

legalsylvain: "Développer"

bealdav: This i not a wallet but https://emojipedia.org/fr/bo%C3%AEte-%C3%A0-outils

image

Are you agree with  `🧰 Dev` ? Easy to find because of color and meaningful and short then happy with responsive.

legalsylvain:
Well. I don't understand this module, and I find the design not in the spirit of odoo. (Duplicating entries, use icon in menu, ...)
However, i'll not block. I let maintainers merge this PR if they found it interesting. 

Regards

bealdav:
Thanks for your final words. However it seems your requested blocked merge with change requested. Is it possible to remove request and consider it as comment ?

Thanks a lot

mourad-ehm (approved):
LGTM

PaulGoubert (approved):
Working well and providing direct access to the main technical menus required for functional usage. BTW, the use of icons becomes quite intuitive fairly quickly.

PR #880, opened 2024-05-12
[17.0] [FIX] date_range: fix Translated stored related field warning

ramiadavid:
Fix warning:

`2024-05-12 18:17:49,361 7526 WARNING odoo odoo.fields: Translated stored related field (date.range.type_name) will not be computed correctly in all languages `

Removed relatad field type_name used for order table and repaced with type_id

### DIFF
--- a/date_range/models/date_range.py
+++ b/date_range/models/date_range.py
@@ -8,7 +8,7 @@ from odoo.exceptions import ValidationError
 class DateRange(models.Model):
     _name = "date.range"
     _description = "Date Range"
-    _order = "type_name, date_start"
+    _order = "type_id, date_start"
     _check_company_auto = True
 
     @api.model
@@ -27,7 +27,6 @@ class DateRange(models.Model):
         domain="['|', ('company_id', '=', company_id), ('company_id', '=', False)]",
         check_company=True,
     )
-    type_name = fields.Char(related="type_id.name", store=True, string="Type Name")
     company_id = fields.Many2one(
         comodel_name="res.company", string="Company", index=1, default=_default_company
     )
--- a/date_range/models/date_range_type.py
+++ b/date_range/models/date_range_type.py
@@ -12,6 +12,7 @@ from odoo.exceptions import ValidationError
 class DateRangeType(models.Model):
     _name = "date.range.type"
     _description = "Date Range Type"
+    _order = "name,id"
 
     @api.model
     def _default_company(self):


MarwanBHL (approved):

date_range/models/date_range.py
```diff
@@ -27,7 +27,7 @@ def _default_company(self):
         domain="['|', ('company_id', '=', company_id), ('company_id', '=', False)]",
         check_company=True,
     )
-    type_name = fields.Char(related="type_id.name", store=True, string="Type Name")
+
```
pedrobaeza: Don't add empty line here.

date_range/models/date_range.py
```diff
@@ -8,7 +8,7 @@
 class DateRange(models.Model):
     _name = "date.range"
     _description = "Date Range"
-    _order = "type_name, date_start"
+    _order = "type_id, date_start"
```
pedrobaeza, line 11: `type_id` destination model has name order?

pedrobaeza (changes requested):
Please remove `[17.0]` from the commit message. It should be in the PR title, but not on the commit message.

date_range/models/date_range.py
ramiadavid: No, the date.range.type model does not have any sorting parameters established, but I am not sure that it can be sorted by name either, since it is a translated field, I am not sure how Postgres will interpret that when it is stored as a JSON column.
I also don't see the fact that they are ordered by name as something very important.

date_range/models/date_range.py
ramiadavid: done

date_range/models/date_range.py
ramiadavid: Another option is that this field is not translatable, so it can be left as it was.

pedrobaeza: I think it's worth to try to put such order and see if it works.

ramiadavid: I really didn't expect this, because it really orders it correctly and to each user in their language.

date_range/models/date_range_type.py
```diff
@@ -12,6 +12,7 @@
 class DateRangeType(models.Model):
     _name = "date.range.type"
     _description = "Date Range Type"
+    _order = "name"
```
pedrobaeza: Always put `id` as fallback:

```suggestion
    _order = "name,id"
```

ramiadavid: done

pedrobaeza (approved):
/ocabot merge minor

PR #881, opened 2024-05-14
[14.0][UPD]base_tier_validation: Add search panel based on the model for the tier definition and sort the tier definition list with sequence and make it as handle widget.

Ppamungkas:
Add search panel based on the model for the tier definition and sort the tier definition list with sequence and make it as handle widget.

### DIFF
--- a/base_tier_validation/views/tier_definition_view.xml
+++ b/base_tier_validation/views/tier_definition_view.xml
@@ -6,16 +6,16 @@
         <field name="name">tier.definition.tree</field>
         <field name="model">tier.definition</field>
         <field name="arch" type="xml">
-            <tree>
+            <tree default_order="sequence">
+                <field name="sequence" widget="handle" />
                 <field name="model_id" />
                 <field name="name" />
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow:
@Ppamungkas could you check pre-commit?

Ppamungkas:
@LoisRForgeFlow done with the pre-commit. Sorry for not checking it first.

LoisRForgeFlow:
@Ppamungkas No worries, thanks for fixing it. Now, could you squash the 2 commits into a single one and adjust a bit the commit message to follow the [guidelines](https://www.odoo.com/documentation/12.0/developer/reference/guidelines.html#commit-message-structure)?

LoisRForgeFlow (commented):
There is an underlying issue pending to solve and it is that sequence is applied in the inverse order to define the reviews:

image

image

image



image

Your improvement is good, but I think it should be handle in conjunction with the reverse sequence problem or simple show the definition in sequence desc order.

PR #884, opened 2024-05-16
[16.0][FIX] base_tier_validation: use record from props instead of env in tier_review widget

AnizR:
Based on #793 

I had an issue adding a Many2one of a model that uses tier validation (in my case: purchase.order) in the form view of an other model (in my case: maintenance.request).

The PR #793 was partially fixing my issue but I still had an issue in the tier_review widget:
`const records = this.env.model.root.data.review_ids.records;`
We are trying to take the model from the env but, `this.env.model` represents the other model (in my case: maintenance.request).

It is better to retrieve the data from the props instead


IT-Ideas (approved):
Added JS commit makes totally sense.

PR #886, opened 2024-05-21
[FIX] server_action_mass_edit: Allow to modify readonly fields

rov-adhoc:


legalsylvain (changes requested):
At the first sight, I think it's not a good move.

Could you elaborate the context ? 

Thanks !

PR #887, opened 2024-05-22
[14.0] base_tier_validation_forward: Improve wizard usability 

grindtildeath:


TDu (approved):

grindtildeath:
@OCA/tools-maintainers anyone to reopen this please? :pray:

base_tier_validation/models/tier_validation.py
```diff
@@ -101,6 +101,7 @@ def _get_sequences_to_approve(self, user):
                 sequences.append(my_sequence)
         return sequences
 
+    @api.depends_context("uid")
```
LoisRForgeFlow, line 104: can you elaborate, why is it a missing decorator?

LoisRForgeFlow: forgot to ping you @grindtildeath

grindtildeath: @LoisRForgeFlow It's such a long time ago I don't remember exactly why it was needed. :disappointed: 

But if `self.env.user` is used in the compute function, it means the function still depends on connected user. I guess it could havve been we had some local tests where the value was not recomputed after altering the user, but I cannot tell for sure.

LoisRForgeFlow: it is hard for me to approve then. Without steps to reproduce or a description of what's being solved... :confused:

dreispt (approved):

LoisRForgeFlow (changes requested):
This discussion is pending to resolve https://github.com/OCA/server-ux/pull/887#discussion_r1831099383

I can't approve without further explanation. @dreispt maybe you have understood what is being fixed and can help me to understand?

PR #888, opened 2024-05-23
[16.0][IMP] configurable domain for readonly fields in validation

Kev-Roche:
As discussed here : https://github.com/OCA/server-ux/pull/860
This PR make the domain easily editable for readonly fields in validation.

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -590,6 +590,9 @@ class TierValidation(models.AbstractModel):
         new_node = etree.fromstring(str_element)
         return new_node
 
+    def _get_tier_validation_readonly_domain(self):
+        return [("review_ids", "!=", [])]
+
     @api.model
     def get_view(self, view_id=None, view_type="form", **options):
         res = super().get_view(view_id=view_id, view_type=view_type, **options)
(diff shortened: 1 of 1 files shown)


base_tier_validation/models/tier_validation.py
```diff
@@ -590,6 +590,9 @@ def _add_tier_validation_reviews(self, node, params):
         new_node = etree.fromstring(str_element)
         return new_node
 
+    def _get_readonly_domain(self):
```
hbrunn: this one is very likely to clash with some model function of the same name. I'd suggest something like `_get_tier_validation_readonly_domain` or similar. And for new functions, I think a docstring is always good.

hbrunn (changes requested):
I was thinking more in the lines of: Extract the body of every for loop in `get_view` into its own function, so programmers can modify each aspect more easily.

But that's probably indeed better done in the v17 version, and also there what you propose seems useful to me. So for me, just renaming the function would make this good to merge.

base_tier_validation/models/tier_validation.py
Kev-Roche: Good point, I changed it.

hbrunn (approved):

Abranes (approved):

PR #889, opened 2024-05-24
15.0 mig mass operation abstract

dani-xtendoo:
mass operation abstract migration to 15.0


legalsylvain (commented):
hi. I think that the authorship is reserved to contributors that makes majors changes. here this is quite simple adapatations to make the module working in V15.

Except that topic, LGTM. Thanks for porting this module.

PR #895, opened 2024-06-13
[13.0][FIX] base_optional_quick_create: add mail dependency

maisim:
Fix

psycopg2.ProgrammingError: column ir_model.is_mail_activity does not exist
LINE 1: ...","ir_model"."is_mail_thread" as "is_mail_thread","ir_model"...

on upgrade from 12.0 to 13.0

### DIFF
--- a/base_optional_quick_create/__manifest__.py
+++ b/base_optional_quick_create/__manifest__.py
@@ -9,7 +9,7 @@
     "author": "Agile Business Group,Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/server-ux",
     "license": "AGPL-3",
-    "depends": ["base"],
+    "depends": ["base", "mail"],
     "data": ["views/model_view.xml"],
     "installable": True,
 }


maisim:
:hand:  @OCA/tools-maintainers

thomaspaulb:
Would like to fasttrack but not finding a reference to `is_mail_activity` in the [module code](https://github.com/OCA/server-ux/blob/13.0/base_optional_quick_create) ?

maisim:
> Would like to fasttrack but not finding a reference to `is_mail_activity` in the [module code](https://github.com/OCA/server-ux/blob/13.0/base_optional_quick_create) ?

It is just to get the mail module updated before this one to avoid :

psycopg2.ProgrammingError: column ir_model.is_mail_activity does not exist
LINE 1: ...","ir_model"."is_mail_thread" as "is_mail_thread","ir_model"...


I didn't look further into what the problem was, seeing this error I concluded that there was a dependency on the mail module.
By adding this dependency it works (tm)

But I can take a closer look if you prefer.

maisim:
That said, I find the dependency rather logical (not annoying) because most creations are logged via the mail module.

thomaspaulb:
Do you still have the traceback that was posted around the error?

PR #661, opened 2023-06-06
[16.0][MIG] base_cancel_confirm

aliciagaarzo:
Migration of **base_cancel_confirm** module to version 16

PR #898, opened 2024-06-17
[16.0][FIX] base_cancel_confirm: correct get_view method

JordiMForgeFlow:
- Add missing **options argument
- Align 'get_view' logic with other similar implementations like https://github.com/OCA/server-ux/blob/16.0/base_tier_validation/models/tier_validation.py#L594 or https://github.com/OCA/edi-framework/blob/16.0/edi_oca/models/edi_exchange_consumer_mixin.py#L118. The current implementation used in the migration to V16 does not work (used in https://github.com/OCA/account-invoicing/pull/1744)

### DIFF
--- a/base_cancel_confirm/model/base_cancel_confirm.py
+++ b/base_cancel_confirm/model/base_cancel_confirm.py
@@ -6,6 +6,7 @@ from lxml import etree
 
 from odoo import _, fields, models, tools
 from odoo.exceptions import ValidationError
+from odoo.tools.misc import frozendict
 
 
 class BaseCancelConfirm(models.AbstractModel):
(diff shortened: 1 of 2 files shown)


JordiMForgeFlow:
@aliciagaarzo I can see you made these changes in the migration to v16: https://github.com/OCA/server-ux/pull/661/commits/8edbb21c58ac4f2d45ba7d09867f1f670c234f1e

Why did you changed it like this?

I am trying to add a module for the account move (https://github.com/OCA/account-invoicing/pull/1744) but with the changes you added in V16 the view gets broken:

image


I am trying to align with similar implementations in the OCA.

JordiMForgeFlow:
@kittiu @hugo-cordoba could you look into this one?

JordiMForgeFlow:
CC @OCA/tools-maintainers

thomaspaulb (approved):

GuillemCForgeFlow (approved):
makes sense 👍🏿

PR #899, opened 2024-06-20
[16.0][FIX] base_tier_validation, search only active definitions

ChrisOForgeFlow:
If main context has active_test setup as false, can be processed inactive definitions, this commit force to search only active definitions

@ForgeFlow


DavidJForgeFlow (approved):

LoisRForgeFlow (approved):
Seems legit

PR #902, opened 2024-06-27
[FIX] [16.0] base_tier_validation: Tier Validation Exceptions

Shide:
Add Fields Exceptions that allows certain fields can be modified during the validation process, after the validation process or in both cases.

https://www.loom.com/share/f684c7672d534e4d91f4e3cadc5ce1ee?sid=44e7eb31-0a71-4ec0-a8fa-dd589abeac8a

MT-6609 @moduon @yajo @rafaelbn @EmilioPascual @fcvalgar @Gelojr @LoisRForgeFlow please review if you want :)

### DIFF
--- a/base_tier_validation/README.rst
+++ b/base_tier_validation/README.rst
@@ -50,7 +50,7 @@ See `purchase_tier_validation <https://github.com/OCA/purchase-workflow>`_ as an
 Configuration
 =============
 
-To configure this module, you need to:
+To configure Tier Validations, you need to:
 
 #. Go to *Settings > Technical > Tier Validations > Tier Definition*.
 #. Create as many tiers as you want for any model having tier validation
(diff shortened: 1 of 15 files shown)


Gelojr (approved):
Nice job @Shide Your video makes it very clear
Thanks

fcvalgar (approved):
Functional review,

LGTM, thank you @Shide

rafaelbn (approved):
It looks great! Nice video! ❤️ 

> https://www.loom.com/share/f684c7672d534e4d91f4e3cadc5ce1ee?sid=44e7eb31-0a71-4ec0-a8fa-dd589abeac8a

We would appreciate @LoisRForgeFlow review 😄 🙏🏼 

Thank you!

LoisRForgeFlow (approved):
I think this block was an unexpected outcome of https://github.com/OCA/server-ux/pull/860#pullrequestreview-2073201756.

The idea behind the original change was legit: why would you allow to modify any record once it is validated? If you allow to do so, the validation loses a bit of its purpose.

But it was too restrictive and I agree that there are exceptions. Your proposal addresses that in a good and extensible way so :+1: from my side.

BTW, Thanks for taking the time to explain in a video :1st_place_medal:

PR #903, opened 2024-06-27
[IMP] document_quick_access_folder_auto_classification: Remove edi_storage_oca unnecessary dependancy

etobella:
The dependancy was only necessary for testing purpose, but it is really not needed


luisDIXMIT (approved):
LGTM!

PR #911, opened 2024-07-03
[16.0][IMP] base_tier_validation:  Add subscription mode in tier definition

BernatPForgeFlow:
Sometimes we may use Tier Validation for records that are communicating with the customer, and we don't want validators to receive notifications for the entire discussion. We introduce a new field 'Subscription mode' that allows to configure if we only want to receive Tier Validation notifications.


AlexPForgeFlow:
@LoisRForgeFlow can we reopen?

LoisRForgeFlow:
Closed in favor of #1043 1043

PR #913, opened 2024-07-04
[FIX] [16.0] base_tier_validation: Group to bypass Validation rules

Shide:
A security group has been added to allow skipping all validations and allowing you to edit records freely.

In submodules, security groups can be added to skip only validations that belong to a specific model or group.

PR Series:

- https://github.com/OCA/server-ux/pull/913 (This one)
- https://github.com/OCA/sale-workflow/pull/3222 (Group for sale)

MT-6238 @moduon @rafaelbn @Gelojr @yajo @fcvalgar @LoisRForgeFlow please reivew if you want :)

### DIFF
--- a/base_tier_validation/README.rst
+++ b/base_tier_validation/README.rst
@@ -78,6 +78,9 @@ To configure Tier Validation Exceptions, you need to:
 * If check *Write after Validation*, records will be able to be modified only in the defined fields when the Validation process is finished.
 * If check *Write after Validation* and *Write under Validation*, records will be able to be modified defined fields always.
 
+
+You can allow certain users to bypass Validations rules by adding them to the *Skip All Tier Validations* group.
+
 Known issues / Roadmap
 ======================
 
--- a/base_tier_validation/i18n/base_tier_validation.pot
+++ b/base_tier_validation/i18n/base_tier_validation.pot
@@ -6,6 +6,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 16.0\n"
 "Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2024-07-04 08:23+0000\n"
+"PO-Revision-Date: 2024-07-04 08:23+0000\n"
 "Last-Translator: \n"
 "Language-Team: \n"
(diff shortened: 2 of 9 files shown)


LoisRForgeFlow:
@Shide Can you describe a bit more your use case?

When someone can approve all tiers, can effectively "skip" (all tiers approved automatically) the validation process. I don't really see a reason to have this group at a first glance.

Thanks!

Shide:
But when the record is under validation or validated, the record is still in readonly and cannot be modified.

Right now we are facing this issue:
Sale is requested for validation and something validates it.
Sale is shipped and auto-invoiced (needs an invoice when goods are received by the customer).
Customer call us and tells that the price is incorrect.
Someone in Accounting team makes a complete refund and goes to the order to change price and generate the invoice again.

Here comes the problem: The record is in readonly state because it's validated and cannot be modified.
but
Odoo let you modify the price in a confirmed sale (sale state) when the invoiced quantity is 0. (+1 for the first invoice, -1 for the refund). So this line could have a price change if Tier Validation doesn't exists.

The people that requests tier, validates tier and modifies the order are different people with different roles.
Commercial, Commercial manager, Accounting

With this improvement, I could give the Accounting people (only peple we want, normally power users) the ability to modify orders no matter in what validation step they are.
Also, the sale is in readonly form to the other roles and the traceability of Validation process is not lost.

With the improvement in sale, we could give only this ability to sales (In case we want to install Purchase Validation or something like this).

I've thinking to restart validation, but we don't want to lose the Validation information because it's important to know who requested validation and who approves it.

https://www.loom.com/share/da4ff071dab745acaeecbbf9719d735c?sid=56f6e99b-256d-43d5-b457-4c3a44af0f9f

LoisRForgeFlow:
@Shide 

I see, thanks for the details.

I think this is good use case for the exceptions feature you did. You can add an exception for the price after validation. That would be the same level of security as tier validation just before the forced readonly changes. What else could we miss? the control on who can do it, in this sense what do you think about adding groups to exceptions? Wouldn't it be more configurable and enhance exceptions?

Shide:
Closed in favor of https://github.com/OCA/server-ux/pull/915

PR #916, opened 2024-07-08
15.0 fix base tier validation tier definition

Pani-k-folk:
Fix: Show only the names of users assigned as Reviewers specifically for that company.


Pani-k-folk:
I will resend for ease of verification

PR #917, opened 2024-07-08
[15.0][FIX] base_tier_validation:Tier definition

Pani-k-folk:
Fix:  Modify list of specific users to show only those who belong to that particular company

### DIFF
--- a/base_tier_validation/models/tier_definition.py
+++ b/base_tier_validation/models/tier_definition.py
@@ -7,6 +7,7 @@ from odoo import _, api, fields, models
 class TierDefinition(models.Model):
     _name = "tier.definition"
     _description = "Tier Definition"
+    _check_company_auto = True
 
     @api.model
     def _get_default_name(self):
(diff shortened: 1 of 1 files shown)


Pani-k-folk:
Hi @LoisRForgeFlow, Can you review this. This is the first OCA that I've attempted to modify. Hopefully, it will be beneficial to everyone.  ( = w=)

FrancoMaxime (approved):
LGTM: code review

newtratip (approved):

LoisRForgeFlow (changes requested):
Can you fix your commit message to follow the guidelines:

[TAG] module_name: short description of change

LoisRForgeFlow (changes requested):
The fix is also not correct, look at this example in runbot:

image

image


The check company is based in the company_id field that for users is not like in many other models, in users is just the default company, the companies that can be acceses are multiple in a user (company_ids). Thefore if you add the check_company you cannot create a tier definition in company B for a user with access in both A and B company.

PR #915, opened 2024-07-05
[FIX] [16.0] base_tier_validation: Tier Validation Exception Groups

Shide:
Add groups to Tier Validation Exceptions.

This will allow to set some fields only to certain group of users.

This PR is a simplification of https://github.com/OCA/server-ux/pull/913 and https://github.com/OCA/sale-workflow/pull/3222

https://www.loom.com/share/543a78d46cef4323b30a56cc5ad7329f

MT-6238 @moduon @rafaelbn @yajo @LoisRForgeFlow please review if you want :)

PR #918, opened 2024-07-08
[15.0][IMP] base_tier_validation: Allow define fields that is allowed modify under validation process as system parameters.

sergio-teruel:
cc @Tecnativa  TT49995

ping @CarlosRoca13 @chienandalu

### DIFF
--- a/base_tier_validation/README.rst
+++ b/base_tier_validation/README.rst
@@ -61,6 +61,7 @@ To configure this module, you need to:
 * If check *Notify Reviewers on Creation*, all possible reviewers will be notified by email when this definition is triggered.
 * If check *Comment*, reviewers can comment after click Validate or Reject.
 * If check *Approve by sequence*, reviewers is forced to review by specified sequence.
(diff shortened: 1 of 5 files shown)


LoisRForgeFlow:
Hi @sergio-teruel 

You might be interested on checking the latest improvement in v16 called "exceptions" https://github.com/OCA/server-ux/pull/902. It is fully configurable and there is also an ongoing improvement to allow to define exceptions based on security groups https://github.com/OCA/server-ux/pull/915.

Backporting it to v15 might be a good option for you.

sergio-teruel:
Hi @LoisRForgeFlow 
I've been checking and I see that I need more commits since 16.0. So at this point I'll leave it at that since this won't be carried over to later releases.

LoisRForgeFlow (changes requested):
@sergio-teruel Ok, as you prefer. I will submit a blocking review to remember that this won't ever be merged as it diverges from the main line followed in newer versions that can at some point be backported.

sergio-teruel:
Ok. Don't worry .. I close this PR and I fix in private module.

PR #920, opened 2024-07-08
[FWP][16.0] base_tier_validation: Add reminder option

TDu:
Foward port from v14 https://github.com/OCA/server-ux/pull/873


LoisRForgeFlow (approved):
Thanks!

PR #926, opened 2024-07-22
[17.0][FIX] base_tier_validation: Tier Validation Exception Groups

LoisRForgeFlow:
Fwport of https://github.com/OCA/server-ux/pull/915

The second commit has been skipeed as it was a big diff but apparently it was only modifying minor styling that was not described in the original PR, so I let the author decide if that style changes are still needed in v17.

cc @Shide


LoisRForgeFlow:
Proceeding with the fwport:

/ocabot merge minor

PR #928, opened 2024-07-24
[16.0][IMP] base_tier_validation: Change _compute_need_validation()

BernatPForgeFlow:
A document does not need validation only if there are no pending reviews, otherwise we should evaluate as True. We add a condition in request_validation to be able to request more than one time if there are new tier validations.

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -233,10 +233,12 @@ class TierValidation(models.AbstractModel):
                     ]
                 )
             )
-            valid_tiers = any([rec.evaluate_tier(tier) for tier in tiers])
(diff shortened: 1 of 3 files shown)


LoisRForgeFlow (approved):
LGTM :+1:

StefanRijnhart:
Would you consider adding a test to demonstrate what this effectively changes in the flow?

AlexPForgeFlow (approved):

StefanRijnhart:
Thanks for the update. Do I understand correctly that the commit message may be updated to say 
```
more than one time if there are new tier definitions
```
rather than new tier _validations_?

StefanRijnhart (approved):
Thanks for the updates!

PR #934, opened 2024-08-06
Fixes #216 in V16. Changes the ir.model name call so that it is trans…

### LINKED ISSUE #216: [13.0] Reviews widget model name translations
Module **base_tier_validation**

Review widget in systray is not showing the model name translation.

The model res_users In the function **review_user_count** return record._description as model **name** without translation.

Should call name_get function of the model to get name translated

Steps to reproduce:

Install module base_tier_validation and purchase_tier_validation to test.

You have to install a new language

When request validation on purchase the users reviewers can check the systray review widget.

Then change user language and there is no translation for model name.

Before:

image

After:

image

arpol-dev:
…lated in the systray review menu


LoisRForgeFlow (changes requested):
Hi,

Thanks for contributing. Could you have a look at pre-commit?

Also, could you adjust your commit message to follow the guidelines? (basically: [TAG] module_name: short title for changes)

PR #936, opened 2024-08-20
[FIX] base_revision: UX, open Form if only one revision record

dreispt:

PR #937, opened 2024-08-20
[FIX] base_revision: UX, open Form if only one revision record

dreispt:
Forward port of #936﻿


SirAionTech (approved):
Thanks for the PR!
Only code review, looks good to me.

A test would be nice, but not blocking in my opinion because it's a very little change.

dreispt:
This is a client UI action. I'm sot sure how I could test it. Any example you could share?

SirAionTech:
> This is a client UI action. I'm sot sure how I could test it. Any example you could share?

I was thinking to simply check in Python what `create_revision` returns when it creates only one revision.

PR #938, opened 2024-08-23
[FIX] base_tier_validation: warning on install

ica-adhoc:
This fixes removes the warning when install because two fields in tier_validation_exception model have the same name.


LoisRForgeFlow (approved):
Simple enough

/ocabot merge patch

LoisRForgeFlow:
@ica-adhoc can you fwport to v17?

StefanRijnhart:
Already fixed in 17 in https://github.com/OCA/server-ux/commit/a35ae39a2603622a258d394e125ee923c05bc11d

LoisRForgeFlow:
@StefanRijnhart Thanks for the pointer

PR #939, opened 2024-08-26
[17.0][IMP] base_tier_validation: support computed state field

StefanRijnhart:
Computed fields bypass `write`, so we need to override `_write` for that case.
Also, the current value before the update needs to be fetched from the database
because the new value is already set in the cache.

### DIFF
--- a/base_tier_validation/README.rst
+++ b/base_tier_validation/README.rst
@@ -44,6 +44,12 @@ See
 `purchase_tier_validation <https://github.com/OCA/purchase-workflow>`__
 as an example of implementation.
 
+Additionally, if your state field is a (stored) computed field, you need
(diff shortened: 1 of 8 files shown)


CasVissers-360ERP (approved):
Functional review

StefanRijnhart:
@LoisRForgeFlow Can I ask your opinion on this one? This change makes it trivial to add tier validation to models with a computed state field such as sale blanket orders, stock pickings etc.

antonioburic (approved):

houssine78 (approved):
LGTM

houssine78:
@LoisRForgeFlow any chance you can get a look at this PR ? thanks !!

LoisRForgeFlow (approved):
I think it makes sense, code review :+1: 

Thanks for adding a test case

PR #944, opened 2024-09-03
[16.0][IMP] server-ux: Add the possibility to disable create/edit from relational fields in form views

lmarion-source:
This module allows you to configure models so that you cannot create/edit a new instance/an existing instance
from a relational field on another model


etobella:
You edited a lot of files of other modules. And the commit description is not on the right format

lmarion-source:
@etobella My pre-commit changed some of your files. I removed those changes, they are not relevant.

base_optional_create_edit/README.rst
```diff
@@ -0,0 +1,91 @@
...
+=========================
+Base Optional Create Edit
+=========================
+
+Add the possibility to remove create and create/edit from many2one fields, for a specific model
+
+Installation
+============
+
+To install this module, you need to:
```
etobella: Add some readmes too. Right now, we have no description of this module

etobella:
Also, the commit message is not properly formed

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

base_optional_create_edit/models/ir_ui_view.py
```diff
@@ -0,0 +1,38 @@
...
+    @api.model
+    def get_views(self, views, options=None):
+        res = super().get_views(views, options)
+        views = res.get("views")
+        view_form = views["form"]["arch"]
+        tree = etree.fromstring(view_form)
+        view_fields = set(tree.xpath(".//field[not(ancestor::field)]"))
+
+        field_names = []
+        for field_name, model in self._fields.items():
```
acsonefho: For each relational field on the view, you'll execute a `search(...)`.
Please find another solution to do it only once :pray: 
And you can improve your domain to include your new field `avoid_create_edit`.

IMO your domain should be something like (it's just a draft):
```
comodel_names = [model.comodel_name for field_name, model in self._fields.items() if model.relational]
domain = [('avoid_create_edit', '=', True), ('model', 'in', comodel_names)]
models = self.env["ir.model"].search(domain)
fields_names = [field_name for field_name, model in self._fields.items() if model.comodel_name in models.model]
```

base_optional_create_edit/models/ir_ui_view.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class Model(models.AbstractModel):
+    _inherit = "base"
+
+    @api.model
+    def get_views(self, views, options=None):
+        res = super().get_views(views, options)
+        views = res.get("views")
+        view_form = views["form"]["arch"]
```
acsonefho: What's happening if you're loading views without form view?
You should be more broad for this generic code:
```
view_form = res.get("views", {}).get("form", {}).get("arch")
if view_form:
    tree = ...
    ...
```
Because if someone is not fit as you expect (but still correct of course), everything will be broken.

acsonefho (changes requested):

kikopeiro:
@lmarion-source 

Excuse me, but to understand it correctly, what does development contribute compared to this other one?

https://github.com/OCA/web/tree/16.0/web_m2x_options

lmarion-source:
Hello @kikopeiro ,

m2x_options allows to disable create/edit model by model. In my case, you just have to set a config parameter "avoid_create_edit" to True and this field is disabled for create/edit on all models it appears.

acsonefho (approved):

SAnnabelle (approved):
Functionally it's correct

PR #951, opened 2024-10-10
[18.0][MIG] server_action_mass_edit: Migration to 18.0

trisdoan:
### Note
- based on: https://github.com/OCA/server-ux/pull/980


florenciafrigieri2:
@trisdoan Hi! I tested the module but it didn't worked.
I created a server action to change the salesperson in the SOs, but when selecting the action within sales an error came up (I was not able to even see the wizard).

This are the last lines of the error:
  File "/mnt/data/odoo-addons-dir/server_action_mass_edit/wizard/mass_editing_wizard.py", line 93, in onchange
    dynamic_fields["selection__" + line.field_id.name] = fields.Selection(
  File "/opt/odoo/odoo/fields.py", line 2772, in __init__
    self._selection = dict(selection) if isinstance(selection, list) else None
ValueError: dictionary update sequence element #0 has length 0; 2 is required

I leave here a print of the server action
testserveraction

trisdoan:
> @trisdoan Hi! I tested the module but it didn't worked. I created a server action to change the salesperson in the SOs, but when selecting the action within sales an error came up (I was not able to even see the wizard).
> 
> This are the last lines of the error: File "/mnt/data/odoo-addons-dir/server_action_mass_edit/wizard/mass_editing_wizard.py", line 93, in onchange dynamic_fields["selection__" + line.field_id.name] = fields.Selection( File "/opt/odoo/odoo/fields.py", line 2772, in **init** self._selection = dict(selection) if isinstance(selection, list) else None ValueError: dictionary update sequence element #0 has length 0; 2 is required
> 
> I leave here a print of the server action testserveraction

Hi @florenciafrigieri2, thank you for the review

I just added a fix, could you please try again?

TDu (approved):
I don't see the error mentioned by @florenciafrigieri2 anymore, but I am not familiar with this module !
Code review looks good.

server_action_mass_edit/tests/test_mass_editing.py
```diff
@@ -0,0 +1,406 @@
...
+        "warning": {
+            "title": "This is a fake onchange",
+        },
+    }
+    return result
+
+
+@common.tagged("-at_install", "post_install")
+class TestMassEditing(common.TransactionCase):
+    def setUp(self):
```
henrybackman, line 25: Would `setUpClass` work for the common test setup to avoid running the setup separately for each test case?

server_action_mass_edit/tests/test_mass_editing.py
```diff
@@ -0,0 +1,406 @@
...
+        self.assertFalse(
+            result["operation_description_warning"],
+        )
+        self.assertEqual(
+            result["operation_description_danger"],
+            "None of the 1 record(s) you have selected can be processed.",
+        )
+
+    def test_wiz_fields_view_get(self):
+        """Test whether fields_view_get method returns arch.
```
henrybackman, line 133:
```suggestion
    def test_wiz_get_view(self):
        """Test whether get_view method returns arch.
```

server_action_mass_edit/wizard/mass_editing_wizard.py
```diff
@@ -0,0 +1,339 @@
...
+                        ).with_context(
+                            mass_edit=True,
+                        ).write(values)
+        return super().create([{}])
+
+    def _prepare_create_values(self, vals_list):
+        return vals_list
+
+    def read(self, fields=None, load="_classic_read"):
+        """Without this call, dynamic fields build by fields_view_get()
```
henrybackman, line 323:
```suggestion
        """Without this call, dynamic fields build by get_view()
```

henrybackman (approved):
Couple of minor comments

florenciafrigieri2 (approved):
Functional test! Now is working ok!

stefsava:
It seems like there is a regression.
I can't bulk insert product attributes, while I could do it in 17.0.

Before it asked me to add a line,
now it doesn't ask and there are multiple columns. 

With 17.0
image
image

With 18.0 and this PR
image
image

legalsylvain (changes requested):
See @stefsava comment.

trisdoan:
Hello @stefsava, I checked and it looks to me that the issue is not because of the module
1) I tried with different o2m fields in another model
server_mass_action_edit

2) For product attribute, after `https://github.com/odoo/odoo/pull/160270`, there is a list view created and it has attribute `create='false'`. I changed to `create='true'` (for testing only), here's result:
server_mass_action_edit_2

stefsava:
Hello @trisdoan 

your suggestion doesn't solve my use case.

If I change create='true' I can see "Add a row", but the choice doesn't lead to anything useful.

Instead I found a workaround that brings the behavior back to that of 17.0 that I think can help you understand the problem detected.

I discovered that 17.0 only has the view "product.template.attribute.line.form" while
18.0 has added "product.template.attribute.line.view.list", and for some reason that I don't understand the list view is chosen instead of the form view.

My workaround was simply to deactive the list view, although I fear it may introduce regressions elsewhere.

image

trisdoan:
~~DRAFT: working on failed test~~

trisdoan:
Hi, I updated the code to include a pre-hook which handles the renaming (benefit those who use the addon in 14.0 and now migrate to 18.0)

PaulGoubert (approved):
Functionnal review OK

grindtildeath:
Thanks for the mig.
While we are at it, can we please bring https://github.com/OCA/server-ux/pull/963 in the new version? :pray: 

cc @etobella @pedrobaeza @legalsylvain

pedrobaeza:
As said already in the other PR, that should be a different module, not integrated in this one. And even more in these recent versions, where onchanges are rare, as most of them have been replaced by computed writable fields.

legalsylvain:
> As said already in the other PR, that should be a different module, not integrated in this one. And even more in these recent versions, where onchanges are rare, as most of them have been replaced by computed writable fields.

As @grindtildeath, I'm also in favor of the integration of #963 in V18. 
- The main critic about #963 was the addition of a new dependency in a stable and widely deployed module, that could generate for certain types of deployment. In https://github.com/OCA/odoo-community.org/issues/193, add dependency in a migration PR was a consensus.
- in #963, most of people was considering the PR as a bugfix. 
- In the meantime, there is an unanswered question here : https://github.com/OCA/server-ux/pull/963#issuecomment-2449760770. If you could elaborate a real use case when it is risky, opinion could change.

Thanks  !

pedrobaeza:
The module is there for 10 versions, and now it's a critical problem to not play onchanges. Come on...

etobella:
My 2 cents.

Odoo is deprecating the use of onchange. If you see numbers, on 15, we had 513 onchange and 1645 depends on odoo core code. On 18, we have 461 onchange and 2935 depends. I think we shouldn't add this code and leave it on another module, the cost is minimal and if we check numbers, odoo is removing onchange for depends because we can. I would recommend to remove onchange on most of the modules we create if possible.

rousseldenis:
I'm in favor too in #963 integration.

> The module is there for 10 versions, and now it's a critical problem to not play onchanges. Come on...

In fact, I think I see the behavior expected by @pedrobaeza.

This module was used to mass edit fields without touching any other value in the record or elsewhere. A kind of database editor. Stop me if I'm wrong.

BUT:

- That behavior is not documented in README. So, people are expecting the changes to automatically change the other values like in any other Odoo flow.
- Current users are maybe not aware that their changes do not impact other fields (the remaining onchanges ones).

So, I think saying a misbehavior present for several versions is not a bug is not correct. As said before, maybe nobody saw that problem...

@pedrobaeza So, you rely on compute methods code and not in onchanges one ? That's weird.

IMHO, we should improve modules behavior when we can. When Odoo remove all onchanges, we will remove it.

>  I would recommend to remove onchange on most of the modules we create if possible.

@etobella Of course, this should maybe added in contribution guidelines. Concerning the numbers, I would say if there is only 1 onchange, that code should be still applicable.

pedrobaeza:
Yes, onchanges are designed for changes in UI during the manual manipulation of the record, relying also in view features like readonly or non existing fields for its behavior. Computed methods, on contrary, are designed to work in a view-agnostic mode, and are more confident. Executing onchanges in a wild mode doing mass editions can lead to unexpected behaviors.

And anyway, the same you have the module `sale_crm` as a glue module between `sale` and `crm`, you must have `server_action_mass_edit_onchange` for the same, and make it optional for people.

Please stop wasting the time of all the good contributors involved in this discussion while there's a solution that fits all, although you consider it a bug. The bug is solved installing the module in all your instances.

etobella:
According to Odoo documentation:
https://www.odoo.com/documentation/16.0/developer/tutorials/getting_started/09_compute_onchange.html?highlight=onchange

Odoo is not recommending onchange. So, I do not recommend to add this functionality here because it might be a problem later.

rousseldenis:
> The bug is solved installing the module in all your instances.

Which one and where?

> Please stop wasting the time of all the good contributors

I suppose I'm on the other side

> here's a solution that fits all,

I thought @grindtildeath @simahawk @legalsylvain @mmequignon were not in that way. But maybe I'm wrong

pedrobaeza:
> Which one and where?

Seriously? You are the one saying that it's a bug to not play onchanges. So the bug is solved creating the glue module and installing it.

rousseldenis:
> > Which one and where?
> 
> Seriously? You are the one saying that it's a bug to not play onchanges. So the bug is solved creating the glue module and installing it.

Ok, don't be harsh, I've just reread the comments and some people stil say (or never said the contrary) that this is a bug:

- https://github.com/OCA/server-ux/pull/963#issuecomment-2449625834
- https://github.com/OCA/server-ux/pull/963#pullrequestreview-2406042937
- https://github.com/OCA/server-ux/pull/963#issuecomment-2447856267

They were the basis for my questioning.

pedrobaeza:
Why forcing those that are totally against this "bugfix" to have it, even only in the new version? If there's no other solution, I understand that this debate makes sense, but the simpler solution to avoid the debate is to have that extra module. We are not saying that is forbidden to have it on OCA. We are just asking to not integrate it in the base module.

legalsylvain:
> Odoo is deprecating the use of onchange. 

No. I read the text of your link (https://www.odoo.com/documentation/16.0/developer/tutorials/getting_started/09_compute_onchange.html?highlight=onchange) and there is no section about deprecation or about "obsolete feature that should not be used in the futur". Just explanation about pro and cons. I remember clearly an answer of Odoo SA developper, during a recent Odoo experience, saying that both tools remain valid.

V12 : 504 onchanges / 638 depends
V13 : 572 onchanges / 752 depends
V14 : 479 onchanges / 1258 depends
V15 : 507 onchanges / 1625 depends
V16 : 447 onchanges / 2168 depends
V17 : 452 onchanges / 2703 depends
V18 : 459 onchanges / 2934 depends

I don't see any massive withdrawal of onchange to make it disappear. On the other hand, there is a more massive use of the depends function. Between V12 and V18, onchange : -9% ; depends : +360%

regards.


> Ok, don't be harsh, I've just reread the comments and some people stil say (or never said the contrary) that this is a bug.

@rousseldenis : As mentioned [in this long comment](https://github.com/OCA/server-ux/pull/963#issuecomment-2448804851), Odoo SA developpers says it's a bug also ;-) (in fact Odoo prohibits mass edition on fields if there are onchange on it : see commit message and diff : https://github.com/odoo/odoo/commit/14198a25cb2a53f42acf920c6a6222f2d638f86d#diff-4c25e9b1afc64d76fd27b33de1bc39a16b8b79fc254f21ab350e51b7deae9b02R40)

grindtildeath:
I agree we should prefer compute to onchange but IMO this discussion is missing the point.

Basically, we went from "do not introduce this in a stable version" to "do not introduce that because _I don't want it_" ... 

FWIW my patch is adding a bugfix (or a feature if you prefer) that is opt-in and won't change anything if someone does not want to use it. The only drawback being an added dependency on a module that is not in the same repository but is still in the OCA, to avoid repeating code.

But I guess I'm just a bad contributor wanting to waste the time of the good ones...

pedrobaeza:
You are a good contributor, but stucked on imposing to some other good contributors something they don't want, like it's depending on other module. The solution for all is the extra module. Put if you want in the README of the base module that there's the glue module for the onchange case, but don't force people to install `onchange_helper` on the base module.

P.S.: I have never said to put it on the migration. I oppose to the dependency change in any version.

legalsylvain:
> The solution for all is the extra module. Put if you want in the README of the base module that there's the glue module for the onchange case, but don't force people to install onchange_helper on the base module.

Please Pedro. The main question is : "_is it a mistake not to execute the onchange ?_" If the answer is yes, propose to make an extra module, and write in the readme of the first module "_There is a bug in this module, but you can fix it installing another module_" doesn't makes sense.

For the time being, 
- there are more OCA contributors who think there's a mistake than who don't.
- Odoo SA developers also think it's a mistake.

1) I asked [a question](https://github.com/OCA/server-ux/pull/963#issuecomment-2449685310) 2 monthes ago ("_Provide a use case where applying onchange would be **dangerous**?_"). I need an answer to change my mind. (except speed of execution that depends on the number of items and because there may also be performance problems with api.depends)

2) When I explain in one sentence to my novice users what the mass editing wizard is for, I tell them that “it avoids the need to edit items one by one.”
When you edit one by one, the onchange is executed.

pedrobaeza:
There are also a lot of persons not wanting to put this dependency. Why the people wanting to put it worth more than the one not wanting?

And more having a solution for all, which is the extra module.

rousseldenis:
> _There is a bug in this module, but you can fix it installing another module_" doesn't makes sense.

> When you edit one by one, the onchange is executed

That are the points that make me feel uncomfortable with the former implementation and with the 'solution'.

pedrobaeza:
I'm horrified now seeing that in 16.0, this `onchange` thing was introduced wildly in the migration:

https://github.com/OCA/server-ux/commit/617be5f65e8c30c509e24ab00cab30f9849ac5ac#diff-8d785e151e9ad076da0aff7904315bc3ca9dc70ad1fca741b35e07beced4198fR75

That should go out as it was in <=15.0.

grindtildeath:
@pedrobaeza Why is the dependency such an issue? 

If we want to go for an extra module, IMO we should restrict the possibility to mass edit fields having an onchange that must be played, or at least introduce a warning that editing such fields might introduce inconsistencies, which would then make it more complex than just introducing the onchange ability as done in my PR...

pedrobaeza:
It's not only the dependency, which I also don't want to bring to all my projects, but the fact of giving users the possibility of executing onchanges without knowing what they imply (I don't even know if there's a side effect depending on the combination of fields to change!!). I have already said about some risks changing other fields apart from the ones you are explicitly saying you want to change, and having the check is not enough, as these users touch directly these server actions. Having this option isolated in a module, I won't install it in my instances to avoid that problem.

trisdoan:
Hello @pedrobaeza, @grindtildeath , @legalsylvain , @etobella , @rousseldenis , I extracted the proposal of Akim into a new module: https://github.com/OCA/server-ux/pull/985

tate11:
> Hello @pedrobaeza, @grindtildeath , @legalsylvain , @etobella , @rousseldenis , I extracted the proposal of Akim into a new module: #985

I tried it locally, it doesn't work.

2025-01-13 22:11:13,708 18900 ERROR odoo_v18 odoo.sql_db: bad query: b'INSERT INTO "product_supplierinfo" ("company_id", "create_date", "create_uid", "currency_id", "date_end", "date_start", "delay", "discount", "min_qty", "price", "product_code", "product_id", "product_name", "product_tmpl_id", "sequence", "write_date", "write_uid") VALUES (2, \'2025-01-13 22:11:13.699676\', 2, 31, NULL, NULL, 1, \'0.00\', \'0.00\', \'0.00\', NULL, NULL, NULL, 281, 1, \'2025-01-13 22:11:13.699676\', 2) RETURNING "id"'
ERROR: null value in column "partner_id" of relation "product_supplierinfo" violates not-null constraint
DETAIL:  Failing row contains (26, null, 1, 2, 31, null, 281, 1, 2, 2, null, null, null, null, 0.00, 0.00, 0.00, 2025-01-13 22:11:13.699676, 2025-01-13 22:11:13.699676).
 
2025-01-13 22:11:13,712 18900 WARNING odoo_v18 odoo.http: The operation cannot be completed:
- Create/update: a mandatory field is not set.
- Delete: another model requires the record being deleted. If possible, archive it instead.

Model: Supplier Pricelist (product.supplierinfo)
Field: Vendor (partner_id)
Screenshot 2025-01-14 at 01-09-58 Vendors (Product)
Screenshot 2025-01-14 at 01-11-41 Products

trisdoan:
> > Hello @pedrobaeza, @grindtildeath , @legalsylvain , @etobella , @rousseldenis , I extracted the proposal of Akim into a new module: #985
> 
> I tried it locally, it doesn't work.
> 
> 2025-01-13 22:11:13,708 18900 ERROR odoo_v18 odoo.sql_db: bad query: b'INSERT INTO "product_supplierinfo" ("company_id", "create_date", "create_uid", "currency_id", "date_end", "date_start", "delay", "discount", "min_qty", "price", "product_code", "product_id", "product_name", "product_tmpl_id", "sequence", "write_date", "write_uid") VALUES (2, '2025-01-13 22:11:13.699676', 2, 31, NULL, NULL, 1, '0.00', '0.00', '0.00', NULL, NULL, NULL, 281, 1, '2025-01-13 22:11:13.699676', 2) RETURNING "id"' ERROR: null value in column "partner_id" of relation "product_supplierinfo" violates not-null constraint DETAIL: Failing row contains (26, null, 1, 2, 31, null, 281, 1, 2, 2, null, null, null, null, 0.00, 0.00, 0.00, 2025-01-13 22:11:13.699676, 2025-01-13 22:11:13.699676).
> 
> 2025-01-13 22:11:13,712 18900 WARNING odoo_v18 odoo.http: The operation cannot be completed:
> 
>     * Create/update: a mandatory field is not set.
> 
>     * Delete: another model requires the record being deleted. If possible, archive it instead.
> 
> 
> Model: Supplier Pricelist (product.supplierinfo) Field: Vendor (partner_id) Screenshot 2025-01-14 at 01-09-58 Vendors (Product) Screenshot 2025-01-14 at 01-11-41 Products

Hello @tate11, it looks to me the issue is not due to the module

I got same error with version 17.0

Could you please provide steps to reproduce?

StephaneMangin:
@legalsylvain A little UP on this topic to go forward.

legalsylvain:
> @legalsylvain A little UP on this topic to go forward.

Unfortunately, we don't find any solution for the onchange bug. So at this step, we are stuck. I don't know how to move forward. 

Regards.

pedrobaeza:
It's not a bug. Just have the base module and add an extra module for executing the onchanges.

StefanRijnhart (approved):
Thank you @trisdoan! Tested successfully.

@tate11 I think the issue is that the supplierinfo list view sets partner_id to readonly (https://github.com/odoo/odoo/blob/18.0/addons/product/views/product_supplierinfo_views.xml#L93). If I reset this attribute in my local database the mass editing of vendors on products works as expected. I don't think we can fix it easily here.

About the onchange situation, @trisdoan has already done the work to split it up in a separate module (in https://github.com/OCA/server-ux/pull/985) so I think we can go ahead on this one and continue the discussion there.

pedrobaeza:
Yes, we can continue thus:

/ocabot merge nobump

omalbastin:
Why in v18, we have external dependency to openupgradelib package?

server_action_mass_edit/hooks.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2024 Camptocamp (<https://www.camptocamp.com>)
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html)
+from openupgradelib import openupgrade
+
+
+def pre_init_hook(env):
```
legalsylvain, line 6: Indeed. @trisdoan : could you elaborate, why you re-introduced this pre-init hook. It looks quite obsolete, and force to install openupgradelib.

thanks for your answer !

legalsylvain:
any news @trisdoan ?

thanks!

PR #952, opened 2024-10-10
[17.0][IMP] base_technical_features: Remove DeprecationWarning: XML declarations in HTML module descriptions

victoralmau:
Remove DeprecationWarning: XML declarations in HTML module descriptions

`WARNING devel py.warnings: /opt/odoo/custom/src/odoo/odoo/addons/base/models/ir_module.py:178: DeprecationWarning: XML declarations in HTML module descriptions are deprecated since Odoo 17, base_technical_features can just have a UTF8 description with not need for a declaration`

@Tecnativa


pedrobaeza (changes requested):
The HTML file has been entirely written instead of just removing the first line. I think it's due to your IDE changing line feed or similar.

pedrobaeza (approved):
/ocabot merge patch

PR #953, opened 2024-10-15
[18.0][MIG] base_technical_features: Migration to 18.0

Kimkhoi3010:
I found only 1 commit to port but it was ignored, I added it to blacklist: [[17.0][FW] base_technical_features: port from 14.0](https://github.com/OCA/server-ux/pull/954)

# Reason for change:
The original logic in _postprocess_access_rights removes base.group_no_one for non-admin users, which restricts their access to certain UI components. Extending _get_view ensures that technical users (part of base_technical_features.group_technical_features) still have access to these components without needing full admin privileges.


base_technical_features/models/base.py
```diff
@@ -0,0 +1,38 @@
...
+
+        :param int view_id: The ID of the view to retrieve. If None, defaults
+                            to the specified view type.
+        :param str view_type: The type of the view to return ('form', 'list', etc.).
+        :param dict options: Additional options for fetching the view.
+
+        :return: A tuple containing:
+            - etree.Element: The modified view architecture as an etree node.
+            - ir.ui.view: The original view record retrieved.
+        """
```
chienandalu: Don't override the docstring. Add a regular comment if needed instead.

base_technical_features/models/ir_ui_menu.py
```diff
@@ -0,0 +1,17 @@
...
+from odoo import api, models
+
+
+class IrUiMenu(models.Model):
+    _inherit = "ir.ui.menu"
+
+    @api.model
+    def _visible_menu_ids(self, debug=False):
+        """Set debug = True, so that group_no_one is not filtered out of the
+        user's groups"""
```
chienandalu, line 12: Same here

base_technical_features/models/base.py
```diff
@@ -1,18 +1,38 @@
...
 
 class Base(models.AbstractModel):
     _inherit = "base"
 
-    def user_has_groups(self, groups):
-        """Return True for users in the technical features group when
-        membership of the original group is checked, even if debug mode
-        is not enabled.
+    @api.model
+    def _get_view(self, view_id=None, view_type="form", **options):
```
chienandalu: Is it enogh to just replace the group in view?

base_technical_features/models/base.py
```diff
@@ -1,18 +1,38 @@
...
+        :param dict options: Additional options for fetching the view.
+
+        :return: A tuple containing:
+            - etree.Element: The modified view architecture as an etree node.
+            - ir.ui.view: The original view record retrieved.
+        """
+        arch, view = super()._get_view(view_id=view_id, view_type=view_type, **options)
+        arch_str = etree.tostring(arch, encoding="unicode")
+        arch_str = arch_str.replace(
+            "base.group_no_one", "base_technical_features.group_technical_features"
```
chienandalu: Maybe this would be more correct, as otherwise a user with no `base_technical_features.group_technical_features` wouldn't access to the regular `base.group_no_one` behavior in debug mode
```suggestion
            "base.group_no_one", "base.group_no_one,base_technical_features.group_technical_features"
```

chienandalu (changes requested):
Thanks :) Some comments below

base_technical_features/models/base.py
Kimkhoi3010: That is the docstring available from this commit: [204294](https://github.com/OCA/server-ux/commit/2042944b2127655430035894e2f57d8c6e349ce3)

base_technical_features/models/ir_ui_menu.py
Kimkhoi3010: That is the docstring available from this commit: [204294](https://github.com/OCA/server-ux/commit/2042944b2127655430035894e2f57d8c6e349ce3)

henrybackman (approved):

base_technical_features/models/base.py
```diff
@@ -0,0 +1,26 @@
...
+from odoo import api, models
+
+
+class Base(models.AbstractModel):
+    _inherit = "base"
+
+    @api.model
+    def _get_view(self, view_id=None, view_type="form", **options):
+        arch, view = super()._get_view(view_id=view_id, view_type=view_type, **options)
+        arch_str = etree.tostring(arch, encoding="unicode")
```
simahawk, line 15: wouldn't be better to check if the string is there 1st?

simahawk (approved):
LG overall, minor remark.

@chienandalu good for you?

chienandalu (approved):
:+1:

PR #958, opened 2024-10-22
[16.0][ADD] mail_suggested_recipient_unchecked: New module

victoralmau:
New module

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT51197


mail_suggested_recipient_unchecked/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module uncheck the suggested recipients by default.
```
pedrobaeza:
```suggestion
This module unchecks the extra suggested recipients by default when going
to write a new message in a chatter thread.

Odoo suggests new recipients on several parts of the suite to accommodate
business needs. Example: on CRM opportunities, it proposes the lead
partner to be added to the thread (or to create a partner from the email 
if not exists) when the partner is not follower yet. Another similar example
happens with project tasks with a linked customer that is not yet follower.
```

mail_suggested_recipient_unchecked/readme/USAGE.rst
```diff
@@ -0,0 +1,3 @@
+#. Go to Projects and create a task linked to a customer (Joel Willis for example).
```
pedrobaeza: This is a specific example, not a generic one, even not reproducible without installing `project`.

CarlosRoca13 (approved):

mail_suggested_recipient_unchecked/readme/USAGE.rst
```diff
@@ -0,0 +1,3 @@
+#. Go to Contacts and create a new one with an email address.
```
pedrobaeza: This is still a specific case, although included in the `base` module. The idea is to redact a generic text like "#. On a chatter where recipients can be suggested, click on "Send Message", and bla, bla, bla", and then you can put an example where it happens, like the contact one, but don't make users thing that this is only working for that case.

mail_suggested_recipient_unchecked/static/src/js/suggested_recipient_info.esm.js
```diff
@@ -0,0 +1,24 @@
...
+    fields: {
+        overrideIsSelected: attr({
+            compute() {
+                return this.partner ? this.isChecked : false;
+            },
+            default: false,
+        }),
+        // We need to define an auxiliary field that allows us to set the default
+        // value to false, since overwriting an existing field with this input is
+        // not supported: https://github.com/Tecnativa/odoo/blob/518518f53d81d4725babc42b51bfd1336cbf56c9/addons/mail/static/src/model/model_core.js#L310-L311
```
chienandalu: Better link Odoo's repo
```suggestion
        // not supported: https://github.com/odoo/odoo/blob/518518f53d81d4725babc42b51bfd1336cbf56c9/addons/mail/static/src/model/model_core.js#L310-L311
```

chienandalu (commented):
Tested :+1: 

Just a small detail:

chienandalu (approved):

mail_suggested_recipient_unchecked/readme/USAGE.rst
pedrobaeza: Now it's the example what is missing, hehe.

victoralmau: Changed, is it correct now?

pedrobaeza (approved):
I expected a more complete example, but I stop nitpicking

/ocabot merge nobump

PR #631, opened 2023-04-14
[IMP]-15.0-Improve error message when update record in batch

hieulucky111:
Add one more option in the Mass Editing Wizard to allow user to be easy recognize which record is making error.
Ex:
When we update Product Template in Batch without checkbox "Write Record By Record"
image

When use check the checkbox "Write Record By Record"
image

PR #962, opened 2024-10-28
[16.0][IMP] server_action_mass_edit: Improve error message when update record in batch

carlosdauden:
FW-port from https://github.com/OCA/server-ux/pull/631

@Tecnativa
ping @sergio-teruel @chienandalu

### DIFF
--- a/server_action_mass_edit/wizard/mass_editing_wizard.py
+++ b/server_action_mass_edit/wizard/mass_editing_wizard.py
@@ -3,10 +3,19 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 import json
+import logging
 
 from lxml import etree
+from psycopg2 import IntegrityError
 
 from odoo import _, api, fields, models
+from odoo.exceptions import (
+    AccessDenied,
+    AccessError,
+    MissingError,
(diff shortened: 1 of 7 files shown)


chienandalu (commented):
Looks like something else might be missing as the installation fails in the CI

sergio-teruel (approved):
Functional review ok

server_action_mass_edit/wizard/mass_editing_wizard.py
```diff
@@ -281,9 +299,44 @@ def create(self, vals_list):
                             values.update({split_key: m2m_list})
 
                 if values:
```
chienandalu: Maybe you can avoid the complexity warning `C901` like this
```suggestion
                if not values:
                    continue
```

carlosdauden: Changed, but `noqa: C901` already is needed

carlosdauden: @chienandalu is OK?

chienandalu (approved):

sergio-teruel (approved):

PR #967, opened 2024-11-07
[16.0] [ADD] multisearch_field

TB-Ph35:
This module add multi-searching with separator for all search.

It add to model.search() code to enable searching with separator. 
example:
1.  you are searching "Tom|kelly;hart"
2. If the separator are "|" and ";"
3. it will search for "Tom|kelly;hart", "Tom|kelly", "Tom", "kelly;hart", "kelly" and "hart"


base_multi_search_separator/readme/USAGE.rst
```diff
@@ -0,0 +1,6 @@
+To configure the separator you must:
```
legalsylvain, line 1: better to put this part in a CONFIGURE.rst section.

In the USAGE file you can put a print screen. (something like "before" / "after").

base_multi_search_separator/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+* This module allows mulit_searching for all search.
```
legalsylvain: could you elaborate the description ? say exactly what the modules does.
(there are some base_search_xxx module in the OCA, like base_name_search_improved, base_search_fuzzy, etc...)

base_multi_search_separator/monkeypatching.py
```diff
@@ -0,0 +1,56 @@
...
+                value_list_set = value_list_set.union(dom[2].split(separator))
+            for value in value_list_set:
+                domain_search_field = (dom[0], dom[1], value)
+                nbr = nbr + 1
+                domain_value.append(domain_search_field)
+            nbr = nbr - 1
+        else:
+            domain_list.append(dom)
+    for _ in range(nbr):
+        domain_list.append("|")
```
legalsylvain: I think you can manipulate all the domain with expression.OR and expression.AND. It will be more readable I think.

base_multi_search_separator/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "MultiSearch",
```
legalsylvain: too generic title in my opinion.

legalsylvain (commented):
Nice work ! thanks for sharing. 

some minor remarks inline. Otherwise, LGTM.

bealdav:
Thanks @TB-Ph35 for this contribution.

Regarding the module name, it's not an easy topic, isn't it? 
If we want to make the name meaningful, we need to add meaningful terms and remove some. 
I guess `base_` doesn't help to understand what the module does. 

What the module does, if I understood correctly, is to allow searching for multiple values ​​of a particular field. 

`separator` is the way to do it, not what it does

To make this understandable with the module name, some candidates could be: 
- field_multisearch 
- multisearch_field 
- 

another suggestion?

legalsylvain:
FYI, my remarks regarding the name was about the value "name" in the manifest, not the technical name.

TB-Ph35:
Thanks for all the comment.

With those comment, i have made changes. But not the names module. It is not easy like you said, and that's really not my game. I will think about it.

base_multi_search_separator/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "Field MultiSearch with separotor",
```
legalsylvain:
```suggestion
    "name": "Field MultiSearch with separator",
```

typo.

PaulGoubert:
LGTM. Thank you for your work, Thomas.

I tried on the Customer Invoice by doing a search with:

Just one remark: once we want to copy and paste several lines from an XLS sheet, we don't have a separator, but only a space.
So in that case, we have to add the separator and remove the space before copying to make the search work.
Capture d’écran 2024-11-07 à 15 24 03
It could be very interesting to have the space consideration in the search, or even as a separator itself.

PaulGoubert:
Concerning the name, for my point of view, the main thing is to have multisearch in the module name. Then as @bealdav mentionned, field_multisearch or multisearch_field look good for me. 

If the Space is concidered as a "separator", the name separator is not required in the module name.

TB-Ph35:
I will look for adding the space. But now it can be add to the list of separator because for some search it broke the search. Like for self.env["resource.calendat.leaves"]. It split datetimeformat and that make this error : 

> psycopg2.errors.InvalidDatetimeFormat: invalid input syntax for type timestamp: "16:23:06"
LINE 1: ...')) OR ("resource_calendar_leaves"."date_from" <= '16:23:06'...

if i don't understand it wrong

PaulGoubert:
The search with space work fine now with / before the search. 
For product for ex : /E-COM06 E-COM07 E-COM08

Thank you @TB-Ph35

PaulGoubert (approved):
Functionnal review OK

multisearch_field/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "Field MultiSearch with separotor",
```
bealdav:
```suggestion
    "name": "Field MultiSearch with separator",
```

multisearch_field/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
+This module allows mulit_searching for all search.
+For that, it add code before the search function.
+
+It will allows you to search for multi_value separate by the character that you
```
bealdav:
```suggestion
It will allow you to search for multiple values ​​separated by the character you
```

multisearch_field/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
...
+have configure.
```
bealdav:
```suggestion
have configured.
```

multisearch_field/data/multi_search_data.xml
```diff
@@ -0,0 +1,6 @@
+<odoo noupdate="1">
+    <record id="multi_search_separator" model="ir.config_parameter">
+        <field name="key">multi_search_separator_base</field>
```
bealdav:
```suggestion
        <field name="key">multi_search_separator</field>
```

multisearch_field/monkeypatching.py
```diff
@@ -0,0 +1,54 @@
...
+
+search_original = BaseModel.search
+
+
+def search(self, domain, offset=0, limit=None, order=None, count=False):
+    """Override of the Python method to remove the dependency of the unit
+    fields"""
+    if self.env["ir.model.access"].check_access_rights("read", raise_exception=False):
+        list_separator = self.env["ir.config_parameter"].get_param(
+            "multi_search_separator_base"
```
bealdav:
```suggestion
            "multi_search_separator"
```

multisearch_field/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,6 @@
+To configure the separator you must:
+    * be in dev_mode
+    * go to Settings / technical / system parameter
+    * search multi_search_separator_base
```
bealdav:
```suggestion
    * search multi_search_separator
```

multisearch_field/tests/test_multi_search.py
```diff
@@ -0,0 +1,50 @@
+# Copyright 2024 Akretion (http://www.akretion.com).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import TransactionCase
+
+
+class TestMultisearch(TransactionCase):
+    def setUp(self):
+        super(TestMultisearch, self).setUp()
```
bealdav:
```suggestion
        super().setUp()
```

bealdav (commented):
Great job Thomas, really useful module, some typo comments here. 
Could you crop your images to focus on parts without white page color to see that you want to show.
Probably squash commits could be nice
Thanks a lot.

bealdav (approved):
Nice, thanks a lot

multisearch_field/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    "name": "Field MultiSearch with separator",
+    "license": "AGPL-3",
+    "summary": "add modification on search to search multi_value  with separator",
+    "version": "16.0.1.0.0",
+    "author": "Akretion, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/server-ux",
+    "depends": [
+        "base",
+    ],
+    "application": False,
```
bealdav, line 11: Not mandatory but it seems this key is useless whereas `maintainers` key can be useful

multisearch_field/monkeypatching.py
```diff
@@ -0,0 +1,54 @@
...
+search_original = BaseModel.search
+
+
+def search(self, domain, offset=0, limit=None, order=None, count=False):
+    """Override of the Python method to remove the dependency of the unit
+    fields"""
+    if self.check_access_rights("read"):
+        list_separator = (
+            self.env["ir.config_parameter"].sudo().get_param("multi_search_separator")
+        )
```
sebastienbeau: We should not use a monkeypatch but instead inheriting the base class

```
class Base(models.AbstractModel): 
    _inherit = "base"             
                                  
```

Also inheriting the search is broken (you can try on runboat to do a search and add a group by), you will see that this is broken

instead maybe inheriting the "_where_calc" will solve the issue

```
def _where_calc
```


@paradoxxxzero can you do a review ?

sebastienbeau (changes requested):
Need fixing as groupby is broken

multisearch_field/monkeypatching.py
TB-Ph35: thanks for the report and help.

I try some modification and if we modify the _where_calc instead of the search it resolve the groupby and work the search.

I will try to push modification as soon as possible

PaulGoubert (approved):

base_multi_search_separator/readme/USAGE.rst
bealdav: Finally done

bealdav (approved):
Just tested it works well

Image

bealdav:
@legalsylvain could you approve or review again ? Thanks

legalsylvain:
It's up to @sebastienbeau !

sebastienbeau (approved):
LGTM (code review)

PR #970, opened 2024-11-13
[16.0][IMP] base_tier_validation: support computed state field

antonioburic:
Computed fields bypass `write`, so we need to override `_write` for that case. Also, the current value before the update needs to be fetched from the database because the new value is already set in the cache.

backport of https://github.com/OCA/server-ux/pull/939


LoisRForgeFlow:
@antonioburic sorry I missed this one, you can open it again and ping me. I'll try to review and help to get it merged.

PR #977, opened 2024-12-02
[16.0][FIX] base_tier_validation: last comment instead of first comment

mathben:
- With multiple approbation, the first comment is repeated into mail.message and propagate wrong message


LoisRForgeFlow (changes requested):
Pending to confirm bug (see discussion https://github.com/OCA/server-ux/pull/978#issuecomment-2513068191)

PR #978, opened 2024-12-02
[17.0][FIX] base_tier_validation: last comment instead of first comment

mathben:
- With multiple approbation, the first comment is repeated into mail.message and propagate wrong message

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -575,7 +575,7 @@ class TierValidation(models.AbstractModel):
             lambda r: (self.env.user in r.reviewer_ids) and r.comment
         )
         if has_comment:
-            comment = has_comment.mapped("comment")[0]
+            comment = has_comment.mapped("comment")[-1]
             return _("A review was accepted. (%s)") % comment
         return _("A review was accepted")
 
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow:
@mathben could you provide a list of steps to reproduce the issue?

mathben:
1. Configuration
- Need a multiple tier.definition with configuration :
a) approve_sequence enable
b) has_comment enable
c) configure a definition domain to support a case where you need a user A and B

2. Test
The user A need to approve with a comment.
The issue occur when user B approve with a new comment. The comment is shown into validation comment, but the comment into mail.message use the comment of user A.

An example of the problem :
fix_message

With the fix, into mail.message, the User A will have the message "A review was accepted. (A new message, different!)"
fix_done

LoisRForgeFlow:
@mathben still unable to reproduce:

image

I have created the minimum case according to your description: 2 tier definitions in sequence and with coment, reviewed by different users. Comments are posted properly.

Can it be that it does not affect v17? I see you have created the PR in all versions but maybe only testing in one?

Kind regards,

mathben:
@LoisRForgeFlow 
> Can it be that it does not affect v17? I see you have created the PR in all versions but maybe only testing in one?

I only tested it on v14. I am happy to see the bug isn't in v17. I need to update my environment, I am not ready now to test it. We can close my pull request (v14 to v17), I will reopen it when I will be able to test it.
Thanks!

LoisRForgeFlow:
Sure @mathben feel free to reopen if you finally confirm the bug in any version.

PR #980, opened 2024-12-05
[18.0][FIX] base_technical_features: fix incorrect replacement

trisdoan:
### Context
- Sometimes group is negated like this: `groups="!base.group_no_one"`
- Without this fix, result of arch is `groups="!base.group_no_one,base_technical_features.group_technical_features"`, which is wrong


simahawk (approved):
Seems legit

PR #855, opened 2024-04-12
17.0 mig base substate

Wodran14:

PR #982, opened 2024-12-12
[17.0][FIX] base_substate: migrate track_template function

len-foss:
Note that because inheritance was poorly implemented in inheriting modules (such as sale_substate) this function was never called, hiding the bug.

auto_delete_message was removed in odoo@9140ce06

Note that the issue was reported before the migration: 
https://github.com/OCA/server-ux/pull/855#discussion_r1745694487

### DIFF
--- a/base_substate/models/base_substate_mixin.py
+++ b/base_substate/models/base_substate_mixin.py
@@ -28,19 +28,19 @@ class BaseSubstateMixin(models.AbstractModel):
                     }
                 )
 
-    def _track_template(self, tracking):
-        res = super()._track_template(tracking)
-        first_rec = self[0]
-        changes, tracking_value_ids = tracking[first_rec.id]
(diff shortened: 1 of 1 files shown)


StefanRijnhart (approved):
Thanks! We can see inhttps://github.com/OCA/sale-workflow/pull/3483 that it's working.

bosd:
Thanks, Can you add tests like in V15 version in #735

len-foss:
@bosd no, I don't think this test really makes sense as the test class is really too poor of a mock to be a good test. I would rather make a test_base_substate module, but sale_substate already exist.

StefanRijnhart:
@bosd Would you consider to approve this PR, so that we can move forward with https://github.com/OCA/sale-workflow/pull/3483?

bosd (approved):
LGTM

PR #966, opened 2024-11-01
[18.0][MIG] base_tier_validation: Migration to 18.0

kevinkhao:
supersedes https://github.com/OCA/server-ux/pull/964
superseded by https://github.com/OCA/server-ux/pull/992

PR #991, opened 2025-01-15
[17.0][IMP] base_tier_validation: Add help text to comment option

bosd:

PR #990, opened 2025-01-15
[18.0][MIG] base_tier_validation: Migration to 18.0

rjaraspearhead:
supersedes #966 

This PR enhances the menu icon and the display of items pending approval.

PR #992, opened 2025-01-16
[18.0][MIG] base_tier_validation: Migration to 18.0

xaviedoanhduy:
supersedes: https://github.com/OCA/server-ux/pull/966

changes are follow up from old PR:

- fix the icon on systray 
- fix the dymanic color alert on list reviewers
- fix of not recalculating reviewCount when updating review requests (`TierReviewService`). don't use useState to calculate tierReviewCounter because the service will update `services["mail.store"]`
- replaced old-style string interpolation with `f-string `and placeholder formatting
- remove old jquery and replace with javascript available objects
- optimize multi-company rule
- optimize using new API ORM `_read_group` for better performance 
- replace deprecated output directives (`t-esc`)


bosd:
Can you place your commits in a separate commit?
As there is also #990. Maybe we can cherry pick to merge these changes into there. If there is any difference.

xaviedoanhduy:
> Can you place your commits in a separate commit? As there is also #990. Maybe we can cherry pick to merge these changes into there. If there is any difference.

hi @bosd, that commit fixed the icon bug, but i consider that fix as part of PR Migration (because old PR is not yet merged). and author of this PR (https://github.com/OCA/server-ux/pull/990) also mentioned we have duplicated work, you can read the conversation here: https://github.com/xaviedoanhduy/server-ux/pull/1#issuecomment-2593101765

bosd:
Thanks for the clarification. I was getting a bit lost between the migration pr of this module.
Guess it was better to review this one.
Will do so when I have time. 

Side note, on this one codecov is red.

base_tier_validation/models/res_users.py
```diff
@@ -0,0 +1,53 @@
...
+        domain = [
+            ("status", "=", "pending"),
+            ("can_review", "=", True),
+            ("id", "in", user.review_ids.ids),
+        ]
+        review_groups = self.env["tier.review"].read_group(domain, ["model"], ["model"])
+        for review_group in review_groups:
+            model = review_group["model"]
+            Model = self.env[model]
+            reviews = self.env["tier.review"].search(review_group.get("__domain"))
```
ivantodorovich: 💡 tip: This could probably be improved by leveraging the private `_read_group` method, and aggregating `['id:recordset']`, and thus not needing to do a search afterwards

base_tier_validation/models/tier_review.py
```diff
@@ -141,7 +135,10 @@ def _compute_todo_by(self):
         for rec in self:
             todo_by = False
             if rec.reviewer_group_id:
-                todo_by = _("Group %s") % rec.reviewer_group_id.name
+                todo_by = self.env._(
+                    "Group %(reviewer_group_name)s",
+                    reviewer_group_name=rec.reviewer_group_id.name,
```
ivantodorovich: I'd keep the original translatable term here, to not invalidate existing translations.
There's really no benefit in using kword arguments when there's only one param

```py
todo_by = this.env._("Group %s", rec.reviewer_group_id.name)
```

base_tier_validation/models/tier_review.py
```diff
@@ -173,7 +172,9 @@ def _get_reminder_activity_type(self):
 
     def _notify_review_reminder_body(self):
         delay = (fields.Datetime.now() - self.create_date).days
-        return _("A review has been requested %s days ago.") % (delay)
+        return self.env._(
+            "A review has been requested %(delay)s days ago.", delay=delay
```
ivantodorovich: Same as previous comment.
Since there's only 1 param, IMO there's no benefit in using a kword argument and it's better to keep the original translate term

base_tier_validation/models/tier_validation.py
```diff
@@ -169,21 +170,20 @@ def _get_to_validate_message_name(self):
         return self._description
 
     def _get_to_validate_message(self):
-        return (
-            """<i class="fa fa-info-circle" /> %s"""
-            % _("This %s needs to be validated")
-            % self._get_to_validate_message_name()
+        msg_fragment = self.env._(
+            f"This {self._get_to_validate_message_name()} " f"needs to be validated"
```
ivantodorovich: ❌ This change is not correct.

The translatable term can't be dynamic, otherwise translations aren't possible.
Please keep the previous implementation, or adapt to fix it 🙏🏻

base_tier_validation/models/tier_validation.py
```diff
@@ -221,7 +221,9 @@ def _compute_next_review(self):
             review = rec.review_ids.sorted("sequence").filtered(
                 lambda x: x.status == "pending"
             )[:1]
-            rec.next_review = review and _("Next: %s") % review.name or ""
+            rec.next_review = review and self.env._(
+                "Next: %(review_name)s", review_name=review.name or ""
+            )
```
ivantodorovich: Same as previous comments. It's a shame to invalidate existing translatable terms just to use a kword argument on a term with only 1 arg

base_tier_validation/models/tier_validation.py
```diff
@@ -406,17 +409,19 @@ def _tier_validation_check_state_on_write(self, vals):
...
                         raise ValidationError(
-                            _(
+                            self.env._(
                                 "This action needs to be validated for at least "
-                                "one record. Reviews pending:\n - %s "
-                                "\nPlease request a validation."
+                                "one record. Reviews pending:\n "
+                                "- %(pending_reviews_str)s "
+                                "\nPlease request a validation.",
+                                pending_reviews_str=pending_reviews_str,
```
ivantodorovich: Same as previous comments. It's a shame to invalidate existing translatable terms just to use a kword argument on a term with only 1 arg

base_tier_validation/models/tier_validation.py
```diff
@@ -576,13 +581,13 @@ def _notify_accepted_reviews_body(self):
         )
         if has_comment:
             comment = has_comment.mapped("comment")[0]
-            return _("A review was accepted. (%s)") % comment
-        return _("A review was accepted")
+            return self.env._("A review was accepted. (%(comment)s)", comment=comment)
```
ivantodorovich: Same as previous comments. It's a shame to invalidate existing translatable terms just to use a kword argument on a term with only 1 arg

base_tier_validation/models/tier_validation.py
```diff
@@ -626,11 +631,14 @@ def _notify_rejected_review_body(self):
...
-            }
-        return _("A review was rejected by %s.") % (self.env.user.name)
+            return self.env._(
+                "A review was rejected by %(user)s. (%(comment)s)",
+                user=self.env.user.name,
+                comment=comment,
+            )
+        return self.env._(
+            "A review was rejected by %(user_name)s.", user_name=self.env.user.name
+        )
```
ivantodorovich: Same as previous comments. It's a shame to invalidate existing translatable terms just to use a kword argument on a term with only 1 arg

base_tier_validation/models/tier_validation.py
```diff
@@ -672,12 +680,16 @@ def _rejected_tier(self, tiers=False):
                 rec._notify_rejected_review()
 
     def _notify_created_review_body(self):
-        return _("A record to be reviewed has been created by %s.") % (
-            self.env.user.name
+        return self.env._(
+            "A record to be reviewed has been created by %(user_name)s.",
+            user_name=self.env.user.name,
```
ivantodorovich: Same as previous comments. It's a shame to invalidate existing translatable terms just to use a kword argument on a term with only 1 arg

base_tier_validation/models/tier_validation.py
```diff
@@ -672,12 +680,16 @@ def _rejected_tier(self, tiers=False):
...
         )
 
     def _notify_requested_review_body(self):
-        return _("A review has been requested by %s.") % (self.env.user.name)
+        return self.env._(
+            "A review has been requested by %(user_name)s.",
+            user_name=self.env.user.name,
```
ivantodorovich: Same as previous comments. It's a shame to invalidate existing translatable terms just to use a kword argument on a term with only 1 arg

base_tier_validation/models/tier_validation.py
```diff
@@ -732,7 +744,9 @@ def request_validation(self):
         return created_trs
 
     def _notify_restarted_review_body(self):
-        return _("The review has been reset by %s.") % (self.env.user.name)
+        return self.env._(
+            "The review has been reset by %(user_name)s.", user_name=self.env.user.name
+        )
```
ivantodorovich: Same as previous comments. It's a shame to invalidate existing translatable terms just to use a kword argument on a term with only 1 arg

base_tier_validation/static/src/components/tier_review_widget/tier_review_widget.esm.js
```diff
@@ -1,33 +1,35 @@
...
+        return records.map((record) => record.data);
     }
+
     onToggleCollapse(ev) {
-        var $panelHeading = $(ev.currentTarget).closest(".panel-heading");
-        if (this.collapse) {
-            $panelHeading.next("div#collapse1").hide();
+        const panelHeading = ev.currentTarget.closest(".panel-heading");
+        if (!panelHeading) return;
+        const collapseDiv = document.getElementById("collapse1");
```
ivantodorovich: This doesn't seem right. Previous implementation wasn't looking into the global `document`, it was looking at the immediate next sibling of the parentHeading.

I think you need to do something like this (not tested):

```js
const collapseDiv = panelHeading.nextElementSibling.matches("div#collapse1") ? panelHeading.nextElementSibling : null;
```

ivantodorovich (changes requested):
Hey, thanks for taking care of this!

There are a few things that need to be corrected, IMO.
Could you check my remarks? 🙏🏻 

Thanks!

miikanissi (approved):
Functional test looks good to me - no issues found

base_tier_validation/static/src/components/tier_review_widget/tier_review_widget.esm.js
xaviedoanhduy: both achieve the same purpose. I use `document` to replace old `jquery` and using it to directly query the `Id` of an element will improve Accessing the DOM performance.

base_tier_validation/models/res_users.py
bosd: @ivantodorovich Good spot. that was on my mind as well.
There are some more places in the code of this module where the orm is flooded by searches from within a for loop.

The `_read_group` method was introduced in the orm in V17.
IMO that improvement could go into V17 and be forward ported here. Should not be blocking for this pr.

(But I've been told, there is even a more improved method in the V18 orm. But I don't know the details or have looked into that one.)

bosd (changes requested):
Thanks for your efforts and work on this: :+1: 

Functional test:

Colors are not showing in review bar as mentioned by @celm1990 in https://github.com/OCA/server-ux/pull/990#pullrequestreview-2554239509

base_tier_validation/static/src/components/tier_review_widget/tier_review_widget.esm.js
ivantodorovich: Well. Yes and no.
Strictly speaking, the new implementation try to find the `#collapse1` element on the whole page.
Whilst the previous implementation (and the one I proposed written in vanillajs), will limit the scope to the immediate next/sibling element.

base_tier_validation/models/res_users.py
ivantodorovich: I think `:recordset` was added in 18.0.
In 17.0 a similar thing can be accomplished with `:array_agg`, and then browsing the list of ids.

ivantodorovich (approved):

base_tier_validation/static/src/components/tier_review_widget/tier_review_widget.esm.js
xaviedoanhduy: thanks, good information and i will try it in next update

xaviedoanhduy:
> Thanks for your efforts and work on this: 👍
> 
> Functional test:
> 
> Colors are not showing in review bar as mentioned by @celm1990 in https://github.com/OCA/server-ux/pull/990#pullrequestreview-2554239509

hi, color issue on reviewers list has been fixed. thanks for the tests

bosd (approved):
Funcitonal test. LGTM :+1: 

Non-blocking / nitpicking.
JS error/warning
image

UI can be improved a bit.
image

core:
image

xaviedoanhduy:
> Funcitonal test. LGTM 👍
> 
> Non-blocking / nitpicking. JS error/warning image

base_tier_validation/models/res_users.py
xaviedoanhduy: FYI, `_read_group` and `ir.recordset` already exist in version 17.0, see: https://github.com/odoo/odoo/pull/110737

and by the way I also applied the above code improvement from your suggestion, thank @ivantodorovich

bosd:
Thanks, If you want, you can include https://github.com/OCA/server-ux/pull/991

base_tier_validation/models/res_users.py
bosd: @xaviedoanhduy Thanks!
Actually there are still some places in the code where a search is performed inside a for loop. 
(Bad practice for performance.)
https://github.com/OCA/server-ux/pull/992/files#diff-7cca2e2b43a392c79d13759687633b9d88cfe56e6c2238b762e2c8d461690607R123-R124

Maybe you can attend to this, or put it on the roadmap.

ivantodorovich: > FYI, _read_group and ir.recordset already exist in version 17.0, see: https://github.com/odoo/odoo/pull/110737

Oh, indeed. Thanks!

base_tier_validation/static/src/components/tier_review_widget/tier_review_widget.esm.js
```diff
@@ -0,0 +1,41 @@
...
+            ? panelHeading.nextElementSibling
+            : null;
+        if (!collapseDiv) return;
+        this.state.collapse = !this.state.collapse;
+        if (this.state.collapse) {
+            collapseDiv.style.display = "none";
+        } else {
+            collapseDiv.style.display = "block";
+        }
+    }
```
celm1990: The first time you click the widget, it does not open; a second click is needed to collapse it. Please take a look. On the second attempt, it works fine, but if you refresh the view again, it doesn't work. I think the initial state might be related.

celm1990 (changes requested):
LGTM, just a minor change and let's go.
Additionally, please squash the administrative commits a bit. Thanks for the effort!

base_tier_validation/static/src/components/tier_review_widget/tier_review_widget.esm.js
celm1990: @xaviedoanhduy could you please check this

xaviedoanhduy: hi @celm1990, in my country it is Lunar New Year. I will start working again on 02/03/2025. Thank you for raising the issue and btw happy Lunar New Year 2025

bosd: Happy lunar year! The year of intelligence and love

xaviedoanhduy: hi @bosd and @celm1990, the issue of clicking twice on the widget when loading the page for the first time has been resolved

base_tier_validation/models/res_users.py
xaviedoanhduy: hello, I've improved some search functions that are called in several loops into a separate commit

bosd (approved):

celm1990 (approved):
Tested locally, LGTM.

Just a non-blocking comment: I think it's possible to squash the administrative commits a bit, following this guide: [Merge commits in pull requests](https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests).

Thanks for the effort!

ping @sebalix @LoisRForgeFlow @pedrobaeza

StefanRijnhart (approved):
Thanks everyone, great work here on the review and the follow up!

StefanRijnhart:
Merge failed on https://github.com/OCA/oca-addons-repo-template/pull/291. Doh!

StefanRijnhart:
Let's try again after this hour long scheduled brownout

/ocabot merge nobump

bosd:
Thanks for the team work on this migration!

Is any of you using this one already in production?
Have you encountered the following bugs in V18:
- After validation of a record, it is impossible to duplicate (An error thrown, regression in V18) 
- Unable to send email after record validation as discovered in https://github.com/OCA/account-invoicing/issues/1998

miikanissi:
> Thanks for the team work on this migration!
> 
> Is any of you using this one already in production? Have you encountered the following bugs in V18:
> 
> * After validation of a record, it is impossible to duplicate (An error thrown, regression in V18)
> * Unable to send email after record validation as discovered in [[18.0][MIG] account_move_tier_validation: Unable to mail invoice account-invoicing#1998](https://github.com/OCA/account-invoicing/issues/1998)

We are using this in production. We have noticed a few issues. 

* Some computed and/or related fields that change after validation can trigger an under validation exception. This can also happen if some fields are written by an action. This sounds like the same issue as the unable to send email issue. Essentially, nothing can change on the record after validation. We have fixed this by just adding a ton of fields in the validation exceptions rules. 
* Readonly fields configured in the field definition are editable. For example, the `date_approve` field on a Purchase Order is a readonly field in the field definition. But after setting up validations the field will be editable. We had to manually make a view level change to set readonly="1" on the affected fields. TLDR; Readonly Field definition is ignored.

LoisRForgeFlow:
> > Thanks for the team work on this migration!
> > Is any of you using this one already in production? Have you encountered the following bugs in V18:
> > 
> > * After validation of a record, it is impossible to duplicate (An error thrown, regression in V18)
> > * Unable to send email after record validation as discovered in [[18.0][MIG] account_move_tier_validation: Unable to mail invoice account-invoicing#1998](https://github.com/OCA/account-invoicing/issues/1998)
> 
> We are using this in production. We have noticed a few issues.
> 
> * Some computed and/or related fields that change after validation can trigger an under validation exception. This can also happen if some fields are written by an action. This sounds like the same issue as the unable to send email issue. Essentially, nothing can change on the record after validation. We have fixed this by just adding a ton of fields in the validation exceptions rules.
> * Readonly fields configured in the field definition are editable. For example, the `date_approve` field on a Purchase Order is a readonly field in the field definition. But after setting up validations the field will be editable. We had to manually make a view level change to set readonly="1" on the affected fields. TLDR; Readonly Field definition is ignored.

@miikanissi Thanks for the heads-up.

It looks like there is not feedback from the migrator and reviewers in v18. If your comments are still applicable, I would kindly ask you to create issues in this repository to track the problem. Otherwise this comment will be forgotten here.

PR #996, opened 2025-01-21
[16.0][FIX] base_technical_features: Add depends key in __manifest__.py

tarteo:
Without this the module will not be listed by upstream_dependencies https://github.com/odoo/odoo/blob/05ce1a521d9702ce490100354d2b19007c35bb62/odoo/addons/base/models/ir_module.py#L537

PR #998, opened 2025-01-22
[17.0][FIX] base_technical_features: Add depends key in __manifest__.py

tarteo:
Same as https://github.com/OCA/server-ux/pull/996


pedrobaeza (approved):
/ocabot merge patch

PR #1004, opened 2025-01-24
[FIX] sequence_check_digit: Constrain should work with multiple records

etobella:


luisDIXMIT (approved):
LGTM!

etobella:
Fast tracking as this can be a problem when generating multiple sequences

/ocabot merge patch

PR #1009, opened 2025-01-29
base_export_manager: js code was not migrated. 

sebastienbeau:
The current version of the js code was not migrated, so the export button was always visible

You can see the following error message in log when opening a tree view
```
Missing dependencies:  Array [ "web.Sidebar" ]
Non loaded modules: Array [ "base_export_manager.base_export_manager" ]
```

Backport it from v15.0 Thanks to @nicomacr


sebastienbeau:
@Kev-Roche @paradoxxxzero

FrancoMaxime (approved):
LGTM: code review

PR #1012, opened 2025-01-30
[16.0][FIX] base_custom_filter: fix access right error

AungKoKoLin1997:
This PR fixes the access rights error that occurs when creating a custom filter to add a new groupby for a specific model, and when a user without the settings access rights tries to open the specific module's tree view.

@qrtl QT5153


yostashiro (approved):
LGTM. Without this fix, non-admin users would encounter an access error if the model being accessed has a custom group-by record.
image

kanda999 (approved):
Functional check

StefanRijnhart (approved):

PR #1017, opened 2025-02-03
[18.0][MIG] base_search_custom_field_filter: Migration to 18.0

absal-smile:


absal-smile:
[pre-commit](https://github.com/OCA/server-ux/actions/workflows/pre-commit.yml)

absal-smile:
@pedrobaeza

pedrobaeza:
@absal-smile you should require to cla@odoo-community.org for being included in the Smile ICLA for not having every workflow to be approved.

base_search_custom_field_filter/models/ir_ui_custom_field_filter.py
```diff
@@ -35,16 +34,30 @@ def _get_related_field(self):
...
+            if self.search(
+                [
+                    ("model_id", "=", vals.get("model_id")),
+                    ("name", "=", vals.get("name")),
+                ]
+            ):
+                raise exceptions.ValidationError(
+                    _("A filter with the same name already exists for this model.")
+                )
+        return super().create(vals_list)
```
carlos-lopez-tecnativa: If the intention is to avoid duplicates, I think using an `unique sql_constraints` would be a better approach.

carlos-lopez-tecnativa (commented):
Tested and working fine. I think some administrative commits can be squashed. Please let me know what you think about using an `sql_constraint` instead, as it would also apply on write (unlike create, which does not take it into account).

TT54421

base_search_custom_field_filter/models/ir_ui_custom_field_filter.py
absal-smile: I think if I replace it with sql constraint I have to remove the translation of the name, and I don't know if it will have any other impact on other fields

carlos-lopez-tecnativa: Ok, I understand that doing it by `SQL` is not working correctly because the field `name` is `translated`. So, please add the check in the `write` method too, not only in `create`. If I create a record with a different name and later edit this record and set a name exactly equal to another record, it should be forbidden, and currently, this case is not controlled

image

carlos-lopez-tecnativa (changes requested):

base_search_custom_field_filter/tests/test_filter.py
```diff
@@ -0,0 +1,123 @@
...
+        filter_form.name = "Valid Expression"
+        filter_form.expression = "name"
+        filter_form.save()
+
+        arch = self.model.get_view(False, "search")["arch"]
+        search = etree.fromstring(arch)
+        self.assertTrue(search.xpath("//search/field[@name='name']"))
+
+
+def test_duplicate_filter(self):
```
carlos-lopez-tecnativa: Please check the indentation of this function.

carlos-lopez-tecnativa (changes requested):

carlos-lopez-tecnativa (approved):
LGTM
@pedrobaeza @victoralmau could you please review this?

base_search_custom_field_filter/models/base.py
```diff
@@ -49,15 +49,22 @@ def get_view(self, view_id=None, view_type="form", **options):
     def get_views(self, views, options=None):
         """Inject fake field definition for having custom filters available."""
         res = super().get_views(views, options)
+
```
pedrobaeza: Please don't insert empty lines inside the method (the same for all the occurrences).

base_search_custom_field_filter/models/base.py
```diff
@@ -49,15 +49,22 @@ def get_view(self, view_id=None, view_type="form", **options):
...
         for custom_filter in custom_filters:
             field = custom_filter._get_related_field()
-            field_name = custom_filter.expression
-            res["models"][self._name][field_name] = field.get_description(self.env)
-            # force this for avoiding to appear on the rest of the UI
-            res["models"][self._name][field_name]["selectable"] = False
-            res["models"][self._name][field_name]["sortable"] = False
-            res["models"][self._name][field_name]["store"] = False
+            if field:
```
pedrobaeza: For reducing the diff, do the negative one `if not field: continue`. Anyway, why do you need this safeguard? Please document it through a code comment.

base_search_custom_field_filter/models/ir_ui_custom_field_filter.py
pedrobaeza: The current implementation is not correct. `vals["name"]` may be something like `{"en_US": "Test"}`, so it won't match. It's better to implement the SQL constraint.

base_search_custom_field_filter/models/ir_ui_custom_field_filter.py
```diff
@@ -35,16 +34,47 @@ def _get_related_field(self):
...
         for record in self:
             try:
                 record._get_related_field()
             except KeyError as e:
                 raise exceptions.ValidationError(
-                    _("Incorrect expression: %s.") % (ustr(e))
+                    _("Incorrect expression: %s.") % (str(e))
                 ) from e
+
+    def write(self, vals):
```
pedrobaeza: It has to be moved to SQL constraint as said.

base_search_custom_field_filter/tests/test_filter.py
```diff
@@ -2,31 +2,119 @@
...
+        cls.custom_filter_model = cls.env["ir.ui.custom.field.filter"]
 
-    def test_1(self):
-        filter_form = Form(self.env["ir.ui.custom.field.filter"])
+    def test_00(self):
+        """Test de validation des expressions de filtre"""
+        filter_form = Form(self.custom_filter_model)
         filter_form.model_id = self.env.ref("base.model_res_partner")
         filter_form.name = "Title"
+
```
pedrobaeza, line 23: Don't add empty lines inside the method (the same for all the occurrences).

pedrobaeza (changes requested):

carlos-lopez-tecnativa:
ping @absal-smile

absal-smile:
ping @carlos-lopez-tecnativa

base_search_custom_field_filter/models/ir_ui_custom_field_filter.py
```diff
@@ -0,0 +1,95 @@
...
+        string="Model name",
+    )
+    name = fields.Char(required=True, translate=True)
+    name_normalized = fields.Char(
+        compute="_compute_name_normalized",
+        store=True,
+        readonly=True,
+        index=True,
+        help="Normalized name used for uniqueness validation.",
+    )
```
carlos-lopez-tecnativa: I'm not sure this is the best solution, adding a new field based on another translatable field.
Please keep this simple: just add the SQL constraint on the `name` field. What is the problem with this approach?

carlos-lopez-tecnativa (changes requested):

base_search_custom_field_filter/models/ir_ui_custom_field_filter.py
carlos-lopez-tecnativa: ping @absal-smile

carlos-lopez-tecnativa (approved):
LGTM

@pedrobaeza @victoralmau could you please review this?

base_search_custom_field_filter/models/ir_ui_custom_field_filter.py
```diff
@@ -36,7 +36,6 @@ class IrUiCustomFilter(models.Model):
     )
 
     def _get_related_field(self):
-        """Determine the chain of fields."""
```
pedrobaeza: Why are you removing these docstrings?

pedrobaeza: I mean all the docstrings, not just this one.

pedrobaeza (approved):
/ocabot merge nobump

PR #1023, opened 2025-02-05
[18.0][FIX] base_tier_validation: review table key undefined

miikanissi:
The following OwlError is received when multiple tier reviews are present:

```OwlError: Got duplicate key in t-foreach: undefined```

This happens due to `review.id` being undefined as it is not passed to the
template. We should use `review.sequence` instead as it gets passed and
is unique.


ryanbwildpackbev (approved):
LGTM. Functional testing done

LoisRForgeFlow (approved):
Thanks for the fix

PR #1024, opened 2025-02-10
[17.0][FIX] base_tier_validation - fix 'false positive' raise

luc-demeyer:
base_tier_validation/models/tier_validation.py may raise an exception in the _tier_validation_check_write_allowed() method when there is no change (when vals[fld] == rec[fld]).
I encountered this case while using the account_move_tier_validation module and posting a number of validated supplier invoices via the list view (Action -> Post entries).
In this case the account.move sequence mixin generates a write on the sequence_prefix field but vals["sequence_prefix"] is equal to rec.sequence_prefix hence this is not 'real' write and we should not prevent the posting in such a case (when Confirming these same invoices via the form view no raise is generated).
The solution is to limit raise to writes that do modify the field value.

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -343,14 +343,14 @@ class TierValidation(models.AbstractModel):
         if not not_allowed_fields:
             return []
 
-        not_allowed_field_names, allowed_field_names = [], []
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (changes requested):
@luc-demeyer could you add a test case that covers the case to avoid future regressions?

PR #1034, opened 2025-03-06
[17.0][FIX] base_tier_validation - wrong order of reviewer when click…

DucTruongKomit:
**Current Behavior**
the sequence order of review on Request Validation button be always reverted.

For example:
from Tier Definition, creating
 On Purchase Request, User A will be set as sequence 1 (this person review 1st).
  On Purchase Request, User B will be set as sequence 2 (this person review 2nd).

After clicking Request Validation button, User B always be asked to reviewed 1st.

**Expected Behavior**
Reviewing should be following the order of sequence.

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -719,7 +719,7 @@ class TierValidation(models.AbstractModel):
                         ("model", "=", self._name),
                         ("company_id", "in", [False] + self.env.company.ids),
                     ],
-                    order="sequence desc",
+                    order="sequence",
                 )
                 sequence = 0
                 for td in tier_definitions:


LoisRForgeFlow (changes requested):
Hi @DucTruongKomit 

This makes sense and it is an issue that needed to be addresed. However, to be able to merge this, you will need to provide migration scripts that reverts all the sequences in existing tier definitions to not change behavior in existing installations. Could you do that?

Thanks for contributing!

DucTruongKomit:
> Hi @DucTruongKomit
> 
> This makes sense and it is an issue that needed to be addresed. However, to be able to merge this, you will need to provide migration scripts that reverts all the sequences in existing tier definitions to not change behavior in existing installations. Could you do that?
> 
> Thanks for contributing!

I will try to arrange time and work on it.

PR #1038, opened 2025-03-10
[16.0][FIX] base_tier_validation: Do not update the counter if it is not possible to review it.

victoralmau:
Do not update the counter if it is not possible to review it.

Example use case:
- Create tier definition with Mitchell Admin as reviewer user.
- Create a record (`purchase.requisition` for example) with Marc Demo and request a validation.
- Do not update the counter in Marc Demo (because you cannot review it).

**Before**
antes-demo

**After**
despues-demo
despues-admin

Please @pedrobaeza and @sergio-teruel can you review it?

@Tecnativa TT55411

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -655,8 +655,9 @@ class TierValidation(models.AbstractModel):
                     if rec.evaluate_tier(td):
                         sequence += 1
                         vals_list.append(rec._prepare_tier_review_vals(td, sequence))
-                self._update_counter({"review_created": True})
         created_trs = tr_obj.create(vals_list)
+        if any(self.mapped("can_review")):
+            self._update_counter({"review_created": True})
         self._notify_review_requested(created_trs)
         return created_trs
 
--- a/base_tier_validation/tests/test_tier_validation.py
+++ b/base_tier_validation/tests/test_tier_validation.py
@@ -127,11 +127,8 @@ class TierTierValidation(CommonTierValidation):
         )
         # Request validation
         self.test_record.with_user(self.test_user_2.id).request_validation()
-        self.test_record.invalidate_model()
         test_record.with_user(self.test_user_2.id).request_validation()
-        test_record.invalidate_model()
         self.test_record_2.with_user(self.test_user_2.id).request_validation()
-        self.test_record_2.invalidate_model()
         # Get review user count as systray icon would do and check count value
         docs = self.test_user_1.with_user(self.test_user_1).review_user_count()
         for doc in docs:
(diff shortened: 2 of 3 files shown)


LoisRForgeFlow (approved):
Makes sense, Thanks :+1: 

I assume this issue also happens in v17 and v18?

base_tier_validation/models/tier_validation.py
```diff
@@ -655,8 +655,10 @@ def request_validation(self):
                     if rec.evaluate_tier(td):
                         sequence += 1
                         vals_list.append(rec._prepare_tier_review_vals(td, sequence))
-                self._update_counter({"review_created": True})
         created_trs = tr_obj.create(vals_list)
+        self._invalidate_cache(["review_ids"])
+        if self.can_review:
```
pedrobaeza: Is this multi-record aware?

base_tier_validation/models/tier_validation.py
```diff
@@ -655,8 +655,10 @@ def request_validation(self):
                     if rec.evaluate_tier(td):
                         sequence += 1
                         vals_list.append(rec._prepare_tier_review_vals(td, sequence))
-                self._update_counter({"review_created": True})
         created_trs = tr_obj.create(vals_list)
+        self._invalidate_cache(["review_ids"])
+        if any(item.can_review for item in self):
```
pedrobaeza: Let's do it a bit more elegant:

```suggestion
        if any(item.mapped("can_review")):
```

base_tier_validation/tests/test_tier_validation.py
```diff
@@ -941,7 +938,7 @@ def test_27_reevaluate_validation(self):
         reviews = test_record.with_user(self.test_user_2.id).reevaluate_reviews()
         # Check need validation
         self.assertTrue(test_record.need_validation)
-        self.assertEqual(len(reviews), 2)
+        self.assertEqual(len(reviews), 1)
```
pedrobaeza, line 944: Are these changes related to not calling `_update_counter`. If so, the name is misleading...

victoralmau: No, all these changes are related to using `_invalidate_cache`, it is something I don't like, but it is related to the `review_ids` field being a One2many and the inverse_name to which it refers (`res_id`) being an integer field and not a many2one, that would simplify things (we have already discussed that casuistry in some occasion), but I don't want to open that debate now.

pedrobaeza: Uhm, isn't `Many2oneReference` field type for that?

victoralmau: Ok, thanks, changed. The test changes are still needed, although I have added some validations that help to understand it better.

sergio-teruel (approved):

base_tier_validation/models/tier_validation.py
```diff
@@ -34,7 +34,6 @@ class TierValidation(models.AbstractModel):
         inverse_name="res_id",
         string="Validations",
         domain=lambda self: [("model", "=", self._name)],
-        auto_join=True,
```
pedrobaeza: Why removing this?

victoralmau: I have understood that it is not necessary when changing the res_id field to fields.Many2oneReference (I have not seen examples in odoo core where in a one2many define auto_join). 

Do you think it is necessary to keep it?

pedrobaeza: This is done for performance reason and it's not related with the change of the field type. With this sentence, you allow the ORM to join in one SQL fetchings of both tables without caring about the specific security rules of the children table, which is correct in this case AFAIK.

pedrobaeza (approved):
/ocabot merge patch

PR #1043, opened 2025-03-14
[16.0][IMP] base_tier_validation: Add subscription mode in tier definition

AlexPForgeFlow:
Sometimes we may use Tier Validation for records that are communicating with the customer, and we don't want validators to receive notifications for the entire discussion. We introduce a new field 'Subscription mode' that allows to configure if we only want to receive Tier Validation notifications.

Original: https://github.com/OCA/server-ux/pull/911


LoisRForgeFlow:
@AlexPForgeFlow I think this improvement https://github.com/OCA/server-ux/pull/1083 it is a better approach for the underlying issue.

PR #1045, opened 2025-03-17
[18.0][FIX] base_tier_validation: request tier with multi-company

Saran440:
Document belongs to `YourCompany`, but user requesting validation has multiple companies and has selected `MyCompany`. I think request should use the tier from document's company (YourCompany) instead of the currently selected company.

Steps to Reproduce (with Purchase):
1. Create purchase order -> it will default to `YourCompany`
2. The user has multi-company and the current selected company is `MyCompany`
3. User clicks Request Validation > It uses the tier from `MyCompany` (but it should use tier from `YourCompany`)
Selection_002

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -22,6 +22,7 @@ class TierValidation(models.AbstractModel):
     _tier_validation_buttons_xpath = "/form/header/button[last()]"
     _tier_validation_manual_config = True
     _tier_validation_state_field_is_computed = False
+    _tier_validation_company_field = "company_id"
 
     _state_field = "state"
     _state_from = ["draft"]
@@ -250,7 +251,7 @@ class TierValidation(models.AbstractModel):
                 .search(
                     [
                         ("model", "=", self._name),
-                        ("company_id", "in", [False] + self.env.company.ids),
+                        ("company_id", "in", [False] + rec._get_company().ids),
                     ]
                 )
             )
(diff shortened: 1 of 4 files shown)


base_tier_validation/models/tier_validation.py
```diff
@@ -720,7 +720,7 @@ def request_validation(self):
                 tier_definitions = td_obj.search(
                     [
                         ("model", "=", self._name),
-                        ("company_id", "in", [False] + self.env.company.ids),
+                        ("company_id", "in", [False] + rec.company_id.ids),
```
carlos-lopez-tecnativa: The problem I see is that the models inheriting from this class are not guaranteed to have a `company_id` field (it might have a different name or not exist at all).

Additionally, you only changed it in this function, but I noticed that `self.env.company` is used in many other places. You should adapt the logic everywhere to maintain consistency.

One idea: You could create a separate function to retrieve the company and use it here. This function could then be easily inherited in other modules/models where `company_id` has a different name. What do you think?

LoisRForgeFlow: I agree with @carlos-lopez-tecnativa

Saran440: @carlos-lopez-tecnativa @LoisRForgeFlow 
Thank you for your review.
I have added the `_tier_validation_company_field` class attribute so that other modules can inherit it.

Additionally, I have split it into a new method to allow for hooks.

LoisRForgeFlow (commented):
Thanks! Could you add a tests case? you could add company_id to tier.validation.tester2 and keep the other without company, and test that everything works as expected in both cases

LoisRForgeFlow (approved):
Thanks! LGTM :+1:

carlos-lopez-tecnativa (approved):
LGTM

carlos-lopez-tecnativa:
@Saran440 Could you please backport these changes to the older versions? I believe they can be applied from V13 to V17

PR #1054, opened 2025-03-27
[18.0][FIX] base_tier_validation: Add filter to check if review_ids is approved when writing on an approved record

TomVermeerenDynapps:


natuan9 (approved):

LoisRForgeFlow (changes requested):
Can you provide steps to reproduce the issue you are trying to solve? I'm not sure if I'm understand what you are fixing.

TomVermeerenDynapps:
> Can you provide steps to reproduce the issue you are trying to solve? I'm not sure if I'm understand what you are fixing.
@LoisRForgeFlow 
When a record is validated, and (in this example) the state is changed to "confirmed", this error pops up because we're technically not allowed to write. Adding this code makes it possible to confirm when all the reviews are approved.
image

LoisRForgeFlow:
Hi @TomVermeerenDynapps 

Sorry for the delay on the response. I have quickly tried using sale_tier_validation (it looks like this is the one you are using based on your screenshot) but without luck. Everythings works as expected.

Could you be more specific in the steps to reproduce the error? It should be a step-by-step guide to reproduce the error (e.g.: create a tier definition like this, configure user with the following user rights, etc) because with basic demo data it worked for me.

Thanks for contributing!

Kimkhoi3010:
Hi @TomVermeerenDynapps ,
Thanks for the screenshot and context.

I’ve tried to reproduce the issue on my side by following the general steps, but so far I haven’t been able to trigger the same error. I’ve also installed the sale_subscription module and tested up to the “Request Validation” step, but everything seems to be working as expected.

It would be helpful if you could share more details about the tier validation configuration you’re using (as @LoisRForgeFlow also mentioned earlier).

Thanks in advance! That would really help in understanding the root cause.

PR #1055, opened 2025-03-27
[17.0] [FIX] base_tier_validation: Fixed readonly behavior of fields.

NachoAlesLopez:
The module `base_tier_validation` behaves strangely when computed fields are present. For example, the following screenshots come from a base installation of Odoo with the `sale_tier_validation` and `sale_purchase` module added, which causes the purchase order count to be editable when inside a smartbutton:

image

Or, when paired with the `sale_margin` module, it allows the editing of the margin when the field should be read-only:

image

This is because the base module doesn't take into account the read-onliness of the fields, especially computed fields. This PR should fix that behavior and add a test case to ensure the cases shown before remain patched.

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -847,7 +847,9 @@ class TierValidation(models.AbstractModel):
                 if node.attrib.get("name") in excepted_fields:
                     continue
                 new_r_modifier = self._get_tier_validation_readonly_domain()
-                old_r_modifier = node.attrib.get("readonly")
+                old_r_modifier = node.attrib.get(
+                    "readonly", self._fields.get(node.attrib["name"]).readonly
+                )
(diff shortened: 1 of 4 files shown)


LoisRForgeFlow (changes requested):
@NachoAlesLopez Sorry for the slow review

I cannot reproduce the error and I also see that the PR now has conflicts. Maybe the issue got fixed via another PR? Could you confirm if this PR is still relevant?

Thanks!

NachoAlesLopez:
@LoisRForgeFlow It looks like the changes from https://github.com/OCA/server-ux/pull/1107 fixed the issue. I'll close the PR in that case. Thanks for letting me know of the fix!

LoisRForgeFlow:
@NachoAlesLopez That's good news! thanks for checking and confirming :+1:

PR #1056, opened 2025-03-29
[16.0][FIX] document_quick_access: We should search models with sudo

etobella:
Otherwise, an error can be raised on standdard users


luisDIXMIT (approved):
LGTM!

kluna1998 (approved):
LGTM!

PR #1057, opened 2025-03-31
[17.0][IMP] base_menu_visibility_restriction:Add excluded_menu_ids for res.groups

lyx-2018:
Add a menu field to the user group to facilitate the configuration of the menu according to the user group

### DIFF
--- a/base_menu_visibility_restriction/views/res_groups_views.xml
+++ b/base_menu_visibility_restriction/views/res_groups_views.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <!--
+        License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+    -->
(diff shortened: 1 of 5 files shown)


xtanuiha (approved):
Good job!👍

clearRM (approved):
good idea!

victoralmau (approved):

base_menu_visibility_restriction/views/res_groups_views.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <!--
+        License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
```
pedrobaeza:
```suggestion
        License AGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
```

pedrobaeza (changes requested):

PR #1059, opened 2025-04-04
[18.0][MIG] base_tier_validation_definition_server_action: Migration to 18.0

chaule97:


wlin-kencove:
image


While I can test the basic functionality of adding a tier validation  server action I cannot actually run the validation  since no models are currently enabled to to run with tier validation. 

image


I think we need additional dependencies in order to test comprehensively. (see example for sale orders workflow above)

PR #1022, opened 2025-02-05
[18.0][MIG] base_tier_validation_forward: Migration to 18.0

absal-smile:

PR #1065, opened 2025-04-16
[18.0][MIG] base_tier_validation_forward

Saran440:
This PR continue from https://github.com/OCA/server-ux/pull/1022

cc: @absal-smile


Saran440:
@chaule97 I updated code from your comment https://github.com/OCA/server-ux/pull/1022.
Please recheck.

bosd (approved):
Quick code review. :+1

base_tier_validation_forward/models/tier_validation.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def _notify_forwarded_reviews_body(self):
+        has_comment = self.review_ids.filtered(
+            lambda review: (self.env.user in review.reviewer_ids) and review.comment
+        )
+        if has_comment:
+            comment = has_comment.mapped("comment")[0]
+            return self.env._(
+                "A review was forwarded from %(user_name)s %(comment)s"
+            ) % ({"user_name": self.env.user.name, "comment": comment})
```
chaule97:
```suggestion
            return self.env._(
                "A review was forwarded from %(user_name)s %(comment)s",
                user_name=self.env.user.name,
                comment=comment,
            )
```

base_tier_validation_forward/models/tier_validation.py
```diff
@@ -0,0 +1,98 @@
...
+        return self.env._("A review was forwarded by %s.") % (self.env.user.name)
```
chaule97:
```suggestion
        return self.env._("A review was forwarded by %s.", self.env.user.name)
```

base_tier_validation_forward/wizard/forward_wizard.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def add_forward(self):
+        """Add extra step, with specific reviewer"""
+        self.ensure_one()
+        rec = self.env[self.res_model].browse(self.res_id)
+        prev_comment = self.env["comment.wizard"].browse(
+            self._context.get("comment_id")
+        )
+        prev_comment.write(
+            {"comment": self.env._(">> %s") % self.forward_reviewer_id.display_name}
```
chaule97:
```suggestion
            {"comment": self.env._(">> %s", self.forward_reviewer_id.display_name)}
```

chaule97 (approved):

base_tier_validation_forward/models/tier_review.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2020 Ecosoft Co., Ltd. (http://ecosoft.co.th)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+class TierReview(models.Model):
+    _inherit = "tier.review"
+    _order = "sequence"
+
+    name = fields.Char(compute="_compute_definition_data", store=True, related=False)
```
astirpe: We spotted a bug when we tried to install this 18.0 module to a customer.
To fix the bug, here we need a translate=True as well.

```suggestion
    name = fields.Char(compute="_compute_definition_data", store=True, related=False, translate=True)
```

Explanation:
In `base_tier_validation` field name is a related field and is `translate=True`. But here the related field is replaced with a computed field. This seems causing the loosing the translate=True setting for this field.
 
In `_compute_definition_data(...)` we have this line:
  `rec.name = rec.definition_id.name`
 
In case multiple languages are installed, you will get the following error:

```
psycopg2.errors.UndefinedFunction: function jsonb_path_query_first(character varying, unknown) does not exist
LINE 4: 'en_US', jsonb_path_query_fi...
^
```

The reason is that the jsonb string in rec.definition_id.name is assigned to rec.name which is a varchar.

Not sure if the explanation is clear, just let me know if you need more info.

astirpe (changes requested):

base_tier_validation_forward/models/tier_review.py
Saran440: @astirpe I can't reproduce to error. Can you help steps to reproduct please?

astirpe: @Saran440 
I tried to reproduce it locally but I didn't get the error this time. I was only getting the error on a customer environment, don't know why. Sorry.
However it's already fixed now by putting translate=True on the field definition.

Besides of the error, I still think that translate=True is needed here.
And also, in the compute method, instead of this line:
`    rec.name = rec.definition_id.name`
we need a deepcopy of the definition_id.name, so that all the translations will be copied.

Saran440: @astirpe I added translate=True on name field. please re-check.

base_tier_validation_forward/models/tier_review.py
```diff
@@ -0,0 +1,55 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+class TierReview(models.Model):
+    _inherit = "tier.review"
+    _order = "sequence"
+
+    name = fields.Char(
+        compute="_compute_definition_data", store=True, related=False, translate=True
```
simahawk, line 16: IMO we need a comment here mentioning why we keep `translate=True` :wink:

Saran440: Added `NOTE` in field name. Thank you for review.

astirpe (approved):

base_tier_validation_forward/models/tier_review.py
astirpe: Thank you!

rjaraspearhead (approved):

TDu (approved):

PR #1066, opened 2025-04-17
[FIX] tier validation overrides readonly on form of readonly fields (…

EmileNoppeSomko:
…computed fields for example)

Added a skip on adding the custom reaonly attribute on forms when the field is readonly (computed field for example), but the readonly attribute on the form is empty.
Without this Odoo wil not readonly=True on the fields because the custom readonly condition is set and readonly fields become writeable.

### DIFF
--- a/base_tier_validation/__manifest__.py
+++ b/base_tier_validation/__manifest__.py
@@ -3,7 +3,7 @@
 {
     "name": "Base Tier Validation",
     "summary": "Implement a validation process based on tiers.",
-    "version": "18.0.1.1.0",
+    "version": "18.0.1.1.1",
     "development_status": "Mature",
     "maintainers": ["LoisRForgeFlow"],
     "category": "Tools",
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -857,8 +857,13 @@ class TierValidation(models.AbstractModel):
                 node.append(new_node)
                 _merge_view_fields(all_models, new_models)
             excepted_fields = self._get_all_validation_exceptions()
+            readonly_fields = self.env['ir.model.fields'].search([('model', '=', self._name), ('readonly', '=', True)]).mapped('name')
             for node in doc.xpath("//field[@name][not(ancestor::field)]"):
                 if node.attrib.get("name") in excepted_fields:
+                    continue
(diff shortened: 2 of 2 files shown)


bosd:
@EmileNoppeSomko Can you please fix pre-commit?

base_tier_validation/models/tier_validation.py
```diff
@@ -857,9 +857,14 @@ def get_view(self, view_id=None, view_type="form", **options):
                 node.append(new_node)
                 _merge_view_fields(all_models, new_models)
             excepted_fields = self._get_all_validation_exceptions()
+            readonly_fields = self.env['ir.model.fields'].search([('model', '=', self._name), ('readonly', '=', True)]).mapped('name')
             for node in doc.xpath("//field[@name][not(ancestor::field)]"):
                 if node.attrib.get("name") in excepted_fields:
                     continue
+                if node.attrib.get("name") in readonly_fields and not node.attrib.get("readonly"):
+                    """Readonly field with readonly not set on form -> odoo will set this as readonly=True
+                    Make sure not to override"""
```
TDu, line 866:
```suggestion
                    # Readonly field with the attribute readonly not set on form -> will become editable
```
This should be a comment not a string

base_tier_validation/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Base Tier Validation",
     "summary": "Implement a validation process based on tiers.",
-    "version": "18.0.1.1.0",
+    "version": "18.0.1.1.1",
```
TDu, line 6: This change is not needed in the PR.
The version number will be automatically changed depending on the merge command.

TDu (commented):
Thanks for the fix. I hit the same issue today.
Here is how to install, run and fix the pre-commit error https://github.com/OCA/maintainer-tools/wiki/Install-pre-commit

TDu:
@EmileNoppeSomko  Would be great if you could do those changes...

TDu:
I opened #1077  to fix this PR
@EmileNoppeSomko cherry pick the fixup and integrate it here, if you want

StefanRijnhart:
Closing after https://github.com/OCA/server-ux/pull/1077 is merged. Thanks everyone for working on this important fix!

PR #1069, opened 2025-04-30
[16.0][IMP] base_tier_validation: Add message_main_attachment_id field to BASE_EXCEPTION_FIELDS

victoralmau:
Add `message_main_attachment_id` field to BASE_EXCEPTION_FIELDS

Allow adding an attachment when validation has been requested

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT56198

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -12,7 +12,12 @@ from odoo.exceptions import ValidationError
 from odoo.osv.expression import OR
 from odoo.tools.misc import frozendict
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
The reason is for allowing to add attachments, as this field is computed and written at least on the first one.

victoralmau:
Ping @CarlosRoca13

CarlosRoca13 (approved):

victoralmau:
I confirm that this change is not necessary in 17.0 or 18.0 because of the way attachments and messages work.

PR #1070, opened 2025-05-02
[16.0][IMP] base_tier_validation: Filter the definitions by all selected companies and by the company of record.

victoralmau:
Filter the definitions by all selected companies and by the company of record.

Example use case:
- There are several definitions (each linked to a different company).
- If a user has several companies selected at the same time, when requesting a validation, reviews should only be generated for the definitions of the corresponding company of record (purchase.order for example).

@Tecnativa TT56204


LoisRForgeFlow:
@victoralmau have you seen this one https://github.com/OCA/server-ux/pull/1052 ? It is actually a backpor from v18

victoralmau:
Oops, sorry, you are right, it solves the same problem, thanks.

PR #1076, opened 2025-05-08
[18.0][IMP] base_tier_validation: Allow attachment generation after validation

bosd:
On a lot of models, it is a regular flow to have a pdf file generated after the validation flow has finished.
e.g. on sale orders or invoices.

As discovered / discussed in: https://github.com/OCA/account-invoicing/issues/1998#issuecomment-2862437855

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -12,7 +12,11 @@ from odoo.exceptions import ValidationError
 from odoo.tools import SQL
 from odoo.tools.misc import frozendict
 
-BASE_EXCEPTION_FIELDS = ["message_follower_ids", "access_token"]
+BASE_EXCEPTION_FIELDS = [
+    "message_follower_ids",
+    "access_token",
+    "message_main_attachment_id",
+]
 
 
 class TierValidation(models.AbstractModel):


LoisRForgeFlow:
@victoralmau Maybe you have some good inputs for this one. In this PR you said that it was not needed version 17 and 18 (https://github.com/OCA/server-ux/pull/1069#issuecomment-2857505161).

victoralmau:
I have rechecked the behavior of this (e.g. through the `[purchase_request_tier_validation](https://github.com/OCA/purchase-workflow/tree/18.0/purchase_request_tier_validation)` module) and user (Marc Demo) if you can:
- Add attachments manually.
- Create notes with attachments.
- Send an email with attachments.

ejemplo-2

Can someone else check it to confirm that this change is not necessary (also not in v17)?

bosd:
I encountered the error on sale orders and invoices on V18.
Wierd, that purchase orders seems to be working.

bosd:
One can check the full flow in this pr: https://github.com/OCA/account-invoicing/pull/2002
First added the test to show it is failling without the fixes.
2nd commit adding the Fix for the account_move model
3 commit.. adding the modified base module with `message_main_attachment_id` in exclusion list. Tests are passing now :confetti_ball: 

>  In this PR you said that it was not needed version 17 and 18 

Additional Context:
I can confirm that in earlier versions of odoo 18, the account_move_tier_validation was working and the extension of these exclusions was not needed. Somewhere among the last couple of months the module stopped working. (e.g. unable to send invoice to customer after validation).

Maybe there is a code change in odoo core?

victoralmau (approved):
Ok, thanks for the clarification, that explains why I thought it was not necessary. 
We will then add these changes to correct this problem.

Can you also add it to 17?

bosd:
> Can you also add it to 17?

Sure, althoug did not check if it is now required on v17 as well.

LoisRForgeFlow (approved):
Thanks for checking guys!

/ocabot merge patch

bosd:
@LoisRForgeFlow Thanks for the merge!

Just tested the account_move_tier_validation on the V17 branch. There this fix is not needed. 

...(yet)

Issue #1067, opened 2025-04-17
Base tier validation wrongfullt overwrites readonly attribute on readonly fields (when readonly is empty on form)


EmileNoppeSomko:
## Module
base_tier_validation

## Describe the bug
Forms with tier validation will have readonly fields like computed fields become writeable in the pre review phase.
In some cases this can cause indirect errors. I believe this has to do with force_save which is supposed to be off on specific readonly fields, which triggers anyway when reaonly is False.

## To Reproduce
**Affected versions**: 18.0.1.1.0 (Probably all versions, I did not check any others)

1. Have a computed field with readonly=True (any readonly field will probably do)
2. Add it to a form with an empty readonly condition

**Expected behavior**
Just like without a validation module, the readonly attribute should be True instead of bool(review_ids)

**Additional context**
I wrote a fix here https://github.com/OCA/server-ux/pull/1066/files
Have a nice day!

PR #1077, opened 2025-05-13
[18.0][FIX] base_tier_valiation: readonly attribute

TDu:
This is to fix the PR #1066 
And in relation to issue https://github.com/OCA/server-ux/issues/1067

And related to this change in 16.0 https://github.com/OCA/server-ux/pull/860

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -861,13 +861,20 @@ class TierValidation(models.AbstractModel):
                 node.append(new_node)
                 _merge_view_fields(all_models, new_models)
             excepted_fields = self._get_all_validation_exceptions()
+            all_fields = self.fields_get(attributes=("readonly",))
             for node in doc.xpath("//field[@name][not(ancestor::field)]"):
-                if node.attrib.get("name") in excepted_fields:
+                field_name = node.attrib.get("name")
+                if field_name in excepted_fields:
                     continue
                 new_r_modifier = self._get_tier_validation_readonly_domain()
                 old_r_modifier = node.attrib.get("readonly")
                 if old_r_modifier:
                     new_r_modifier = f"({old_r_modifier}) or ({new_r_modifier})"
+                elif all_fields.get(field_name, {}).get("readonly"):
+                    # don't set dynamic readonly attribute for fields
+                    # marked readonly in the ORM (ie computed fields)
+                    # if the view doesn't set one
+                    continue
                 node.attrib["readonly"] = new_r_modifier
             res["arch"] = etree.tostring(doc)
             res["models"] = frozendict(all_models)


base_tier_validation/models/tier_validation.py
```diff
@@ -861,9 +861,20 @@ def get_view(self, view_id=None, view_type="form", **options):
...
+                .mapped("name")
+            )
             for node in doc.xpath("//field[@name][not(ancestor::field)]"):
                 if node.attrib.get("name") in excepted_fields:
                     continue
+                if node.attrib.get("name") in readonly_fields and not node.attrib.get(
+                    "readonly"
+                ):
+                    # Readonly field with readonly not set become editable
+                    # Happens for computed fields
```
sebalix: Is this comment more accurate?
```suggestion
                    # Do not update 'readonly' view attribute for fields defined as readonly server-side
```

base_tier_validation/models/tier_validation.py
```diff
@@ -861,9 +861,20 @@ def get_view(self, view_id=None, view_type="form", **options):
                 node.append(new_node)
                 _merge_view_fields(all_models, new_models)
             excepted_fields = self._get_all_validation_exceptions()
+            readonly_fields = (
+                self.env["ir.model.fields"]
+                .search([("model", "=", self._name), ("readonly", "=", True)])
+                .mapped("name")
+            )
```
hbrunn:
```suggestion
```

base_tier_validation/models/tier_validation.py
```diff
@@ -861,9 +861,20 @@ def get_view(self, view_id=None, view_type="form", **options):
...
             for node in doc.xpath("//field[@name][not(ancestor::field)]"):
                 if node.attrib.get("name") in excepted_fields:
                     continue
+                if node.attrib.get("name") in readonly_fields and not node.attrib.get(
```
hbrunn:
```suggestion
                if all_models[self._name]['fields'][node.attrib.get("name")]['readonly'] and not node.attrib.get(
```

hbrunn (changes requested):

lk-eska:
This is a serious problem. It also exists in 17.0. Since the author has not time to look, can we speed up the merge of this fix?

LoisRForgeFlow:
Hi @lk-eska

I do my best reviewing the incoming PRs for tier validation, but they are quite a few. In this particular case, no one requested my review explicitly yet other than the bot. On top of that there is a review from @hbrunn pending to be attended and not a single approval, so... for now it is not possible to merge.

lk-eska:
@LoisRForgeFlow thanks for the quick response. We had to patch it without waiting for the merge. It is tested and works properly.

As it is now, it has strange behavior on readonly fields...

base_tier_validation/models/tier_validation.py
TDu: @hbrunn I tried the improvement you propose. But `all_models[self._name]` is just a tuple containing the fields names.

hbrunn: ah right, we're in get_view, not get_views. still it's wrong to query the database for this, I've pushed a proposal for how I think this should look in https://github.com/camptocamp/server-ux/pull/3

For the record: This is a migration issue, as at least v16 for which I conceived this *always* puts the readonly attribute into the arch. A test would be helpful

hbrunn:
@LoisRForgeFlow I don't feel you need to justify yourself for not replying within a week, specifically not to people who didn't even take the low effort to leave a review

base_tier_validation/models/tier_validation.py
TDu: Thank you, included and tested.

mmequignon (approved):

hbrunn (approved):
feel free to squash my commit into yours

LoisRForgeFlow (approved):
LGTM, thanks for the patch :+1:

lk-eska:
thank you all! this needs to be rebased to older versions that used the same logic (17 and 16 afaik) as well.

LoisRForgeFlow:
> thank you all! this needs to be rebased to older versions that used the same logic (17 and 16 afaik) as well.

@TDu could you take care of the backports?

TDu:
Backport for 17 #1107

TDu:
Backport to 16 is a little more tricky, there is a deviation in the code, may look at it later

LoisRForgeFlow:
OK, thank you!

PR #1083, opened 2025-05-20
[16.0][FIX] base_tier_validation: Proper notifications on reviews

victoralmau:
Proper notifications on reviews

Changes done:
- The subtypes should not be defined as default=True to avoid that a follower receives notifications of everything (even if he/she doesn't really need it).
- When adding followers they are added to the corresponding subtype so that the notification arrives to whom it corresponds.
- Improve the code so that the reviews to notify are the "following ones according to the sequence".

@Tecnativa TT56354

### DIFF
--- a/base_tier_validation/data/mail_data.xml
+++ b/base_tier_validation/data/mail_data.xml
@@ -6,7 +6,7 @@
         forcecreate="1"
     >
         <field name="name">Tier Validation Requested</field>
-        <field name="default" eval="True" />
+        <field name="default" eval="False" />
         <field name="internal" eval="True" />
         <field name="hidden" eval="True" />
     </record>
@@ -16,7 +16,7 @@
         forcecreate="1"
     >
         <field name="name">Tier Validation Accepted Notification</field>
-        <field name="default" eval="True" />
+        <field name="default" eval="False" />
         <field name="internal" eval="True" />
         <field name="hidden" eval="True" />
     </record>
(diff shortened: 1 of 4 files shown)


LoisRForgeFlow (changes requested):
Just signaling that I would like to review this when ready

victoralmau:
You can review it and indicate what you think about it, although I still leave the WIP tag in the PR title until it is clear that this is the right approach.

Ping @pedrobaeza and @sergio-teruel what do you think about it?

pedrobaeza:
I suppose this is to reduce the noise when you chain approvers so only the approver in the proper moment is notified, isn't it? It's a change of behavior, but it sounds reasonable.

LoisRForgeFlow:
Hi @victoralmau 

I have checked your proposal, in general it makes sense and it is cleaner. However I'm not sure if pushing this change in v16 can disturb some installations. What do you think? Maybe it is best to do it only in 18?

pedrobaeza:
We can remove the migration scripts for not affecting existing installations and minimizing the impact.

LoisRForgeFlow:
> We can remove the migration scripts for not affecting existing installations and minimizing the impact.

OK, let's do that, there is still impact with the rest but at the end of the day this behavior makes more sense.

LoisRForgeFlow (approved):
:+1:

pedrobaeza (approved):
/ocabot merge major

PR #1084, opened 2025-05-22
[16.0][IMP] base_tier_validation: Make the validation_status field store so that it can be used in searchs.

victoralmau:

@Tecnativa TT56391


carlos-lopez-tecnativa (approved):
Tested locally, LGTM. I only think that recomputing these fields to be stored might take a lot of time on databases with a lot of data.

pedrobaeza:
Yes, including this at this stage makes the update very slow. The `search` approach avoids this, but not sure if it's doable.

victoralmau:
Another possibility is to add migration scripts so that this change is not slow. Adding the store fields I think is useful, although you will need the migration scripts when they are done. I work on them a bit to avoid this inconvenience.

base_tier_validation/models/tier_validation.py
```diff
@@ -236,18 +248,6 @@ def _compute_is_reevaluation_required(self):
                 if len(valid_tiers) != len(rec.review_ids.definition_id):
                     rec.is_reevaluation_required = True
 
-    @api.model
-    def _calc_reviews_validated(self, reviews):
```
StefanRijnhart: Would it not be nice to keep this mechanism? It allows to take values for other fields into account when determining the validation status. See how https://github.com/OCA/server-ux/pull/1101 allows to define those fields as dependencies for the computed status in its `_get_validation_status_dependencies`.

base_tier_validation/models/tier_validation.py
```diff
@@ -42,19 +42,14 @@ class TierValidation(models.AbstractModel):
         domain=lambda self: [("model", "=", self._name)],
         auto_join=True,
     )
-    to_validate_message = fields.Html(compute="_compute_validated_rejected")
-    # TODO: Delete in v17 in favor of validation_status field
```
StefanRijnhart, line 46: The fields are still superfluous (because you can now search for `validation_status == 'validated'`, and now they even take up space in the table. In https://github.com/OCA/server-ux/pull/1101 the fields are still queued for removal, but in the mean time their search method was much improved by using the stored validation_status field.

StefanRijnhart (commented):
A much needed improvement! Somewhat simultaneously, https://github.com/OCA/server-ux/pull/1101 was developed for 18.0. We should probably align on the details.

base_tier_validation/models/tier_validation.py
victoralmau: The fields used in https://github.com/OCA/server-ux/pull/1084 in the `_get_validation_status_dependencies()` method are already being used here as depends of the `_compute_validation_status()` method (and some more fields). IMO keeping this method (`_calc_reviews_validated()`) would be a mistake, it is no longer needed, the proper thing is to use the status (rather, the `validated` field), don't you think?

base_tier_validation/models/tier_validation.py
victoralmau: Sorry, do you mean the `to_validate_message` field (not store) or the `validated` and `rejected` fields? 
IMO defining the `validated` and `rejected` fields as store is not a “_problem_” because table space is insignificant, and besides, they are very easy to define their value without the need to maintain _search code for it, don't you think (the idea is to improve and simplify as much as possible, and this is one of those “_easy_” things to do)?

base_tier_validation/models/tier_validation.py
StefanRijnhart: OK, so the approach in #1101 is to make the validation_status leading, and to allow a developer to steer the judgment by overriding `_calc_reviews_validated`, adding additional dependencies in `_get_validation_status_dependencies`, while the approach here is to make `validated` and `rejected` leading, allowing a developer to steer the judgment by overriding `_compute_validated` and `_compute_rejected`, adding additional dependencies on these methods. I guess your approach is simpler indeed!

base_tier_validation/models/tier_validation.py
StefanRijnhart: Given that customizations of the flow now depend on `_compute_validated` and `_compute_rejected` as per the other comment, I agree!

StefanRijnhart (approved):
Unless you have this lying around already, I'll be porting this to 18.0 in the coming days.

pedrobaeza:
@LoisRForgeFlow do you agree with this change?

base_tier_validation/migrations/16.0.3.1.0/pre-migration.py
```diff
@@ -0,0 +1,110 @@
...
+                )
+                """,
+            )
+            openupgrade.logged_query(
+                env.cr,
+                f"""
+                UPDATE {table_name} SET validated = false WHERE id IN (
+                    SELECT DISTINCT(tr.res_id)
+                    FROM tier_review AS tr
+                    WHERE tr.model = '{model_name}' AND tr.status != 'approved'
```
StefanRijnhart:
```suggestion
                    WHERE tr.model = '{model_name}' AND tr.status not in ('approved', 'forwarded')
```
if I understand correctly.

base_tier_validation/migrations/16.0.3.1.0/pre-migration.py
```diff
@@ -0,0 +1,110 @@
...
+                env.cr,
+                f"""
+                UPDATE {table_name} SET validation_status = 'no'
+                """,
+            )
+            openupgrade.logged_query(
+                env.cr,
+                f"""
+                UPDATE {table_name} SET validation_status = 'validated'
+                WHERE validated = true AND rejected = false
```
StefanRijnhart:
```suggestion
                WHERE validated = true AND coalesce(rejected, false) = false
```
The column is still null at this point for records that have not been rejected.

base_tier_validation/migrations/16.0.3.1.0/pre-migration.py
```diff
@@ -0,0 +1,110 @@
...
+                f"""
+                UPDATE {table_name} SET validation_status = 'rejected'
+                WHERE validated = false AND rejected = true;
+                """,
+            )
+            openupgrade.logged_query(
+                env.cr,
+                f"""
+                UPDATE {table_name} SET validation_status = 'pending'
+                WHERE validated = false AND rejected = false AND id IN (
```
StefanRijnhart:
```suggestion
                WHERE validated = false AND coalesce(rejected, false) = false AND id IN (
```

StefanRijnhart:
Port to 18 at https://github.com/OCA/server-ux/pull/1129

StefanRijnhart (changes requested):
As per earlier comments

base_tier_validation/models/tier_validation.py
```diff
@@ -42,19 +42,14 @@ class TierValidation(models.AbstractModel):
...
+    to_validate_message = fields.Html(compute="_compute_to_validate_message")
     validated = fields.Boolean(
-        compute="_compute_validated_rejected", search="_search_validated"
+        compute="_compute_validated", store=True, compute_sudo=True
     )
-    validated_message = fields.Html(compute="_compute_validated_rejected")
+    validated_message = fields.Html(compute="_compute_validated_message")
     need_validation = fields.Boolean(compute="_compute_need_validation")
-    # TODO: Delete in v17 in favor of validation_status field
-    rejected = fields.Boolean(
```
LoisRForgeFlow, line 53: This TODO has been removed, do you think is not feasible to adapt everything to work with the validation_status and completely removing these booleans?

Not saying that it needs to be done right now in this PR, it is more like "is it a bad roadmap for any reason?"

LoisRForgeFlow (commented):
I haven't done a deep review but as a concept, I'm fine with the change

base_tier_validation/models/tier_validation.py
victoralmau: IMO the TODO did not refer to remove the field, it referred to use the validation_status, now the field is compute+store and has the appropriate depends, so I think it is not necessary to do anything else.

In addition, I think it is appropriate to have these fields, they are used in part of the code and are still necessary, and also, it is useful that they exist stored to be able to filter by that data if for example, you have a list of reviews.

StefanRijnhart: @LoisRForgeFlow As I pondered in another (resolved) comment, if anything, `_compute_rejected` and `_compute_validated` serve as entry points for customizations of the flow, replacing `_calc_reviews_validated` and `_calc_reviews_rejected` that are removed in this change. So I agree with @victoralmau.

pedrobaeza: But then `validation_status` may have `rejected` but `rejected` to be False? I see that as an inconsistency. If not removing this in this version for compatibility reasons, it has to be removed ASAP.

LoisRForgeFlow: Exactly, if you look how the validation_status is computed, it could be computed based on the reviews status and forget about a rejected, approved, need review booleans. It is redundant. Of course the refactor is not easy, because it will affect views and a lot of places in the code, but we can have same functionality and extension points with only `validation_status`.

victoralmau: The rejected/validated fields should reflect reality, i.e. show what corresponds according to the validation_status field. 

The interface should not allow (as is currently the case) to validate something if it is already rejected or reject something if it is already validated for example, so I don't really see the inconsistency (unless someone modifies the record and defines it with inconsistent data, as could happen in any part of odoo, sales for example and billing status).

StefanRijnhart: > forget about a rejected, approved, need review booleans

If you want to go that way, that was the approach of https://github.com/OCA/server-ux/pull/1101, and it was not overly complicated.

StefanRijnhart: But again, there was always the encouragement to override the flow in the `_calc_` methods: https://github.com/OCA/server-ux/blob/18.0/base_tier_validation/models/tier_validation.py#L235-L245. It is not unreasonable to keep a version of it, I think, and if so, this PR's approach seems cleaner than before.

pedrobaeza: But as said, that can lead to inconsistencies between the values...

victoralmau: But .... same as any compute store field, right? Maybe I'm missing something, but I still don't see the problem with keeping these fields and making the approach of this PR more “_simple_” than removing them and changing everything necessary in the views. Also, what happens if another module adds a validation state x and in that new state it is rejected? With compute/write fields it could be done, otherwise, it would be more complicated/incorrect (e.g. https://github.com/OCA/server-ux/pull/1101/files#diff-af78280ddd15c60f3e3b3cfc9ffb3808a0bffac45e723028cadb63e3c689e41cR32).

StefanRijnhart: @pedrobaeza 
> But then validation_status may have rejected but rejected to be False? I see that as an inconsistency. If not removing this in this version for compatibility reasons, it has to be removed ASAP.

This is not actually the case. if one were to override `_compute_rejected`, the resulting value may be `False` while an individual tier review was in fact rejected. Because `validation_status` is based on the values of `approved` and `rejected`, this will prevent `validation_status` from being updated to `rejected`. A use case may be to allow one rejection to be offset by multliple approvals.

StefanRijnhart (approved):
Thanks for fixing the migration script!

kevinkhao (approved):

kevinkhao:
@LoisRForgeFlow ping
I can see merits to both approaches. I'm OK with direction of this PR.

LoisRForgeFlow:
I'm OK with this PR, I still agree with Pedro that the booleans (approved, rejected, etc) are redundant but it is not a task for this PR. SO, this is fine for me.

@pedrobaeza I let you the final call here

base_tier_validation/models/tier_validation.py
LoisRForgeFlow: That is a problem on how you inherit it, but at the end of the day it is a data model that represent the same concept in 2 ways. validation_satatus "approved" means the same as `approved` True, etc. From a design stand point, I still think it is not clean. But as said, this is not the goal of this PR, I would just ask to keep the TODO comment because I think it is still relevant and cambe be attended in the future.

pedrobaeza: I have checked again about this, and I think we have to remove the `rejected` and `validated` fields, and more now that you want to make them stored. You can keep `calc_reviews_validated` and `_calc_reviews_rejected` method the same for inheritance (and compatibility), but this time checking `validation_status` instead.

victoralmau: Okay, those changes have been applied:

- The `rejected` and `validated` fields have been removed
- The `calc_reviews_validated` and `_calc_reviews_rejected` methods remain unchanged

pedrobaeza (approved):
/ocabot merge major

LoisRForgeFlow:
@victoralmau Thanks for the work done here :+1:

base_tier_validation/models/tier_validation.py
```diff
@@ -177,33 +153,49 @@ def _get_validated_message(self):
...
+    @api.depends("review_ids", "review_ids.status")
     def _compute_validation_status(self):
         for item in self:
-            if item.validated and not item.rejected:
+            reviews = item.review_ids
+            any_approved = any(reviews.filtered(lambda x: x.status == "approved"))
+            any_rejected = any(reviews.filtered(lambda x: x.status == "rejected"))
+            any_pending = any(reviews.filtered(lambda x: x.status == "pending"))
+            if reviews and any_approved and not any_rejected:
                 item.validation_status = "validated"
```
LoisRForgeFlow, line 195: @victoralmau Careful here, this is not equivalent. For something to be approved, all reviews need to be approved.

See previous condication for "approved": ` return not any([s != "approved" for s in reviews.mapped("status")])`

victoralmau: https://github.com/OCA/server-ux/pull/1135

StefanRijnhart:
@pedrobaeza I would have appreciated a chance to review the heavily modified version that was pushed on Thursday.

pedrobaeza:
AFAIK, only the removal of the 2 fields was added...

StefanRijnhart:
@pedrobaeza No, the computation of validation_status was refactored, and it is not the same as before. Comments on the 17.0 PR.

StefanRijnhart:
Also, not keeping a backwards compatible version of the fields is not something I would expect in a mature version like 16.0.

PR #1088, opened 2025-05-29
[16.0][FIX] base_search_custom_field_filter: Avoid hiding model fields in other parts of the UI

CarlosRoca13:
Before these changes, if a model field was added to the custom filters, it would become disabled in other parts of the interface — for example, when trying to group by that field.

After these changes, the filter for the simple field is added while preserving the ability to group by it. Only fields that don't belong to the data model itself — that is, those using '.' notation — are hidden.

cc @Tecnativa TT56518

ping @pedrobaeza @pilarvargas-tecnativa


pilarvargas-tecnativa (approved):

carlosdauden (approved):

sergio-teruel (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1081, opened 2025-05-20
[18.0][IMP] base_tier_validation: support res.groups in addition to res.users in field-based validation

kevinkhao:

PR #1089, opened 2025-06-03
[18.0][FIX] base_tier_validation: fix reviewer field check

mmequignon:
Raises a misleading exception when reviewer field is correctly set, but is empty on the record.
Put this as draft / WIP as there's another exception raised once this is fixed that I need to tackle next.

### DIFF
--- a/base_tier_validation/models/tier_review.py
+++ b/base_tier_validation/models/tier_review.py
@@ -153,10 +153,42 @@ class TierReview(models.Model):
         reviewer_field = self.env["res.users"]
         if self.reviewer_field_id:
             resource = self.env[self.model].browse(self.res_id)
-            reviewer_field = getattr(resource, self.reviewer_field_id.name, False)
-            if not reviewer_field or not reviewer_field._name == "res.users":
+            field_name = self.reviewer_field_id.name
+            no_field_on_model = field_name not in resource._fields
+            if no_field_on_model:
+                msg = self.env._(
+                    "There are no field %(field_name)s on the %(model_name)s model",
+                )
                 raise ValidationError(
-                    self.env._("There are no res.users in the selected field")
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow (commented):
@mmequignon are going to finish this? If so, remember to add test cases.

mmequignon:
> @mmequignon are going to finish this? If so, remember to add test cases.

done

base_tier_validation/models/tier_review.py
```diff
@@ -153,10 +153,42 @@ def _get_reviewers(self):
...
-            if not reviewer_field or not reviewer_field._name == "res.users":
+            field_name = self.reviewer_field_id.name
+            no_field_on_model = field_name not in resource._fields
+            if no_field_on_model:
+                msg = self.env._(
+                    "There are no field %(field_name)s on the %(model_name)s model",
+                )
+                raise ValidationError(
+                    msg % {"field_name": field_name, "model_name": resource._name}
+                )
```
ivantodorovich, line 164:
```suggestion
                raise ValidationError(
                    self.env._(
                        "There are no field %(field_name)s on the %(model_name)s model",
                        field_name=field_name,
                        model_name=resource._name,
                    )
                )
```

use Odoo's gettext built-in string interpolation
This lets Odoo recover gracefully from incorrect translation terms



(same applies to other cases below)

ivantodorovich (changes requested):

simahawk:
@mmequignon The conflict comes from this fix https://github.com/OCA/server-ux/pull/1081 that probably addresses the same issue.

mmequignon:
> @mmequignon The conflict comes from this fix #1081 that probably addresses the same issue.

You're right, I'm closing this PR.

Issue #1048, opened 2025-03-25
Blocking error on upgrade: Expected singleton in date_range_generator.py due to store=True on date_end field (14.0)


mktsrl:
## Module
date_range

## Describe the bug
After updating the server-ux repository to the latest 14.0 version, the upgrade -u all command fails due to a blocking error in the date_range module.

The error originates from the date_end computed field in date_range_generator.py. In the latest commit, the field date_end has been changed to store=True. This causes a ValueError: Expected singleton when the _compute_date_end method is triggered during the upgrade.

This problem did not occur in previous versions, where the field was not stored. Also, checking the 15.0 and 16.0 versions of the module, the field is not stored there either, which confirms this change is inconsistent and possibly unintentional.

Removing the store=True flag solves the issue: the database upgrade works without errors.

## To Reproduce
**Affected versions**: 14.0

1. Update the server-ux repository to the latest version of branch 14.0.
2. Run -u all on a database containing the date_range module.
3. Observe the following traceback:

  File ".../oca-addons/server-ux/date_range/wizard/date_range_generator.py", line 280, in _compute_date_end
    if not self.type_id or not self.date_start:
  File ".../odoo/fields.py", line 987, in __get__
    record.ensure_one()
  File ".../models.py", line 5050, in ensure_one
    raise ValueError("Expected singleton: %s" % self)
ValueError: Expected singleton: date.range.generator(1, 2)

**Expected behavior**
The database upgrade should complete without errors. The computed field date_end should work without causing singleton issues, as in previous versions where it was not stored.

**Additional context**
Remove store=True from the date_end field definition in date_range_generator.py.

PR #1091, opened 2025-06-03
[14.0][FIX] resolve Error:Expected singleton: date.range.generator (issue 1048)

mktsrl:
Resolve issue [1048](https://github.com/OCA/server-ux/issues/1048)


TennyMkt (approved):

lmignon (changes requested):
IMO it's not the right fix. It could be a workaround but the problem is that the compute methods are not written to be called on a recordset with more than one record in it. If all the compute methods were properly written by making a loop on self, this problem would never occur.

danall59 (approved):
It's OK

PR #1093, opened 2025-06-20
[17.0][FIX] base_tier_validation: make ReviewsTable widget work with manual config

MaxyMoos:
When an Odoo model inherits from `tier.validation` but sets up `_tier_validation_manual_config = True` in the model class, adding `review_ids` in the model form view with `widget="tier_validation"` causes Owl errors on rendering when at least one review is linked to the model:

This fix adds the fields used in the `base_tier_validation.Collapse` widget template to the relevant component under the `relatedFields` key (which is what Odoo does with the Many2ManyTags widget, and where i got the idea to try this), which solves the issue while not changing the existing behavior for models using `_tier_validation_manual_config = False`.

I've made a small Odoo module that allows to quickly reproduce the issue since it asks for a bit of setup. It's available [here](https://github.com/MaxyMoos/base_tier_validation_test_module)

### DIFF
--- a/base_tier_validation/static/src/components/tier_review_widget/tier_review_widget.esm.js
+++ b/base_tier_validation/static/src/components/tier_review_widget/tier_review_widget.esm.js
@@ -35,6 +35,19 @@ ReviewsTable.template = "base_tier_validation.Collapse";
 
 export const reviewsTableComponent = {
     component: ReviewsTable,
+    supportedTypes: ["one2many"],
+    relatedFields: [
+        {name: "id", type: "integer"},
+        {name: "sequence", type: "integer"},
+        {name: "name", type: "char"},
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (approved):
Thanks for the testing module and the description, I tested and was able to reproduce the issue :+1:

LoisRForgeFlow:
@celm1990 as you did that PR to migrate to 17, maybe you can have a look here?

celm1990 (approved):
LGTM

LoisRForgeFlow:
@MaxyMoos needed in v18? could you port it?

MaxyMoos:
@LoisRForgeFlow Yes, I just tested and the issue is in v18 too, I opened up #1117 to bring the fix over

PR #1040, opened 2025-03-10
[16.0][IMP] base_tier_validation: Hide Request Validation button if validation status is Pending

victoralmau:
Hide Request Validation button if validation status is Pending

**Before**
antes

**After**
despues

@Tecnativa TT55409

PR #1094, opened 2025-06-24
[16.0][FIX] base_tier_validation: add field attrs references

hbrunn:
since #1040, the request validation button is visible for records that have already validation request, because the new condition introduced there isn't added to the form by default (forms that have the field through other means won't suffer from this):

image

You can check this on [runbot](http://oca-account-invoicing-16-0-d4443b512f89.runboat.odoo-community.org/web?debug=1#id=6&cids=1&menu_id=115&action=233&model=account.move&view_type=form) as long as the instance is alive.

### DIFF
--- a/base_tier_validation/templates/tier_validation_templates.xml
+++ b/base_tier_validation/templates/tier_validation_templates.xml
@@ -2,6 +2,7 @@
 <odoo>
     <template id="tier_validation_buttons">
         <div>
+            <field name="validation_status" invisible="1" />
             <button
                 name="request_validation"
                 string="Request Validation"


victoralmau (approved):

LoisRForgeFlow (approved):
:+1: Thanks for the runboat link to speed up testing

LoisRForgeFlow:
@hbrunn In v17+ invisible conditions does not require that the referenced fields are explicitly declared in the view. Therefore, I'm not sure if this fix is needed in next versions, could you confirm?

hbrunn:
@LoisRForgeFlow test it, not sure at the moment given this is added dynamically after normal view processing

PR #1097, opened 2025-06-26
[RFC][18.0][IMP] base_tier_validation: cleanup rejected/validated fields, remove extraneous functions

kevinkhao:
This PR is split into 3 commits:
1. Fix performance issues. _search_validated_rejected kills performance when there are many records and can cause server crashes when there is a high number of account.move records
- remove fields rejected and validated
- uses validation_status instead, as intended in the comments
- makes the following fields stored: validation_status, validated_message, rejected_message
- the business logic is kept intact

2. Cleanup the message builder functions
- remove fields validated_message, rejected_message 
- remove associated functions
They are used nowhere in the base module. This module is already heavy, messaging functionality should be in its own module.

3. fix test on base_tier_validation_forward (minor)

### DIFF
--- a/base_tier_validation/README.rst
+++ b/base_tier_validation/README.rst
@@ -66,16 +66,16 @@ To configure this module, you need to:
 
 **Note:**
 
-- If check *Notify Reviewers on Creation*, all possible reviewers will
-  be notified by email when this definition is triggered.
-- If check *Notify reviewers on reaching pending* if you want to send a
-  notification when pending status is reached. This is usefull in a
-  approve by sequence scenario to only notify reviewers when it is their
-  turn in the sequence.
-- If check *Comment*, reviewers can comment after click Validate or
-  Reject.
-- If check *Approve by sequence*, reviewers is forced to review by
-  specified sequence.
+-  If check *Notify Reviewers on Creation*, all possible reviewers will
+   be notified by email when this definition is triggered.
+-  If check *Notify reviewers on reaching pending* if you want to send a
+   notification when pending status is reached. This is usefull in a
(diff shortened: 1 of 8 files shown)


kevinkhao:
@len-foss could you review ?

len-foss (commented):
This should be a great improvement, as it simplifies things a lot. Unfortunately, removing the fields already so late in the 18.0 life-cycle is a bit iffy to say the least.

This PR addresses the fundamental design flaw to this module: the `_compute_status` function doesn't have any triggers. 
This is painfully obvious because of all the `.invalidate_model()` that are written everywhere.
I think if we change this, this should also be changed to work exactly in the way that `_compute_reviewer_ids` works, that is to say:
```
@api.depends(lambda self: self._get_reviewer_fields())
    def _compute_reviewer_ids(self):
```
So that each overriding field can extend the list properly.
Once this is done, every single `.invalidate_model()` should be removed for good.

As an aside on the module usability, I think it would be nice to introduce a little helper:
```
needs_review = fields.Boolean(compute="_compute_needs_review")

@api.depends("need_validation", "validation_status")
def _compute_needs_review(self):
    for record in self:
        record.needs_review = (
            record.need_validation
            or record.validation_status not in ["validated", "no"]
        )
```
In some cases, there's an operation you need to block depending on the validation status, but 'no' also means that it hasn't started yet, which is a bit unwieldly to use.  Naming might not be the best ofc.

I would give it a big thumbs up once all the `.invalidate_model()` are removed, since this is the moment to do it.

kevinkhao:
closing in favour of https://github.com/OCA/server-ux/pull/1101

PR #1098, opened 2025-06-26
[18.0][MIG] mail_suggested_recipient_unchecked: Migration to 18.0

victoralmau:
Migration to 18.0

Please @CarlosRoca13 and @pedrobaeza can you review it?

@Tecnativa TT55538


mail_suggested_recipient_unchecked/models/mail_thread.py
```diff
@@ -0,0 +1,18 @@
...
+from odoo import models
+
+
+class MailThread(models.AbstractModel):
+    _inherit = "mail.thread"
+
+    def _message_add_suggested_recipient(
+        self, result, partner=None, email=None, lang=None, reason=""
+    ):
+        res = super()._message_add_suggested_recipient(
```
pedrobaeza, line 14: Remember to add comments saying why you override the method:

```python
    # Uncheck all the suggested recipients updating the dictionary value
```

pedrobaeza (approved):
¡Better to do this with Python now that it's possible!

Code and runboat review.

Just a small comment.

mail_suggested_recipient_unchecked/models/mail_thread.py
```diff
@@ -0,0 +1,19 @@
...
+from odoo import models
+
+
+class MailThread(models.AbstractModel):
+    _inherit = "mail.thread"
+
+    def _message_add_suggested_recipient(
+        self, result, partner=None, email=None, lang=None, reason=""
+    ):
+        """Uncheck all the suggested recipients updating the dictionary value."""
```
CarlosRoca13:
```suggestion
        # Uncheck all the suggested recipients updating the dictionary value.
```
Add this as a comment, otherwise you're overwriting the original method's docstring.

CarlosRoca13 (approved):
Just a minor comment

victoralmau:
It's ready to merge, right?

marcelsavegnago (approved):

victoralmau:
According to https://github.com/OCA/server-ux/pull/1121#pullrequestreview-3004787138, this module should have been moved to OCA/mail (the `mail_message_destiny_link_template` module should not be in OCA/server-ux either too).

Now that this module is already merged in 18.0, do we see it to OCA/mail or leave it for 19.0?

pedrobaeza:
We can remove it here and put it in OCA/mail. I think there's no problem with that.

PR #1099, opened 2025-06-27
[18.0][MIG] announcement: Migration to version 18.0

david-banon-tecnativa:
@tecnativa @CarlosRoca13 @victoralmau

Updating copier template was necessary to fix pre-commit errors.

TT55499

PR #1100, opened 2025-06-27
[18.0][MIG] announcement_dialog_size: Migration to version 18.0

david-banon-tecnativa:
Depends on:

- [ ] https://github.com/OCA/server-ux/pull/1099

@tecnativa @CarlosRoca13 @pilarvargas-tecnativa  

TT55500


CarlosRoca13:
I think that this glue module is not needed anymore. Installing announcement and web_dialog_size I can resize the announcement dialog. The only reason I would keep this module is if there were a need to set the dialog size independently from the others in Odoo. But from my point of view, I don't see it as necessary.

Maybe @pedrobaeza or @victoralmau can tell us if this is the actual purpose of the module.

pedrobaeza:
@CarlosRoca13 the need was if you have the parameter set for expanding by default. Is that being respected? Anyway, check in the customer instance if it's set, because maybe we don't need it anyway.

CarlosRoca13:
Yes, right now it is respected, and our customer instance have not set the parameter. So this module is definitely not necessary

PR #1101, opened 2025-06-27
[18.0][IMP] base_tier_validation: deprecate rejected/validated fields (for performance)

kevinkhao:
This PR fixes performance issues and clarifies the dependencies. 

- store validation_status to allow performant search on records in validation
- _search_validated_rejected kills performance when there are many records and can cause server crashes when there is a high number of account.move records -> replace by search on validation status
- stop using fields rejected and validated and keep them for compatibility, use validation_status instead, as intended in the comments
- update roadmap 
- future improvements in usability

### DIFF
--- a/base_tier_validation/README.rst
+++ b/base_tier_validation/README.rst
@@ -70,16 +70,16 @@ To configure this module, you need to:
 
 **Note:**
 
-- If check *Notify Reviewers on Creation*, all possible reviewers will
-  be notified by email when this definition is triggered.
-- If check *Notify reviewers on reaching pending* if you want to send a
-  notification when pending status is reached. This is usefull in a
-  approve by sequence scenario to only notify reviewers when it is their
-  turn in the sequence.
-- If check *Comment*, reviewers can comment after click Validate or
-  Reject.
-- If check *Approve by sequence*, reviewers is forced to review by
-  specified sequence.
+-  If check *Notify Reviewers on Creation*, all possible reviewers will
+   be notified by email when this definition is triggered.
+-  If check *Notify reviewers on reaching pending* if you want to send a
+   notification when pending status is reached. This is usefull in a
+   approve by sequence scenario to only notify reviewers when it is
+   their turn in the sequence.
+-  If check *Comment*, reviewers can comment after click Validate or
+   Reject.
+-  If check *Approve by sequence*, reviewers is forced to review by
+   specified sequence.
 
 To configure Tier Validation Exceptions, you need to:
 
@@ -94,14 +94,16 @@ To configure Tier Validation Exceptions, you need to:
 
 **Note:**
 
-- If you don't create any exception, the Validated record will be
-  readonly and cannot be modified.
-- If check *Write under Validation*, records will be able to be modified
-  only in the defined fields when the Validation process is ongoing.
-- If check *Write after Validation*, records will be able to be modified
-  only in the defined fields when the Validation process is finished.
-- If check *Write after Validation* and *Write under Validation*,
-  records will be able to be modified defined fields always.
+-  If you don't create any exception, the Validated record will be
+   readonly and cannot be modified.
+-  If check *Write under Validation*, records will be able to be
+   modified only in the defined fields when the Validation process is
+   ongoing.
+-  If check *Write after Validation*, records will be able to be
+   modified only in the defined fields when the Validation process is
+   finished.
+-  If check *Write after Validation* and *Write under Validation*,
+   records will be able to be modified defined fields always.
 
 Known issues / Roadmap
 ======================
@@ -109,25 +111,54 @@ Known issues / Roadmap
 This is the list of known issues for this module. Any proposal for
 improvement will be very valuable.
 
-- **Issue:**
+-  **Issue:**
 
-  When using approve_sequence option in any tier.definition there can be
-  inconsistencies in the systray notifications.
+   When using approve_sequence option in any tier.definition there can
+   be inconsistencies in the systray notifications.
 
-  **Description:**
+   **Description:**
 
-  Field can_review in tier.review is used to filter out, in the systray
-  notifications, the reviews a user can approve. This can_review field
-  is updated **in the database** in method review_user_count, this can
-  make it very inconsistent for databases with a lot of users and
-  recurring updates that can change the expected behavior.
+   Field can_review in tier.review is used to filter out, in the systray
+   notifications, the reviews a user can approve. This can_review field
+   is updated **in the database** in method review_user_count, this can
+   make it very inconsistent for databases with a lot of users and
+   recurring updates that can change the expected behavior.
 
-- **Migration to 15.0:**
+-  **Default \_tier_validation_manual_config parameter**
 
-  The parameter \_tier_validation_manual_config will become False, on
-  14.0, the default value is True, as the change is applied after the
-  migration. In order to use the new behavior we need to modify the
-  value on our expected model.
+   The parameter \_tier_validation_manual_config will become False, on
+   18.0, the default value is True, as the change is applied after the
+   migration. In order to use the new behavior we need to modify the
+   value on our expected model.
+
+-  **Extraneous functions** All functions:
+
+   \_get_to_validate_message_name
+
+   \_get_to_validate_message
+
+   \_get_validated_message
+
+   \_get_rejected_message
+
+   are related to message building should be moved into their own
+   module. This module is already heavy enough as it is
+
+-  **Obsolete fields:**
+
+   The fields "rejected" and "validated" are kept for compatibility
+   reasons and should be removed in 19.0
+
+-  **Usability and architecture**
+
+   Implement the good feedback listed here about usability
+   https://github.com/OCA/server-ux/pull/1097#pullrequestreview-2961078706
+
+-  **More cleanup**
+
+   There are too many computes on tier reviews as evidenced by the
+   remaining invalidate\_\* function calls in tests. It would be simpler
+   to manage if these functions were in their own standalone function
 
 Changelog
 =========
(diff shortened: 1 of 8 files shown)


kevinkhao:
@len-foss @StefanRijnhart @LoisRForgeFlow could you review ?

base_tier_validation/README.rst
```diff
@@ -120,12 +120,34 @@ improvement will be very valuable.
    make it very inconsistent for databases with a lot of users and
    recurring updates that can change the expected behavior.
 
--  **Migration to 15.0:**
+-  **Extraneous functions** All functions:
 
-   The parameter \_tier_validation_manual_config will become False, on
-   14.0, the default value is True, as the change is applied after the
-   migration. In order to use the new behavior we need to modify the
-   value on our expected model.
```
StefanRijnhart: We still want to change the default for `_tier_validation_manual_config`

base_tier_validation/README.rst
```diff
@@ -120,12 +120,34 @@ improvement will be very valuable.
...
+
+   \_get_rejected_message
+
+   are related to message building should be moved into their own
+   module. This module is already heavy enough as it is
+
+-  **Obsolete fields:**
+
+   The fields "rejected" and "validated" are kept for compatibility
+   reasons and should be removed ASAP
```
StefanRijnhart: `the next migration` or `in 19.0` rather than `ASAP`.

base_tier_validation/models/tier_validation.py
```diff
@@ -63,6 +62,8 @@ class TierValidation(models.AbstractModel):
         ],
         default="no",
         compute="_compute_validation_status",
+        compute_sudo=True,
```
StefanRijnhart: `True` is the default for `compute_sudo` of stored fields, so this can be removed.

base_tier_validation/models/tier_validation.py
```diff
@@ -192,29 +203,39 @@ def _get_rejected_message(self):
...
+    @api.depends("validation_status")
     def _compute_validated_rejected(self):
         for rec in self:
-            rec.validated = self._calc_reviews_validated(rec.review_ids)
-            rec.validated_message = rec._get_validated_message()
-            rec.rejected = self._calc_reviews_rejected(rec.review_ids)
-            rec.rejected_message = rec._get_rejected_message()
-            rec.to_validate_message = rec._get_to_validate_message()
+            for field in ("validated", "rejected"):
+                if rec.validation_status == field:
```
StefanRijnhart: This is already a boolean expression.

```
rec[field] = rec.validation_status == field
```

base_tier_validation/models/tier_validation.py
```diff
@@ -192,29 +203,39 @@ def _get_rejected_message(self):
...
+    @api.depends(lambda self: self._get_validation_status_dependencies())
     def _compute_validation_status(self):
         for item in self:
-            if item.validated and not item.rejected:
+            validated = self._calc_reviews_validated(item.review_ids)
+            rejected = self._calc_reviews_rejected(item.review_ids)
+            if validated and not rejected:
                 item.validation_status = "validated"
-            elif not item.validated and item.rejected:
+            elif not validated and rejected:
```
StefanRijnhart: `elif rejected:`

base_tier_validation_forward/tests/test_tier_validation.py
```diff
@@ -64,4 +64,4 @@ def test_01_forward_tier(self):
         wizard.comment = "Forward tier is reviewed"
         wiz = wizard.save()
         wiz.add_comment()
-        self.assertTrue(record.validated)
+        self.assertEqual(record.validation_status, "validated")
```
StefanRijnhart: Does this actually fix a breaking test, or is this a matter of stop using the deprecated fields?

StefanRijnhart (changes requested):
Please mention module name(s) in the precommit commit.

The offending, performance killing implementation of the rejected and validated fields are still there. It should be very easy to fix them by making them return a domain that includes a clause on the validation_status field.

base_tier_validation/models/tier_validation.py
```diff
@@ -192,29 +203,39 @@ def _get_rejected_message(self):
...
                 item.validation_status = "rejected"
             elif (
-                not item.validated
-                and not item.rejected
+                not validated
+                and not rejected
```
StefanRijnhart: Am I missing something or can these conditions be dropped as well given that `not validated and not rejected` are already handled before the `elif`?

base_tier_validation/README.rst
```diff
@@ -90,40 +90,64 @@ To configure Tier Validation Exceptions, you need to:
...
+   make it very inconsistent for databases with a lot of users and
+   recurring updates that can change the expected behavior.
 
-- **Migration to 15.0:**
+-  **Extraneous functions** All functions:
 
-  The parameter \_tier_validation_manual_config will become False, on
-  14.0, the default value is True, as the change is applied after the
-  migration. In order to use the new behavior we need to modify the
-  value on our expected model.
```
StefanRijnhart, line 130: We still want to change default for `_tier_validation_manual_config` in the next migration.

base_tier_validation/README.rst
kevinkhao: restored

base_tier_validation/README.rst
kevinkhao: modified

base_tier_validation/models/tier_validation.py
kevinkhao: removed

base_tier_validation_forward/tests/test_tier_validation.py
kevinkhao: stop using deprecated fields. On second look this was unnecessary and I have removed it

kevinkhao:
thanks @StefanRijnhart for the feedback
 
> The offending, performance killing implementation of the rejected and validated fields are still there. It should be very easy to fix them by making them return a domain that includes a clause on the validation_status field.

indeed! had a bit of tunnel vision there...

I have modified the readmes/code according to your feedback here https://github.com/OCA/server-ux/pull/1101/commits/e675156470b86bd315bf16473444fa9f06556b2c

base_tier_validation/models/tier_validation.py
```diff
@@ -134,25 +142,15 @@ def _compute_reviewer_ids(self):
     def _search_validated(self, operator, value):
         assert operator in ("=", "!="), "Invalid domain operator"
```
len-foss, line 144: Really small change, but I would also add the deprecation note here (and `_search_rejected`). Migrations can be quite mechanical, but if you remove the field and leave the function it will be extra-confusing and risk to get a "delete at migration 20.0" comment.

len-foss (approved):
The only downside that I see is that it changes computation triggers, which could cause some issues. However, current triggers are wrong and that could also cause bugs, so it's better to fix bugs in a correct version.

kevinkhao:
@len-foss thx for review, feedback integrated

base_tier_validation/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Base Tier Validation",
     "summary": "Implement a validation process based on tiers.",
-    "version": "18.0.1.1.3",
+    "version": "18.0.1.1.4",
```
StefanRijnhart: Pleasse don't bump the version number in the PR. Ocabot will take care of that during the merge.

StefanRijnhart (commented):
Thanks for all the fixes!
Adding a small test for the search methods might help with the code coverage, even though it is not blocking.
Please squash all commits (incl. pre-commit).
Suggest to rename the commit to `[IMP] base_tier_validation: deprecate rejected/validated fields (for performance)`.

kevinkhao:
@StefanRijnhart processed here https://github.com/OCA/server-ux/pull/1101/commits/40e567c22cef20e9dc2a28729298079de86e0718

base_tier_validation/tests/test_tier_validation.py
```diff
@@ -135,11 +142,14 @@ def test_10_systray_counter(self):
...
+        self.assertEqual(len(docs), 2)
         for doc in docs:
-            if doc.get("name") == "tier.validation.tester2":
+            if doc.get("name") == "Tier Validation Tester 2":
                 self.assertEqual(doc.get("pending_count"), 1)
-            else:
+            elif doc.get("name") == "Tier Validation Tester":
                 self.assertEqual(doc.get("pending_count"), 2)
+            else:
+                raise ValidationError(self.env._("This scenario should not happen"))
```
StefanRijnhart: This line still raises concerns by codecov because it is not covered. Given that it's test code, this should be prevented. I think you can replace it with a check at the top of the loop
```
self.assertIn(doc.get("name"), ("Tier Validation Tester", "Tier Validation Tester 2"))
```

StefanRijnhart (commented):
Thanks for adding a test. Please check the codecov warning I tried to clarify for you. You can also squash commits. As long as you don't rebase, it will still be clear what has changed.

StefanRijnhart (commented):
Thanks, please fix precommit and squash commits.

StefanRijnhart (dismissed):
Great work, thanks!

base_tier_validation/models/tier_validation.py
```diff
@@ -192,31 +194,31 @@ def _get_rejected_message(self):
...
+
+    # TODO: delete in 19.0 migration in favor of validation_status field
+    @api.depends("validation_status")
     def _compute_validated_rejected(self):
         for rec in self:
-            rec.validated = self._calc_reviews_validated(rec.review_ids)
-            rec.validated_message = rec._get_validated_message()
-            rec.rejected = self._calc_reviews_rejected(rec.review_ids)
-            rec.rejected_message = rec._get_rejected_message()
-            rec.to_validate_message = rec._get_to_validate_message()
```
StefanRijnhart, line 201: You need to keep on assiging all the values above, because the *message fields still have this compute method.

StefanRijnhart (commented):
There is a competing PR at https://github.com/OCA/server-ux/pull/1084, with some minor differences, but also, it has a migration script to populate the stored fields in a performant way. That is probably a must have. We also need to discuss the details over there.

StefanRijnhart:
Please check https://github.com/OCA/server-ux/pull/1129 for a port of #1084. I did take this PR's cache fix.

PR #1102, opened 2025-06-28
[18.0][ADD] server_action_wizard

dreispt:


server_action_wizard/models/ir_actions_server.py
```diff
@@ -0,0 +1,11 @@
+from odoo import fields, models
+
+
+class IrActionsServer(models.Model):
+    _inherit = "ir.actions.server"
+
+    can_run_from_wizard = fields.Boolean("Can run from wizard")
```
legalsylvain, line 7: Is it possible to have something configurable ?

I see that 4 fields quite custom...

dreispt: The 4 fields are for technical simplicity. It would be easy to add more if that makes sense.
I don't see how it even more generic.

legalsylvain: It will display fields in the wizard, even if it doenst make sense. 

I dont see it very user friendly, or did I miss something ? Could you provide use case with real exemple in the readme section ? (Based on démo data could be great !)

> I don't see how it even more generic.

Using fields.Properties could make the job I guess.

Thanks !

legalsylvain (changes requested):
A remark inline. 

Otherwise, could you add tests ?

Thanks !

andreampiovesana:
who can fix?

server_action_wizard/models/ir_actions_server.py
dreispt: Look at `server_action_wizard_views.xml`: the wizard only shows the fields that have a label set.

With only the  `base` app I couldn't think of an interesting enough example.
I did improve the README.

I like the Properties idea, but it requires some research and I won't have time for it right now.
Good for a follow up improvement PR.

PR #1111, opened 2025-07-01
[17.0][FIX] base_tier_validation: Proper notifications on reviews

victoralmau:
FWP from 16.0: https://github.com/OCA/server-ux/pull/1083

Proper notifications on reviews

Changes done:
- The subtypes should not be defined as default=True to avoid that a follower receives notifications of everything (even if he/she doesn't really need it).
- When adding followers they are added to the corresponding subtype so that the notification arrives to whom it corresponds.
- Improve the code so that the reviews to notify are the "following ones according to the sequence".

@Tecnativa TT56354

### DIFF
--- a/base_tier_validation/data/mail_data.xml
+++ b/base_tier_validation/data/mail_data.xml
@@ -6,7 +6,7 @@
         forcecreate="1"
     >
         <field name="name">Tier Validation Requested</field>
-        <field name="default" eval="True" />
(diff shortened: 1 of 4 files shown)


LoisRForgeFlow (approved):
:+1:

pedrobaeza (approved):
/ocabot merge major

PR #1115, opened 2025-07-01
[17.0]FIX] base_export_manager: add sudo call to models without explicit access

victoralmau:
FWP from 16.0: https://github.com/OCA/server-ux/pull/1113

Changes done:
- [x] Add compute_sudo=True to `label` field to prevent access error
- [x] Improve test with basic test-user

Please @pedrobaeza and  @carlosdauden can you review it?

@Tecnativa TT56933


pedrobaeza (approved):
/ocabot merge patch

PR #1117, opened 2025-07-02
[18.0][FIX] base_tier_validation: make ReviewsTable widget work with manual config

MaxyMoos:
Forward-port of #1093 , since the original issue can be reproduced in v18 too.


LoisRForgeFlow (approved):
Thanks for helping to keep versions aligned :+1:

PR #1120, opened 2025-07-09
[17.0][IMP] base_tier_validation: add support for custom actions in TierReviewMenu

MaxyMoos:
This commit adds the possibility for modules inheriting from `base_tier_validation` to customize the `ir.actions.act_window` to run when a user clicks on the TierReviewMenu entries.

For example, the current behavior when using `account_move_tier_validation` is to use the default views for `account.move`, which leads users to a Kanban view of the Moves by default, and the Tree/List view has no "Partner" column.
With this PR, adding a small JS module such as:
```js
/* @odoo-module */
import { patch } from "@web/core/utils/patch";
import { TierReviewMenu } from "@base_tier_validation/components/tier_review_menu/tier_review_menu.esm";


patch(TierReviewMenu.prototype, {
    async getCustomAction(group) {
        if (group.model === 'account.move') {
            const action = 'account.action_move_journal_line';
            return this.action.loadAction(action);
        }
        return super.getCustomAction(group);
    }
});
```

would then open the classic "Journal Entries" default views, showing the "Partner" column in the tree view, etc...

Of course, the default behavior of this addition is to fallback on the current approach.
Also, we now pass the `group` variable to the `getAvailableViews` call so this can be customized too.

### DIFF
--- a/base_tier_validation/static/src/components/tier_review_menu/tier_review_menu.esm.js
+++ b/base_tier_validation/static/src/components/tier_review_menu/tier_review_menu.esm.js
@@ -29,7 +29,8 @@ export class TierReviewMenu extends Component {
     onBeforeOpen() {
         this.fetchSystrayReviewer();
     }
-    availableViews() {
+    // eslint-disable-next-line no-unused-vars
+    availableViews(group) {
         return [
             [false, "kanban"],
             [false, "list"],
@@ -37,7 +38,14 @@ export class TierReviewMenu extends Component {
             [false, "activity"],
         ];
     }
-    openReviewGroup(group) {
+    async getCustomAction(group) {
+        if (group.model) {
+            const action = await this.orm.call(group.model, "get_tier_review_action");
+            return action || false;
+        }
+        return false;
+    }
+    async openReviewGroup(group) {
         document.body.click();
         // Hack to close dropdown
         const context = {};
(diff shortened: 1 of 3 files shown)


MaxyMoos:
Hi @LoisRForgeFlow, just bumping this as it seems it'll get closed automatically real soon. Would love to get your opinion on whether this improvement seems relevant whenever you get some time.

base_tier_validation/static/src/components/tier_review_menu/tier_review_menu.esm.js
```diff
@@ -29,38 +29,58 @@ export class TierReviewMenu extends Component {
...
             [false, "kanban"],
             [false, "list"],
             [false, "form"],
             [false, "activity"],
         ];
     }
-    openReviewGroup(group) {
+    // eslint-disable-next-line no-unused-vars
+    async getCustomAction(group) {
+        return false;
```
LoisRForgeFlow, line 46: Hi @MaxyMoos 

Sorry for missing this one. It LGTM, the only comment I have is that it might be a good idea to have a python method that is called here to get the action. This will make inheritance even easier.

WDYT?

MaxyMoos: Hi @LoisRForgeFlow ,

Thanks for checking out the PR. Yes, one additional RPC call is probably not a big deal on performance, and much better for Odoo devs, i've implemented this and force-pushed.
Does this new version look good to you?

A custom module would then just need to add such code:
```python
from odoo import api, models

class AccountMove(models.Model):
    _inherit = "account.move"
    
    @api.model
    def get_tier_review_action(self):
        action = self.env["ir.actions.actions"]._for_xml_id(
            "account.action_move_journal_line"
        )
        return action
```

LoisRForgeFlow (approved):
Thanks! LGTM 👍

MaxyMoos:
Just pushed a small update : i added the override example into the `base` model method docstring so the syntax is documented in code and not just in a GitHub comment ;)

MaxyMoos:
@LoisRForgeFlow Does this need another reviewer chiming in before we could consider merging it?

LoisRForgeFlow:
> @LoisRForgeFlow Does this need another reviewer chiming in before we could consider merging it?

@MaxyMoos that would be ideal, from my side looks good

bosd (approved):

PR #1124, opened 2025-07-16
[15.0][ADD] web_hide_custom_search

AungKoKoLin1997:
This module allows controlling the visibility of the “Add Custom Filter” and “Add Custom Group” options in the search view for specific models and specific user groups.

@qrtl QT4975


pedrobaeza:
Shouldn't it start with `web_`?

web_hide_custom_filter/models/ir_model.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2025 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class IrModel(models.Model):
+    _inherit = "ir.model"
+
+    show_custom_filter_groupby_group_ids = fields.Many2many(
```
yostashiro: Let's keep it short.
```suggestion
    custom_search_group_ids = fields.Many2many(
```

web_hide_custom_filter/models/ir_model.py
```diff
@@ -0,0 +1,26 @@
...
+        "res.groups",
+        string="Show Custom Filter/Group-by for Groups",
+        relation="show_custom_filter_groupby_group_rel",
+        help="Only users in these groups will see the 'Add Custom Filter' "
```
yostashiro:
```suggestion
        help="If set, only users in these groups will see the 'Add Custom Filter' "
```

web_hide_custom_filter/models/res_groups.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2025 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResGroups(models.Model):
+    _inherit = "res.groups"
+
+    show_custom_filter_groupby_model_ids = fields.Many2many(
```
yostashiro:
```suggestion
    custom_search_model_ids = fields.Many2many(
```

yostashiro (commented):
Partial review.

Module name suggestion: web_hide_custom_search

Sorry, I just came to think that 'custom search' would cover both filter and group-by.

web_hide_custom_search/static/src/js/custom_filter_patch.esm.js
```diff
@@ -0,0 +1,24 @@
...
+    const CustomFilterItem = require("web.CustomFilterItem");
+    const {patch} = require("web.utils");
+    const {hooks, useState} = owl;
+    const {onWillStart} = hooks;
+    const rpc = require("web.rpc");
+
+    patch(CustomFilterItem.prototype, "web_hide_custom_search.CustomFilterItemPatch", {
+        setup() {
+            this._super.apply(this, arguments);
+            this.visibleState = useState({isVisible: true});
```
yostashiro, line 11: This may be more in line with the usual convention?
```suggestion
            this.state = useState({isVisible: true});
```

yostashiro (commented):
Code LGTM. Could we add tests?

web_hide_custom_search/static/src/js/custom_filter_patch.esm.js
AungKoKoLin1997: I checked using state, but since it's initialized inside the constructor, even if we try to add visible inside setup(), it only contains the original initialized value. So, I’ll just use a new variable instead.

AungKoKoLin1997:
> Could we add tests?

@yostashiro Added some test cases.

yostashiro (approved):
LGTM.

kanda999 (approved):
Functional test: it works as expected

dreispt (approved):

AungKoKoLin1997:
It seems that every test using `self.start_tour` fails in 15.0 for some reason. The same issue occurs in the e-commerce repo.
https://github.com/OCA/e-commerce/pull/1103

@dreispt Do you have any idea related this issue?

kanda999 (changes requested):
When I press the "+" button in pivot view, I will get the error.
This error comes from this module.

2025-08-26_13h48_52

AungKoKoLin1997:
@kanda999 Fixed it. Please review.

kanda999 (approved):

AungKoKoLin1997:
> It seems that every test using `self.start_tour` fails in 15.0 for some reason. The same issue occurs in the e-commerce repo. [OCA/e-commerce#1103](https://github.com/OCA/e-commerce/pull/1103)
> 
> @dreispt Do you have any idea related this issue?

@rvalyi This issue still exists.

AungKoKoLin1997:
@rvalyi Seem like the CI issue is already fixed.
Could you please try to merge again?

PR #1130, opened 2025-07-28
[18.0][MIG] default_multi_user: Migration to 18.0

ArnauCForgeFlow:
Standard Migration to version 18.0


LoisRForgeFlow (changes requested):
When migrating modules with hooks that modify original methods, you need to make sure that nothing is changes apart from what's strictly the purpose (in this case highlighted with comments `# START/END OF CHANGES`. Apart from that there is no intention of modifying the orignal method. In v18 the method changed slightly, and that change needs to be also included here: https://github.com/odoo/odoo/blob/18.0/odoo/addons/base/models/ir_default.py#L159

ArnauCForgeFlow:
@LoisRForgeFlow Done!

LoisRForgeFlow (approved):
Functional review, works as expected :+1:

JasminSForgeFlow (approved):
Code Review

PR #1135, opened 2025-07-31
[16.0][FIX] base_tier_validation: Define the appropriate condition for validation_status=validated

victoralmau:
Define the appropriate condition for validation_status=validated

Related to https://github.com/OCA/server-ux/pull/1084#discussion_r2244852473

Please @LoisRForgeFlow can you review it?

@Tecnativa

PR #1129, opened 2025-07-23
[18.0][IMP] base_tier_validation: store status fields for (faster) searching

StefanRijnhart:
Previous approach in `_search_validated` and `_search_rejected` could easily lead to a MemoryError on large instances.

Port of https://github.com/OCA/server-ux/pull/1084, plus a cache fix from https://github.com/OCA/server-ux/pull/1101

PR #1136, opened 2025-08-04
[17.0][IMP] base_tier_validation: Make the validation_status field store so that it can be used in searchs

victoralmau:
FWP from 16.0: https://github.com/OCA/server-ux/pull/1084 + https://github.com/OCA/server-ux/pull/1135

@Tecnativa TT56391


base_tier_validation/models/tier_validation.py
```diff
@@ -184,37 +154,61 @@ def _get_validated_message(self):
...
+            if (
+                reviews
+                and not any(
+                    reviews.filtered(lambda x: x.status not in validated_states)
+                )
+                and not any_rejected
+            ):
                 item.validation_status = "validated"
-            elif not item.validated and item.rejected:
+            elif reviews and not any_approved and any_rejected:
```
StefanRijnhart: It used to be the case an item was rejected if it is approved by one reviewer but rejected by another. That is what is still in the migration script as well. Why did this change?

victoralmau: This is what was validated and merged at https://github.com/OCA/server-ux/pull/1084

Changing the condition to `elif reviews and any_rejected` seems appropriate to me. I don't think it will ever happen that one review is rejected and then another is approved (in that case, it would always remain in rejected status). Do you agree, @LoisRForgeFlow? If this change is finally merge, I will also apply it to 16.0.

StefanRijnhart: Thanks for the change! I did not get a chance to review the latest changes to #1084, so I commented about this here.

base_tier_validation/migrations/17.0.3.1.0/pre-migration.py
```diff
@@ -0,0 +1,112 @@
...
+                f"""
+                UPDATE {table_name} SET validation_status = 'pending'
+                WHERE validated = false
+                AND coalesce(rejected, false) = false AND id IN (
+                    SELECT DISTINCT(tr.res_id)
+                    FROM tier_review AS tr
+                    WHERE tr.model = '{model_name}' AND tr.status = 'pending'
+                )
+                """,
+            )
```
StefanRijnhart, line 79: For completeness sake, you might want to take my `waiting` query from https://github.com/OCA/server-ux/pull/1129/commits/4cc5921b41fe8407ae0f5b554dbdde6fc1f77a90#diff-a411b29cdb4152dc1e72251852b43693542da5717280aae4cee9395514cbe1a5R112-R122, because this is handled so in the compute method (although validations with a waiting review but no pending review should only happen with pending cache operations).

base_tier_validation/models/tier_validation.py
```diff
@@ -184,37 +154,61 @@ def _get_validated_message(self):
...
+            if (
+                reviews
+                and not any(
+                    reviews.filtered(lambda x: x.status not in validated_states)
+                )
+                and not any_rejected
+            ):
                 item.validation_status = "validated"
-            elif not item.validated and item.rejected:
+            elif reviews and any_rejected:
```
StefanRijnhart: `reviews` is implied by `any_rejected`
```suggestion
            elif any_rejected:
```

base_tier_validation/models/tier_validation.py
```diff
@@ -184,37 +154,61 @@ def _get_validated_message(self):
...
                 item.validation_status = "rejected"
-            elif (
-                not item.validated
-                and not item.rejected
-                and any(item.review_ids.filtered(lambda x: x.status == "pending"))
-            ):
+            elif reviews and not any_approved and not any_rejected and any_pending:
```
StefanRijnhart: Here too, if there are any pending reviews but also an approved review, status should still be pending.
```suggestion
            elif any_pending:
```

base_tier_validation/models/tier_validation.py
```diff
@@ -184,37 +154,61 @@ def _get_validated_message(self):
...
                 item.validation_status = "pending"
             elif (
-                not item.validated
-                and not item.rejected
+                not any_approved
+                and not any_rejected
                 and any(item.review_ids.filtered(lambda x: x.status == "waiting"))
```
StefanRijnhart: Can also be simplified. Maybe also add an any_waiting variable.
```suggestion
                any(item.review_ids.filtered(lambda x: x.status == "waiting"))
```

base_tier_validation/models/tier_validation.py
```diff
@@ -184,37 +154,61 @@ def _get_validated_message(self):
...
+            reviews = item.review_ids
+            any_approved = any(reviews.filtered(lambda x: x.status in validated_states))
+            any_rejected = any(reviews.filtered(lambda x: x.status in rejected_states))
+            any_pending = any(reviews.filtered(lambda x: x.status == "pending"))
+            if (
+                reviews
+                and not any(
+                    reviews.filtered(lambda x: x.status not in validated_states)
+                )
+                and not any_rejected
```
StefanRijnhart: More readable:
```suggestion
                and all(x.status in validated_states for x in reviews)
```

base_tier_validation/migrations/17.0.3.1.0/pre-migration.py
StefanRijnhart: Thanks for the update, but we need to keep the 'pending' query as well (before the 'waiting' query).

StefanRijnhart (changes requested):

StefanRijnhart (commented):
Thanks, looking good in principle but I now see the amount of follow-up PRs needed to cover for the removal of the validated and rejected fields in 16.0. Should we perhaps consider to keep some version of them for backwards compatibility until Odoo 19?

pedrobaeza:
Debated here: https://github.com/OCA/hr-expense/pull/311#issuecomment-3150753075

Let's undo partially the removal, and sorry for the noise and inconveniences.

victoralmau:
> Debated here: [OCA/hr-expense#311 (comment)](https://github.com/OCA/hr-expense/pull/311#issuecomment-3150753075)
> 
> Let's undo partially the removal, and sorry for the noise and inconveniences.

Ok, the validated and rejected fields remain.

victoralmau:
Changes from 16.0 applied (https://github.com/OCA/server-ux/pull/1137#discussion_r2253879731 + https://github.com/OCA/server-ux/pull/1137#discussion_r2253969750)

StefanRijnhart (approved):

LoisRForgeFlow (approved):
LGTM, much cleaner now. Thanks @StefanRijnhart for the good review and @victoralmau for the quick work.

LoisRForgeFlow:
@pedrobaeza @StefanRijnhart anything else before merging?

StefanRijnhart:
@LoisRForgeFlow Agree to merge on my part!

base_tier_validation/migrations/17.0.3.1.0/pre-migration.py
```diff
@@ -0,0 +1,122 @@
...
+    )
+    for (model_name,) in env.cr.fetchall():
+        table_name = model_name.replace(".", "_")
+        # validated column
+        if not openupgrade.column_exists(env.cr, table_name, "validated"):
+            openupgrade.logged_query(
+                env.cr,
+                f"""
+                ALTER TABLE {table_name}
+                ADD COLUMN IF NOT EXISTS validated BOOLEAN
```
pedrobaeza: Why are we creating the columns if the field is non stored computed? AFAIK, only `validation_status` should be created.

victoralmau: This is done (since v16) to simplify the logic in SQL queries and make everything clearer.

StefanRijnhart: Ah yes, this made sense in an earlier iteration but I agree it should probably be refactored out at this point.

victoralmau: So, just to be clear, we need to redo all these queries to merge them into one? (I assume it will be tested sufficiently to avoid merging something that contains errors and blocks everyone, because this is already merged in v16 and is OK).

StefanRijnhart: You don't need to merge the queries into one, you can start out with 'no' values everywhere, then check per query for items set to 'no' with any rejected, pending, waiting reviews, then lastly for (all) approved+forwarded reviews.

StefanRijnhart (approved):
Thanks for improving the migration script! I tested the new and the previous version and results were identical.

carlos-lopez-tecnativa (approved):
LGTM

I left a comment for my future self to remember this when the migration to V19 is done, but if you can change the other uses of the fields validated and rejected in other modules now, that would be great.

https://github.com/OCA/sale-workflow/blob/cc23ae094366ec2701cb2062ff93aca4d67a3927/sale_tier_validation/reports/sale_report.py#L18
https://github.com/OCA/sale-workflow/blob/cc23ae094366ec2701cb2062ff93aca4d67a3927/sale_tier_validation/views/sale_order_view.xml#L20
https://github.com/OCA/purchase-workflow/blob/03545c5ea9d4a07fce4f4ecfa7b412ae701ab8ef/purchase_request_tier_validation/views/purchase_request_view.xml#L48
https://github.com/OCA/purchase-workflow/blob/03545c5ea9d4a07fce4f4ecfa7b412ae701ab8ef/purchase_requisition_tier_validation/views/purchase_requisition_view.xml#L22
https://github.com/OCA/purchase-workflow/blob/03545c5ea9d4a07fce4f4ecfa7b412ae701ab8ef/purchase_tier_validation/views/purchase_order_view.xml#L21

pedrobaeza (approved):
/ocabot merge major

StefanRijnhart:
@victoralmau will you do the port to 18 as well?

PR #1137, opened 2025-08-04
[16.0][FIX+IMP] base_tier_validation: base_tier_validation_waiting compatibiity

victoralmau:
Changes done:
- [x] [OU-IMP+OU-FIX]: base_tier_validation: Change/improve the migration script as discussed in https://github.com/OCA/server-ux/pull/1136#discussion_r2251063342 + compatibiity with `base_tier_validation_waiting`.
- [x] base_tier_validation_waiting: Compatibility with validation_status=waiting (similar to what already happens in 17.0).
- [x] `base_tier_validation`: Misc changes: https://github.com/OCA/server-ux/pull/1136#discussion_r2250810533 + https://github.com/OCA/server-ux/pull/1136#discussion_r2251067906 + https://github.com/OCA/server-ux/pull/1136#discussion_r2251075645 + https://github.com/OCA/server-ux/pull/1136#discussion_r2251085195
- [x] `base_tier_validation`: The validated and rejected fields are restored. Related to https://github.com/OCA/server-ux/pull/1136#issuecomment-3151284585

@Tecnativa

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -42,9 +42,17 @@ class TierValidation(models.AbstractModel):
         domain=lambda self: [("model", "=", self._name)],
         auto_join=True,
     )
+    # TODO: Delete in v19 in favor of validation_status field
+    validated = fields.Boolean(
+        compute="_compute_validated_rejected", search="_search_validated"
+    )
     to_validate_message = fields.Html(compute="_compute_to_validate_message")
     validated_message = fields.Html(compute="_compute_validated_message")
     need_validation = fields.Boolean(compute="_compute_need_validation")
+    # TODO: Delete in v19 in favor of validation_status field
+    rejected = fields.Boolean(
+        compute="_compute_validated_rejected", search="_search_rejected"
+    )
     rejected_message = fields.Html(compute="_compute_rejected_message")
     validation_status = fields.Selection(
         selection=[
@@ -121,6 +129,22 @@ class TierValidation(models.AbstractModel):
                 lambda r: r.status == "pending"
             ).mapped("reviewer_ids")
 
+    # TODO: delete in 19.0 migration in favor of validation_status field
+    @api.model
+    def _search_validated(self, operator, value):
+        assert operator in ("=", "!="), "Invalid domain operator"
+        assert value in (True, False), "Invalid domain value"
+        operator_equal = (operator == "=" and value) or (operator == "!=" and not value)
+        return [("validation_status", operator_equal and "=" or "!=", "validated")]
+
+    # TODO: delete in 19.0 migration in favor of validation_status field
+    @api.model
+    def _search_rejected(self, operator, value):
+        assert operator in ("=", "!="), "Invalid domain operator"
+        assert value in (True, False), "Invalid domain value"
+        operator_equal = (operator == "=" and value) or (operator == "!=" and not value)
+        return [("validation_status", operator_equal and "=" or "!=", "rejected")]
+
     @api.model
     def _search_reviewer_ids(self, operator, value):
         model_operator = "in"
@@ -160,6 +184,13 @@ class TierValidation(models.AbstractModel):
             """Operation has been <b>rejected</b>."""
         )
         return self.validation_status == "rejected" and msg or ""
+
+    # TODO: delete in 19.0 migration in favor of validation_status field
+    @api.depends("validation_status")
+    def _compute_validated_rejected(self):
+        for rec in self:
+            for field in ("validated", "rejected"):
+                rec[field] = rec.validation_status == field
 
     @api.depends("validation_status")
     def _compute_to_validate_message(self):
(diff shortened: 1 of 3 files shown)


base_tier_validation/models/tier_validation.py
```diff
@@ -121,6 +129,24 @@ def _compute_reviewer_ids(self):
...
             ).mapped("reviewer_ids")
 
+    @api.model
+    def _search_validated(self, operator, value):
+        assert operator in ("=", "!="), "Invalid domain operator"
+        assert value in (True, False), "Invalid domain value"
+        pos = self.search(
+            [(self._state_field, "in", self._state_from), ("review_ids", "!=", False)]
+        ).filtered(lambda r: r.validated == value)
+        return [("id", "in", pos.ids)]
```
StefanRijnhart: Why does `_state_from` come into play all of a sudden?

It should be more efficient to return the original domain rather than the id based domain after the search.

base_tier_validation_waiting/models/tier_validation.py
```diff
@@ -51,3 +56,11 @@ def _validate_tier(self, tiers=False):
         self._validate_tier_waiting_reviews(tiers)
         res = super(TierValidation, self)._validate_tier(tiers)
         return res
+
+    def _compute_validation_status(self):
+        res = super()._compute_validation_status()
+        for item in self:
```
StefanRijnhart: The waiting state should only be applied if not approved, rejected, pending.

```suggestion
        for item in self.filtered(lambda tv: tv.validation_status == "no"):
```

StefanRijnhart (changes requested):

base_tier_validation/models/tier_validation.py
victoralmau: These methods are the same as those that existed before https://github.com/OCA/server-ux/pull/1084. Now, in version 17, there is something similar: https://github.com/OCA/server-ux/blob/ba7cf79ee7aadd7c9f78d27030bd1acfad4073ea/base_tier_validation/models/tier_validation.py#L133 
Personally, I have no problem changing them and making them depend directly on the `validation_status` field, but if they are to be kept for “_compatibility_” reasons, I think it is best to avoid any confusion and use them “_as before_.”

StefanRijnhart: Oh, you took the original search methods. The crippling performance of these were the main reason we submitted https://github.com/OCA/server-ux/pull/1101. Please take the methods from this PR (https://github.com/OCA/server-ux/pull/1101/files#diff-5b72a073de187894d5d5401d6574d08eb57b1fcf07fcaa0ad1682b09dd65bc65R143-R147).

base_tier_validation/models/tier_validation.py
```diff
@@ -161,6 +187,11 @@ def _get_rejected_message(self):
         )
         return self.validation_status == "rejected" and msg or ""
 
+    def _compute_validated_rejected(self):
+        for rec in self:
+            rec.validated = self._calc_reviews_validated(rec.review_ids)
+            rec.rejected = self._calc_reviews_rejected(rec.review_ids)
```
StefanRijnhart: Here too, the value should be derived from validation_status. I'm surprised to see the `_calc_reviews` methods still exist as they are not used to compute the validation_status.

victoralmau: According to what is mentioned at https://github.com/OCA/server-ux/pull/1084#discussion_r2243345108, the `_calc_reviews` methods were maintained. Should they also be removed? Personally, I don't understand the need to “remove methods” in some cases (perhaps breaking compatibility if someone is using them), but on the other hand, it is necessary to keep these `validated`/`rejected` fields and/or not make them store (the migration scripts have already defined the appropriate values).

StefanRijnhart: I see. That comment seem to suggest to keep `_calc_reviews_*` for backwards compatibility, but change their implementation to check `validation_status`. You did not change the implementation, but given that `_calc_reviews` methods were meant to influence the value of `validation_status`, it would seem misleading to me. So I think they can be removed, and `validation_status` can be queried directly to compute `validated` and `rejected`.

StefanRijnhart: Or, alternatively you could have kept `_calc_reviews` to *determine* validation status, but it was decided to replace this functionality by the new methods `_validated_states` and `_rejected_states`. For full backwards compatibility, you could choose to revert this. It depends on whether the `_calc_reviews` methods are overridden regularly in depending modules I guess.

This was the approach taken in our own attempt: https://github.com/OCA/server-ux/pull/1101/files#diff-5b72a073de187894d5d5401d6574d08eb57b1fcf07fcaa0ad1682b09dd65bc65R211-R214

StefanRijnhart: (from what I can see, the only override of a `_calc_reviews` method was in base_tier_validation_forward and it was correctly refactored out in https://github.com/OCA/server-ux/pull/1084 already, so I think it's safe to remove the `_calc_reviews` methods entirely).

base_tier_validation/models/tier_validation.py
victoralmau: Okay, changed the `_search_validated()` + `_search_rejected()` methods as indicated in github.com/OCA/server-ux/pull/1101/files#diff-5b72a073de187894d5d5401d6574d08eb57b1fcf07fcaa0ad1682b09dd65bc65R143-R147

base_tier_validation/models/tier_validation.py
victoralmau: Okay, the methods `_calc_reviews_validated()` + `_calc_reviews_rejected()` have been removed and the method `_compute_validated_rejected()` has been modified as indicated in https://github.com/OCA/server-ux/pull/1101/files#diff-5b72a073de187894d5d5401d6574d08eb57b1fcf07fcaa0ad1682b09dd65bc65R201-R204

victoralmau: If this change and https://github.com/OCA/server-ux/pull/1137#discussion_r2254045634 are correct, I will apply them to PR 17.0.

StefanRijnhart (approved):
Looking good now!

pedrobaeza (approved):
/ocabot merge patch

PR #1138, opened 2025-08-06
[18.0][IMP] base_tier_validation: Make the validation_status field store so that it can be used in searchs

victoralmau:
FWP from 17.0: https://github.com/OCA/server-ux/pull/1136

Please @pedrobaeza and @carlos-lopez-tecnativa and @StefanRijnhart can you review it?

@Tecnativa TT56391


StefanRijnhart (approved):
Looking good, thanks!

pedrobaeza (approved):
/ocabot merge major

carlos-lopez-tecnativa (commented):
I have the same error with the migration script reported here: https://github.com/OCA/purchase-workflow/pull/2727#pullrequestreview-3073770181

image

base_tier_validation/models/tier_validation.py
```diff
@@ -232,18 +247,6 @@ def _compute_hide_reviews(self):
         for rec in self:
             rec.hide_reviews = rec[self._state_field] not in self._state_from
 
-    @api.model
-    def _calc_reviews_validated(self, reviews):
```
simahawk, line 236: :warning: these methods should have been deprecated.
It breaks https://github.com/OCA/stock-logistics-workflow/tree/18.0/stock_picking_tier_validation

https://github.com/OCA/stock-logistics-workflow/actions/runs/16695540235/job/47889988766#step:8:1030

@victoralmau can you please check that?

pedrobaeza: Fix in https://github.com/OCA/stock-logistics-workflow/pull/2065

Issue #1144, opened 2025-09-03
[18.0] reviews module – duplicate t-key error when multiple reviewers have same sequence


jaydeep32:
## Module
base_tier_validation

## Describe the bug
When a record has multiple reviewers assigned with the same sequence, opening the form view in debug mode raises an Owl client error due to duplicate keys in a t-foreach loop.

<img width="1116" height="350" alt="Image" src="https://github.com/user-attachments/assets/ae4958c4-40da-4ada-ba35-afb55be12200" />

<img width="952" height="444" alt="Image" src="https://github.com/user-attachments/assets/c19a264f-34b3-4070-bdb7-65dd92e3e065" />

`
Odoo Client Error
UncaughtPromiseError > OwlError
Uncaught Promise > Got duplicate key in t-foreach: 1
`

## To Reproduce
**Affected versions**:  18.0

1. Create a record that requires tier validation.
2. Add two or more reviewers with the same sequence number.
3. Open that record in form view.
4. Activate debug/developer mode.
5. The error appears immediately.

**Expected behavior**
The form view should display all reviewers correctly, regardless of whether they share the same sequence.

**Additional context**
For testing purposes, I updated the line in [base_tier_validation.Collapse](https://github.com/OCA/server-ux/blob/6211f08869c0f5f6306f197aa34135c0e6338bc8/base_tier_validation/static/src/components/tier_review_widget/tier_review_widget.xml#L50) to have `t-key='review.id'` instead of the existing value, and with this change the error no longer occurs. Based on this, I believe the issue is caused by the use of a non-unique key (such as `sequence`) in the `t-foreach` loop.

If this assumption is correct, then adjusting the template to use a unique identifier (e.g., `review.id` or a combination of fields) should resolve the error permanently.

However, if this is not the right approach and the problem lies elsewhere, I would appreciate guidance on the actual root cause and the recommended solution.

jaydeep32:
I have made one solution, MR, as well; please review and merge if anyone thinks that will work as a solution.
[#1147](https://github.com/OCA/server-ux/pull/1147)

LoisRForgeFlow:
Solved via https://github.com/OCA/server-ux/pull/1225

PR #1147, opened 2025-09-08
[18.0][IMP] base_tier_validation: Duplicate t-key error when multiple reviewers have same sequence

jaydeep32:
In server we noticed one BUG and we have already made issue on OCA
:- https://github.com/OCA/server-ux/issues/1144

- When a record has multiple reviewers assigned with the same sequence, opening the form view in debug mode raises an Owl client error due to duplicate keys in a t-foreach loop.

- Affected versions: 18.0

	- Steps to reproduce the behavior:
	- Create a record that requires tier validation.
	- Add two or more reviewers with the same sequence number.
	- Open that record in form view.
	- Activate debug/developer mode.
	- The error appears immediately.

As per lot of testing this seems perfect fix for this issue at the moment so I applied this in new branch.

### DIFF
--- a/base_tier_validation/static/src/components/tier_review_widget/tier_review_widget.xml
+++ b/base_tier_validation/static/src/components/tier_review_widget/tier_review_widget.xml
@@ -47,7 +47,7 @@
                                 <t
                                     t-foreach="_getReviewData()"
                                     t-as="review"
-                                    t-key="review.sequence"
+                                    t-key="review.id"
                                 >
                                     <t
                                         t-if="review.status == 'waiting'"


LoisRForgeFlow (commented):
@jaydeep32 Thanks for your patch!

Could you adjust your commit message to meet the [guidelines](https://www.odoo.com/documentation/12.0/developer/reference/guidelines.html#commit-message-structure)? for example: [IMP] base_tier_validation: short desription"

LoisRForgeFlow:
Closed in favor of https://github.com/OCA/server-ux/pull/1225

PR #1151, opened 2025-09-16
[18.0][IMP] base_tier_validation: Added password confirmation

ArnauCForgeFlow:
Added a new field to the tier definition that allows setting a password confirmation. Later, during tier validation, the system will ask the user to confirm their password in order to validate or reject. If the comment option is enabled, the password confirmation will be requested after the comment wizard.


LauraCForgeFlow (approved):
Functional testing LGTM :+1:

DavidJForgeFlow (approved):

JoanSForgeFlow (approved):

JoanSForgeFlow:
Hi @etobella ,

Could you please review this PR when you have a moment and merge it if everything looks good?

Thank you!

DavidJForgeFlow:
@pedrobaeza could you merge here if looks good to you?

pedrobaeza:
I don't think this is correct, as you are only counting if the users have Odoo login, but not OAuth or other forms of login, and besides, it's some kind of duplicated behavior, as you already logged in. I know the login is not closed, but putting this as a way to assure impersonation due to not closing the sessions seems weird. I wouldn't definitively put this on the base module.

kevinkhao:
Not sure if this is possible to do cleanly but splitting into another module base_tier_validation_password or somesuch will help keep base_tier_validation as small and targeted as possible seems a good idea. The module is already big.

etobella (changes requested):
Following @pedrobaeza comments, it might happen that the user doesn't have a password if they use passkeys (for example).

I would opt for a new module to add this functionality.

Also, review res.users.identitycheck, this might help you to make it usable in all scenarios 😉

LoisRForgeFlow:
Closed in favor of https://github.com/OCA/server-ux/pull/1213

PR #1152, opened 2025-09-17
[16.0][FIX] announcement: Avoid user login error

victoralmau:
Avoid user login error

Use case example:
- Create an announcement and link users A + B
- Modify the announcement and leave it linked only to user A
- Try to log in with user B

Previously, it was not possible to log in because user B still had the announcement linked (unread_announcement_ids) to their user, and when trying to log in, an attempt was made to access a record that the user could not access (except with Announcement Manager permission).

Please @david-banon-tecnativa and @CarlosRoca13 can you review it?

@Tecnativa TT58023

### DIFF
--- a/announcement/models/announcement.py
+++ b/announcement/models/announcement.py
@@ -228,8 +228,18 @@ class Announcement(models.Model):
         return records
 
     def write(self, vals):
-        """Adjust attachments for being accesible to receivers of the announcement."""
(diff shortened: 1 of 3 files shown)


CarlosRoca13 (approved):
Thanks!

pedrobaeza (approved):
/ocabot merge patch

PR #803, opened 2024-01-11
[17.0][PORT] 699 from 14.0: base_tier_validation multiple comment with approve_sequence_bypass

bosd:
Port #699 from 14.0 to 17.0.

when option approve_sequence_bypass is enabled and comments option with
sequence, the user add a comment and all level add this
comment.
This code add comment only for appropriate tier validation

PR #802, opened 2024-01-11
[16.0][PORT] 699 from 14.0: base_tier_validation multiple comment with approve_sequence_bypass

bosd:
Port #699 from 14.0 to 16.0.

PR #699, opened 2023-08-02
[14.0][FIX] base_tier_validation: multiple comment with approve_sequence_bypass

mathben:
[FIX] base_tier_validation: multiple comment with approve_sequence_bypass

    - when option approve_sequence_bypass is enable and comments option with
    sequence, the user add a comment and all level approbation add this
    comment.
    This code add comment only for appropriate tier validation

PR #1156, opened 2025-09-24
[FIX] base_tier_validation: comment support approve_sequence_bypass

mathben:
- the comment will only be added to last review and remove repetition

Improve https://github.com/OCA/server-ux/pull/975
Fix https://github.com/OCA/server-ux/pull/802
Fix https://github.com/OCA/server-ux/pull/803
Fix https://github.com/OCA/server-ux/pull/699

This code was ignoring approve_sequence_bypass


mathben:
Tests doesn't cover this feature, I will try to add more.
Only 1 test approve_sequence_bypass : test_12_approve_sequence_same_user_bypassed

Do you think it's necessary to add an option to decide if comment is repeated for each approbation or only on the last sequence?

mathben:
Add feature to show note when missing notifier, this is the reason of failing test.

PR #1158, opened 2025-09-25
[FIX] base_tier_validation: ensure skip_validation_check bypasses val…

Jord38:
…idations

context flag check was not present in `_tier_validation_check_state_on_write(vals)` and thus I was unable to properly skip the tier validation using the context flag.

now moved up to the write method to ensure validation is skipped if context flag is present


chrisb-c01:
@Jord38 could you possibly also add the use of the context flag to the documentation? Similar to https://github.com/OCA/queue/blob/18.0/queue_job/readme/USAGE.md?plain=1#L295

Jord38:
> @Jord38 could you possibly also add the use of the context flag to the documentation? Similar to https://github.com/OCA/queue/blob/18.0/queue_job/readme/USAGE.md?plain=1#L295

@chrisb-c01 
Added documentation on this context flag and also another existing context flag `skip_check_state_condition`

CasVissers-360ERP (approved):

chrisb-c01 (approved):

Jord38:
@gurneyalex @pedrobaeza can we get this merged?

PR #1161, opened 2025-10-01
[19.0][MIG] announcement: Migration to 19.0

bizzappdev:


gdgellatly (approved):
Tested on runboat - works as expected. Only question is why the need to change fields to store = True otherwise LGTM

announcement/models/announcement.py
```diff
@@ -74,10 +74,12 @@ class Announcement(models.Model):
     notification_start_date = fields.Datetime(
         compute="_compute_notification_start_date",
         help="Technical field to display announcements in the calendar view",
+        store=True,
```
pedrobaeza, line 77: Why modifying this in a migration?

pedrobaeza (changes requested):

PR #1171, opened 2025-10-15
[18.0][FIX] base_tier_validation: prevent error when trying to access review_ids in a one2many

celm1990:
When the validations are in a model that is inside another model(some custom module), after the reviews are approved, trying to open the one2many raises an error. This commit ensures that the value is taken from the correct record instead of env.model.root, which represents the main model and not the model that inherits tier.validation.

image

@pedrobaeza could you review this please?


pedrobaeza (approved):
Seems reasonable.

LoisRForgeFlow (approved):
/ocabot merge patch

Issue #1159, opened 2025-09-26
date_range: Set more one company for one date range


Andrii9090-tecnativa:
**Is your feature request related to a problem?**
Currently, in the date range model (date.range), you can only select one company, so you have to create many date ranges in an Odoo with multiple companies, and management becomes complicated.

**Additional context**
For example, I have to create three different date ranges, and I only need them in two of my companies.
Currently, I have to create six ranges of 3 periods, but it is inconvenient and will cause problems in the future when I need to update the data.

@lmignon can you help with this problem? 
We can't change `compnay_id (Many2one) -> `company_ids` (Many2many) in `date.range`model and `date.range.type` because there are many modules that depend on date_range.
Could there be a better solution?
Thanks

lmignon:
If you change the database structure you must provide a migration script and we must dump the version to major.

Andrii9090-tecnativa:
Hi @lmignon! Can you review this [PR](https://github.com/OCA/server-ux/pull/1172)?

Thanks ❤️

PR #1172, opened 2025-10-16
[18.0][IMP] date_range: Set multicompanies support for the date.range model

Andrii9090-tecnativa:
The date.range model currently supports only a single company (`company_id`). 
This requires creating multiple date ranges for different companies, which complicates management.
To resolve this, we replace the field `company_id` with `company_ids` (Many2many) in the model `date.range`

It's a short video with funcional test
https://www.loom.com/share/f1a0719eafa443058e3a0befc0eed9b4?sid=fbbd55c0-f75d-48aa-b9ec-d93e28af320f

[Issue](https://github.com/OCA/server-ux/issues/1159)
@moduon 
MT-10678


date_range/security/date_range_security.xml
```diff
@@ -11,7 +11,7 @@
         <field name="name">Date Range multi-company</field>
         <field name="model_id" ref="model_date_range" />
         <field name="domain_force">
-            ['|',('company_id', 'in', company_ids),('company_id','=',False)]
+            ['|',('company_ids', 'in', company_ids),('company_ids','=',False)]
```
cav-adhoc, line 14: Maybe user.company_ids.ids should be used here instead of company_ids? That way the rule will correctly filter records based on the user’s companies in multi-company setups.

date_range/models/date_range.py
cav-adhoc, line 16: Since company_ids is now a Many2many field, I think that the default method must return a list of IDs, not the recordset (self.env.company.ids)

Andrii9090-tecnativa: I think it is better to leave only one company as the default so as not to confuse the user.
With the improvement of this PR, the user can add more companies.

date_range/security/date_range_security.xml
Andrii9090-tecnativa: Thank you for your comment! :+1: 
company_ids already contains the companies selected by the user.

cav-adhoc (approved):
LGTM

Andrii9090-tecnativa:
Hi @lmignon can you review this PR?

Thanks :heart:

rafaelbn:
Hello @StefanRijnhart , @sbidoul @lmignon ✋🏼 

Please could you please take a look here just a quick review? 🙏🏼 😄 

Thank you!!!! ❤️ 
Rafael

EmilioPascual:
@Andrii9090 can you rebase it for review it functionally? Thank you.

Andrii9090-tecnativa:
@EmilioPascual runboat is available, you can functionally review it.

EmilioPascual (approved):
LGTM. Functional and code review.

Good work @Andrii9090

Andrii9090-tecnativa:
Hello @StefanRijnhart , @sbidoul @lmignon 

Please could you take a look this PR just a quick review?

Gelojr (approved):
Congratulations on the great work on this PR @Andrii9090 .
The following functional tests have been performed with the results indicated below.

- Test 1: Creation of a minimal Date Range Type – OK
- Test 2: Multi-company Date Range creation using company_ids – OK
- Test 3: Creation of a global Date Range without companies – OK
- Test 4: Overlap blocking when Date Ranges share at least one company and overlap is not allowed – OK
- Test 5: Overlap allowed when Date Ranges do not share any company – OK
- Test 6: Consistency check between Date Range Type company_id and Date Range company_ids – OK
- Test 7: Generate Date Ranges wizard creates Date Ranges with company_ids correctly set – OK
- Test 8: Multi-company record rule visibility based on company_ids – OK
- Test 9: Generate Date Ranges wizard with empty Company field generates global Date Ranges only (no duplication per company) – OK

date_range/migrations/18.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,18 @@
...
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.delete_sql_constraint_safely(
+        env, "date_range", "date_range", "date_range_uniq"
+    )
+    openupgrade.m2o_to_x2m(
+        env.cr, env["date.range"], "date_range", "company_ids", "company_id"
+    )
+    env.cr.execute(
+        "UPDATE ir_rule SET domain_force = %s WHERE name = %s",
```
yajo: issue: It's better to update the record by its XMLID, not by its name.

Another possibility is to delete that record by XMLID in the pre-migration phase, so Odoo will re-create it.

date_range/migrations/18.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,18 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.delete_sql_constraint_safely(
+        env, "date_range", "date_range", "date_range_uniq"
+    )
+    openupgrade.m2o_to_x2m(
+        env.cr, env["date.range"], "date_range", "company_ids", "company_id"
```
yajo, line 10: issue?: we're in post-migration, so probably the `company_id` field is already gone.

Please could you test the migration and make sure it works as expected?

date_range/models/date_range.py
```diff
@@ -73,6 +64,18 @@ def _validate_range(self):
...
+                this.type_id.company_id
+                and this.type_id.company_id not in this.company_ids
+            ):
+                raise UserError(
+                    self.env._(
+                        "The Date Range Type %(drt_name)s assigned to other "
+                        "company, you can't change the company in "
+                        "Date Range %(dt_name)s"
+                    )
+                    % {"drt_name": this.type_id.name, "dt_name": this.name}
```
yajo: suggestion(style): this is more beautiful 💄 


```suggestion
                    drt_name=this.type_id.name,
                    dt_name=this.name,
                )
```

date_range/models/date_range.py
```diff
@@ -81,20 +84,22 @@ def _validate_range(self):
...
                     date_range dt
+                INNER JOIN date_range_res_company_rel as rc
+                ON dt.id = rc.date_range_id
                 WHERE
                     DATERANGE(dt.date_start, dt.date_end, '[]') &&
                         DATERANGE(%s::date, %s::date, '[]')
                     AND dt.id != %s
                     AND dt.active
-                    AND dt.company_id = %s
+                    AND rc.res_company_id IN %s
```
yajo: suggestion: you can use an array to avoid type conversions and failures in case of empty tuples:


```suggestion
                    AND rc.res_company_id = ANY(%s)
```

date_range/models/date_range.py
```diff
@@ -81,20 +84,22 @@ def _validate_range(self):
...
                     AND dt.type_id=%s;"""
             self.env.cr.execute(
                 SQL,
                 (
                     this.date_start,
                     this.date_end,
                     this.id,
-                    this.company_id.id or None,
+                    tuple(this.company_ids.ids) or (None,),
```
yajo: suggestion: following from the previous one:


```suggestion
                    this.company_ids.ids,
```

date_range/models/date_range.py
yajo: thought: did you have into account that, with the new change in that SQL, it can return now more than 1 row? Is it correct that you just ignore the rest of rows? Shouldn't the query be changed to make it still return a single row as it did before?

date_range/models/date_range.py
```diff
@@ -73,6 +64,18 @@ def _validate_range(self):
                 )
             if this.type_id.allow_overlap:
                 continue
+            if bool(
+                this.type_id.company_id
+                and this.type_id.company_id not in this.company_ids
+            ):
```
yajo, line 70: issue: if the type has `allow_overlap=True`, then the company consistency will not be checked because of the `continue` sentence.

issue(style): `bool()` is implicit in an `if`. You can remove that.

yajo (changes requested):
Code review.

date_range/migrations/18.0.2.0.0/post-migration.py
Andrii9090-tecnativa: If we do not do this, we will lose all the data in the company_id column. I tested without this code.

Andrii9090-tecnativa: It’s a short video to test the post-migration script.
https://www.loom.com/share/943de3bc9ea4463da23fb6f8d2e234a6

yajo: OK, I was thinking that at this point the `company_id` column would be already removed. I see by your video that it's working as expected. Thanks!

date_range/migrations/18.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,24 @@
...
+    )
+    res_id = env.cr.fetchone()
+    if res_id:
+        env.cr.execute(
+            "UPDATE ir_rule SET domain_force = %s WHERE id = %s",
+            (
+                "['|',('company_ids', 'in', company_ids),('company_ids','=',False)]",
+                res_id[0],
+            ),
+        )
```
yajo: suggestion: simpler code

```suggestion
    rule = env.ref("date_range.date_range_comp_rule")
    if rule:
        rule.domain_force = "['|',('company_ids', 'in', company_ids),('company_ids','=',False)]"
```

date_range/models/date_range.py
```diff
@@ -81,29 +85,34 @@ def _validate_range(self):
...
             if res:
-                dt = self.browse(res[0][0])
+                date_ranges = self.browse(res[0])
                 raise ValidationError(
-                    self.env._("%(thisname)s overlaps %(dtname)s")
-                    % {"thisname": this.name, "dtname": dt.name}
+                    self.env._(
+                        "%(thisname)s overlaps %(dtnames)s",
+                        thisname=this.name,
+                        dtnames=date_ranges.mapped("name"),
```
yajo:
```suggestion
                        dtnames=", ".join(date_ranges.mapped("name")),
```

date_range/tests/test_date_range.py
```diff
@@ -95,7 +96,7 @@ def test_overlap(self):
                 }
             )
         message = str(cm.exception.args[0])
-        self.assertEqual(message, "FS2016 overlaps FS2015")
+        self.assertEqual(message, "FS2016 overlaps ['FS2015']")
```
yajo:
```suggestion
        self.assertEqual(message, "FS2016 overlaps FS2015")
```

yajo (changes requested):
Some comments to improve code. Mostly OK. Thanks!

yajo (approved):
/ocabot merge major

carlos-lopez-tecnativa (commented):
These changes are causing an error in the module: https://github.com/OCA/account-financial-reporting/blob/dcb2a3ed073c767f6ebc2181e5921825bb230a55/account_financial_report/wizard/general_ledger_wizard.py#L224
image

christian-ramos-tecnativa:
Fixed here https://github.com/OCA/account-financial-reporting/pull/1437

sbidoul:
This is also breaking mis_builder. Is there no way to achieve the use case while keeping some backward compatibility?

sbidoul:
@rafaelbn @yajo @Andrii9090 what do you think? Any chance to introduce some backward compatibility. In https://github.com/OCA/mis-builder/pull/758#issuecomment-3776814327, @pedrobaeza is suggesting a searchable `company_id` field.

yajo:
Apologies! I didn't realize the impact when I reviewed it.

We can use https://github.com/OCA/multi-company/blob/18.0/base_multi_company/models/multi_company_abstract.py to get the backwards compatibility.

pedrobaeza:
We can't add a new dependency in the modules...

sbidoul:
> We can't add a new dependency in the modules...

Why not?

yajo:
Yes, we do that all the time...

yajo:
See https://github.com/OCA/server-ux/pull/1227.

andreampiovesana (commented):
this break all report that use date_range and company_id instead company_ids
you break many apps!
for italy:
- l10n_it_financial_statements_report

juancarlosonate-tecnativa (commented):
This change is severely impacting modules. Because of this, I strongly believe the change should be reverted and properly analyzed in more depth.

alexis-via:
@Andrii9090 If you want to share date.range across companies, set company_id to False (it is not a required field, cf https://github.com/OCA/server-ux/blob/17.0/date_range/models/date_range.py#L30 , and it is on purpose !) and you will be able to use the date.range from all the companies. I often do that in my multi-company deployments and it works well.

PR #1174, opened 2025-10-27
[19.0][MIG] server_action_mass_edit: Migration to 19.0

bizzappdev:


leemannd:
There is also another PR for migration here -> https://github.com/OCA/server-ux/pull/1182

PR #1178, opened 2025-11-03
[DRAFT] Revert "[FIX] base_tier_validation: Proper notifications on reviews"

jcadhoc:
This reverts commit 19e0337c0b11ac91cf6832fc2aef8f74964181f9.


pedrobaeza:
Why?

jcadhoc:
Hello @pedrobaeza 

Whe you have more than one definition created with the same responsable, it doesn´t send email in the last definition approve, we re testing it so far, but we think it's related to this change

pedrobaeza:
But the solution is not to revert something, but fix it. Please expose the complete steps to reproduce the problem, and we can help hunting the bug, or better, add a regression test with such steps that fail right now.

PR #1179, opened 2025-11-05
[18.0][FIX] base_substate: fix window action view violation during migration

StefanRijnhart:
Fixes
```
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "act_window_view_unique_mode_per_action"
DETAIL:  Key (act_window_id, view_mode)=(755, list) already exists.

odoo.tools.convert.ParseError: while parsing /home/odoo/odoo18/src/server-ux/base_substate/views/base_substate_type_views.xml:67, somewhere inside
<record id="act_open_base_substate_type_view_list" model="ir.actions.act_window.view">
        <field name="act_window_id" ref="act_open_base_substate_type_view"/>
        <field name="sequence" eval="10"/>
        <field name="view_mode">list</field>
        <field name="view_id" ref="base_substate_type_view_list"/>
    </record>
```
after updating the module on a database that was migrated using Odoo's enterprise upgrade service.

### DIFF
--- a/base_substate/migrations/18.0.1.0.0/pre-migration.py
+++ b/base_substate/migrations/18.0.1.0.0/pre-migration.py
@@ -1,0 +1,25 @@
+# Copyright 2025 Opener B.V. (<http://opener.amsterdam>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+def migrate(cr, version):
+    """Prevent violation of act_window_view_unique_mode_per_action.
+
+    Violation was triggered after rename of ir.actions.act_window.view in XML.
+    """
+    cr.execute(
+        """
+        update ir_model_data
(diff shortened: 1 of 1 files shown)


Saran440 (approved):
LGTM 👍

bosd (approved):

etobella (approved):
It would have been cleaner to user https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L1279

However, the option is valid, so I think we can proceed with this.

StefanRijnhart:
@etobella Well, as the person that wrote that openupgrade method, I'd say it's a bit of overkill to require the openupgradelib dependency for this. Thanks for the merge!

etobella:
Well, all of my instances have openupgradelib installed... Personally I would recommend to everyone using OCA module to  have it as a lot of migrations include this dependency

PR #1180, opened 2025-11-05
[17.0][IMP] base_optional_quick_create: allow_quick_create context flag

vvrossem:
Introduce an `allow_quick_create` context flag to bypass the restriction applied by models with `avoid_quick_create` enabled. This allows controlled quick creation of records in specific, legitimate cases.

For example, the method `mail.thread::_mail_find_partner_from_emails(emails, records=None, force_create=False, extra_domain=False)`, with `force_create` set to `True`, creates a new partner (if not found) by using `name_create`:
```
partner = self.env['res.partner'].browse(self.env['res.partner'].name_create(contact)[0])
```
cf. https://github.com/odoo/odoo/blob/35ed88e2e6f66147c4c1582761df1e53011b3643/addons/mail/models/mail_thread.py#L1971-L2039

When `avoid_quick_create` is enabled on the `res.partner` model, an unwanted `UserError` is raised.

With the new `allow_quick_create` context flag, such internal operations can safely bypass the restriction while keeping it enforced for regular user actions. 
This requires to extend core methods in order to inject that context flag.

### DIFF
--- a/base_optional_quick_create/models/ir_model.py
+++ b/base_optional_quick_create/models/ir_model.py
@@ -63,6 +63,8 @@ class IrModel(models.Model):
         def _wrap_name_create():
             @api.model
             def wrapper(self, name):
+                if self.env.context.get("allow_quick_create"):
+                    return wrapper.origin(self, name)
                 raise UserError(
                     _(
                         "Can't create %(model)s with name %(name)s quickly.\n"
--- a/base_optional_quick_create/tests/test_quick_create.py
+++ b/base_optional_quick_create/tests/test_quick_create.py
@@ -4,16 +4,23 @@
 from odoo.exceptions import UserError
 from odoo.tests.common import TransactionCase
 
+from odoo.addons.base.tests.common import DISABLED_MAIL_CONTEXT
+
 
 class TestQuickCreate(TransactionCase):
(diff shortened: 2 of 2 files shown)


santostelmo (approved):

SilvioC2C (approved):

base_optional_quick_create/models/ir_model.py
```diff
@@ -63,6 +63,8 @@ def _patch_quick_create(self):
         def _wrap_name_create():
             @api.model
             def wrapper(self, name):
+                if self.env.context.get("allow_quick_create"):
+                    return wrapper.origin(self, name)
```
sebalix, line 67: It doesn't hurt but not sure it's the right approach neither.

If this module is breaking std behavior when installed, the fix should not be proposing a workaround, forcing the creation of new module like `mail_optional_quick_create` to fix the std feature of `mail`.

If the goal is to prevent users to create duplicated records by disabling "Quick create" feature, then it should belong to `web` stack (by using [`web_m2x_options`](https://github.com/OCA/web/tree/18.0/web_m2x_options) for instance), so all internal calls to `name_create` will still work as expected.

I would deprecate this module instead 🤷‍♂️

PR #1183, opened 2025-11-06
[IMP][CL] Blacklist exeptions

CILC98:
An improvement has been made to define an exception by simulating a blacklist, so that if more modules are installed or created, it will not be necessary to add new fields to the exceptions. This is because there are cases where most fields can be edited, but not the category or critical fields.

ATT: César León from Trescloud


CILC98:
@LoisRForgeFlow Hi! I've introduced an improvement to the module to treat exceptions as a blacklist, so that even if I install more modules, the critical fields remain read-only. What do you thing?

PR #1186, opened 2025-11-10
[18.0][FIX] base_tier_validation_forward: fix multiple reviews in 'pending' after forward

StefanRijnhart:
# How to reproduce:
    
* Create two reviews for the same model with different sequences and `Approve by
    sequence` enabled.
* Request validation on a record. Let the user of the first review forward the
    first validation.
   
# Desired result:

The new (forward) review is in state 'pending', and the remaining review is in state 'waiting'.
    
# Actual result:
    
Both the new (forward) review and the remaining review are in state 'pending'.
    
# Suggested solution:
    
`_compute_can_review` has to be deferred during the fowarding process. It was called explicitely afterwards already.

### DIFF
--- a/base_tier_validation/models/tier_review.py
+++ b/base_tier_validation/models/tier_review.py
@@ -94,6 +94,14 @@ class TierReview(models.Model):
 
     @api.depends("definition_id.approve_sequence")
     def _compute_can_review(self):
+        """Update review status. Called manually from various review operations.
+
+        To defer recompute, use context key
+        `tier_validation_defer_compute_can_review`. Be sure to explicitely call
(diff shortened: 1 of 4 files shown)


espo-tony (approved):
👍

len-foss (approved):
It's a bit hacky, but this is because the module's design is pushing the limits of the ORM. So AFAICT it's the best solution we can get given the constraints we have 👍

LoisRForgeFlow (approved):
LGTM, indeed is a bit hacky but the issue is real. Thanks for providing appropriate test cases.

PR #1199, opened 2025-12-03
[18.0][IMP] base_tier_validation: Adjustment to improve inheritance

CILC98:
Improvement to make modifications to the exception search and the fields to enable during and after validation

PR #1206, opened 2025-12-15
[18.0][ADD] Fields like blacklist in tier validation

CILC98:
This module allows you to define the fields that cannot be edited during a validation.

Depends on #1199


CILC98:
@LoisRForgeFlow Hi! What do you think of this proposal? I think it's useful when you have a project with many third-party modules.

PR #1212, opened 2026-01-08
[17.0] [FIX] Translated stored related field warning on `tier.review.name`.

denaux:
Proposal to fix `WARNING odoo.fields: Translated stored related field (tier.review.name) will not be computed correctly in all languages`.

The above warning is raised on every related field that is both stored and translatable. As the field inheritance in question makes a related field computed, this can be fixed by explicitly making the field _not_ related anymore.

Other proposals are welcome!


bosd (approved):

CristianoMafraJunior (approved):
LGTM

PR #1202, opened 2025-12-09
[16.0][IMP] template_content_swapper: add feature to use domain for applying some specific records in report pdfs

AungKoKoLin1997:
This PR adds an improvement to restrict the applying changes only for some specific records by adding domain in configuration.
With this changes, we can only adjust the content of report pdf depends on some condition, not globally.

@qrtl QT5806

PR #1214, opened 2026-01-09
[18.0][IMP] template_content_swapper: add feature to use domain for applying some specific records in report pdfs

AungKoKoLin1997:
Forward port of https://github.com/OCA/server-ux/pull/1202

This PR adds an improvement to restrict the applying changes only for some specific records by adding domain in configuration.
With this changes, we can only adjust the content of report pdf depends on some condition, not globally.

@qrtl QT5088


template_content_swapper/views/template_content_mapping_views.xml
```diff
@@ -8,6 +8,13 @@
                 <field name="name" />
                 <field name="report_id" />
                 <field name="template_id" readonly="report_id != False" />
+                <field name="report_model" invisible="1" />
```
yostashiro:
```suggestion
                <field name="report_model" column_invisible="1" />
```

yostashiro (changes requested):
Could we add tests, or create a test module?

template_content_swapper/models/ir_qweb.py
```diff
@@ -1,17 +1,69 @@
...
+            dom = safe_eval(domain_str)
+        except Exception:
+            _logger.warning(
+                "Invalid domain on template.content.mapping for %s,%s: %s",
+                model_name,
+                res_id,
+                domain_str,
+            )
+            return False
+        return bool(self.env[model_name].search_count([("id", "=", res_id)] + dom))
```
nobuQuartile:
```suggestion
    def _get_eval_context(self):
            return {
                "datetime": datetime,
                "date": date,
                "timedelta": timedelta,
                "context_today": datetime.now,
            }

    def _record_matches_domain(self, model_name, res_id, domain_str):
        """Check if record (model_name, res_id) matches the given domain."""
        try:
            dom = safe_eval(domain_str, self._get_eval_context())
        except Exception:
            _logger.warning(
                "Invalid domain on template.content.mapping for %s,%s: %s",
                model_name,
                res_id,
                domain_str,
            )
            return False
        return bool(self.env[model_name].search_count([("id", "=", res_id)] + dom))
```

We should add eval context and allow these date functions in the domain.

AungKoKoLin1997: Yes. That is good point.
Thanks!

template_content_swapper/models/ir_qweb.py
```diff
@@ -1,17 +1,80 @@
...
+            dom = safe_eval(domain_str, self._get_eval_context())
+        except Exception:
+            _logger.warning(
+                "Invalid domain on template.content.mapping for %s,%s: %s",
+                model_name,
+                res_id,
+                domain_str,
+            )
+            return False
+        return bool(self.env[model_name].search_count([("id", "=", res_id)] + dom))
```
yostashiro: Please use expression.AND for domain concatenation.

template_content_swapper/models/ir_qweb.py
```diff
@@ -1,17 +1,82 @@
...
+            html_str = html_str.replace(m.content_from, m.content_to or "")
+        return html_str
+
+    def _get_eval_context(self):
+        return {
+            "time": tools.safe_eval.time,
+            "datetime": tools.safe_eval.datetime,
+            "dateutil": tools.safe_eval.dateutil,
+            "timezone": timezone,
+            "context_today": lambda: datetime.now(),
```
yostashiro:
```suggestion
            "timezone": tools.safe_eval.pytz.timezone,
            "context_today": lambda: fields.Date.context_today(self),
```

template_content_swapper/models/ir_qweb.py
```diff
@@ -1,17 +1,82 @@
...
+        }
```
yostashiro: Move this to template.content.mapping and add a constraint to check the domain format there (just like we are doing for report_qweb_field_option now).

yostashiro (changes requested):

template_content_swapper/models/ir_qweb.py
AungKoKoLin1997: Done!

yostashiro (approved):
Code review and functional test. LGTM. 👍 

Image

Would be good to add a form view to the template mapping as the list view is getting more cluttered.

AungKoKoLin1997:
> Would be good to add a form view to the template mapping as the list view is getting more cluttered.

@yostashiro Yes, I will follow up with separate PR since it is not related with current changes and to be better for discussing separately.

PR #1217, opened 2026-01-15
[18.0][IMP] base_tier_validation_forward: preserve add_comment return

ArnauCForgeFlow:
The override of **add_comment()** changed the method’s return type, breaking standard wizard chaining and causing incompatibilities with other modules.


LoisRForgeFlow (approved):
Makes sense :+1:

PR #1218, opened 2026-01-15
[19.0][MIG] multi_step_wizard: Migration to 19.0

fvkugit:


sergijPf (approved):
LGTM

alexey-pelykh (approved):
Thanks for the migration! Looks good.

The only meaningful change here is the test infrastructure update in `tests/common.py` — replacing `odoo_test_helper.FakeModelLoader` with the native `add_to_registry` + `setUpClass`/`addClassCleanup` pattern, which is the right approach for 19.0. This also fixes the `tearDown`/`tearDownClass` mismatch that existed in 18.0.

Model, views, and manifest are all correct. CI is green.

multi_step_wizard/tests/common.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2019 Brainbean Apps (https://brainbeanapps.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo.orm.model_classes import add_to_registry
+from odoo.tests import common
+
+
+class CommonTestMultiStepWizard(common.TransactionCase):
```
BhaveshHeliconia, line 8:
```suggestion
class CommonTestMultiStepWizard(Basecommon):
```

multi_step_wizard/tests/common.py
```diff
@@ -0,0 +1,20 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
BhaveshHeliconia, line 12:
```suggestion
```

BhaveshHeliconia (changes requested):
Could we use BaseCommon instead of TransactionCase? In that case, there would be no need to redefine cls.env with tracking_disable=True.

PR #1227, opened 2026-01-21
[FIX] date_range: backwards compatibility with company_id field

yajo:
Fixes backward incompatibilities introduced by https://github.com/OCA/server-ux/pull/1172.

@moduon MT-10678


yajo:
[Tests were failing](https://github.com/OCA/server-ux/actions/runs/21204264523/job/60996837900?pr=1227) with:

```
Run manifestoo -d . check-dev-status --default-dev-status=Beta
date_range (Mature) depends on base_multi_company (Production/Stable)
Error: Process completed with exit code 1.
```

Therefore, I changed the status, so we can continue.

yajo:
With this simple change, we recover the `company_id` field:
image

date_range/models/date_range.py
lmignon, line 31:
```suggestion
```

The `comodel_name` and  `string` properties are already defined on the base model and should not be redefined here 😏

lmignon (approved):
A little comment ....
 This fix should be tested with mis_builder or any other addon that was broken by the previous modification to verify that compatibility has been restored and everything works properly now...

date_range/models/date_range.py
Andrii9090-tecnativa: I have deleted `comodel_name` and `string`

EmilioPascual (approved):

Andrii9090-tecnativa (approved):

pedrobaeza (changes requested):
No new dependencies should be introduced into a base stable module like this one.

victoralmau (changes requested):
IMO, no new dependencies should be added to stable modules. If you want to add compatibility that was previously removed, it should be added to the code of this module.

In any case, I do not understand the need for this change. If any module and/or test needs to be adapted, it is done (as has already been done in github.com/OCA/account-financial-reporting/pull/1437) and is pending in https://github.com/OCA/mis-builder/pull/758. Or perhaps you do not want to merge it for some reason that I am unaware of?

etobella:
we shouldn't add extra dependencies in an already mature migrated module. I prefer to revert the change than doing this

pilarvargas-tecnativa (changes requested):
I also believe that no new dependencies should be added to stable modules.

lmignon:
Hello everyone, 

I really don't want to stir up debate or pass judgment, but I would like to understand the rational reasoning behind why adding a dependency would be a problem.

Kind regards

pedrobaeza:
It forces to install it in all the existing instances (which are all, as this module is used in very base modules), and on that very moment, while you don't migrate that dependency module, you can't use this one, delaying the availability of it, and doing the migration of the other module possibly faulty if you only focus on making what this module uses.

andreampiovesana:
I prefer to ERASE last commit in date_range

carlosdauden:
A satisfactory solution for everyone could be to consider the functionality of sharing the date.range between companies by simply emptying the company_id field, as is done, for example, in res_partner.
Another solution (which could be additional to the previous one) would be a wizard to quickly duplicate the ranges between companies, for exceptional cases where sharing between all companies is not the best option.

Steps to follow:
- Reverse the original change and the PRs that have been created to resolve the problem caused.
- Create a new PR with one of these less conflicting solutions.
- During the migration to 20.0, analyse the possibility of adding the dependency or continuing with this proposed solution.

Related PRs:
https://github.com/OCA/server-ux/pull/1172
https://github.com/OCA/account-financial-reporting/pull/1437
https://github.com/OCA/mis-builder/pull/758

andreampiovesana:
> A satisfactory solution for everyone could be to consider the functionality of sharing the date.range between companies by simply emptying the company_id field, as is done, for example, in res_partner.

This is good

etobella:
> Hello everyone,
> 
> I really don't want to stir up debate or pass judgment, but I would like to understand the rational reasoning behind why adding a dependency would be a problem.
> 
> Kind regards

Sure @lmignon.

there are several reasons,

1. People has different ways to create their installations. in my case, I specify on my customers recipes the specific modules that are used. Adding a new dependancy breaks their installation. I have a CI/CD verifications but that implies changing my customers instance. I know that other people use pip, but as a community, we cannot enforce the way for doing this.
2. I can understand this kind of changes in Beta modules, but that shouldn't happen in Stable/Mature. As the name suggest, they are stable, no major changes should occur (new dependancies are major changes IMO)

lmignon:
Thank you for all your feedbacks regarding the addition of a new dependency. IMO, to keep this improvement and avoid maintenance cost into existing installations, the best approach could be to replicate the approach of the mixin provided by `multi.company.abstract`mixin into the `date.range`model and remove the dependency. At the same time, a comment should be added around this specific code to highlight that this code could be removed when migrating the module to odoo 20 in favour of relaying on the `base_multi_company` addon as dependency.

Regards

etobella:
Personally, I would opt for the reversion of the commits (quick actuation) and then they can propose the changes in a new PR.

yajo:
It is late for rollback, because some downstream modules have already been fixed. Also, a rollback without migration scripts would cause data loss. We have to fix it.

I don't share the vision of not adding the new dependency, mostly when it is such a small one. But I see it is important to others, so we can adapt the fix.

This is the most reasonable solution given the situation: https://github.com/OCA/server-ux/pull/1227#issuecomment-3782988133. Hold on please while I change the PR.

yajo:
Ready to review.

lmignon (commented):
Thank you @yajo for this latest change. It should restore compatibility with the previous implementation without impacting current deployments too much. At a glance, the code looks good to me, but I’ll let the people using this addon on Odoo 18 review and approve this change. I hope it meets the expectations of all users.

date_range/readme/ROADMAP.md
```diff
@@ -0,0 +1,3 @@
+- Inherit from `base_multi_company` and remove `models/date_range_multicompany.py`
```
pedrobaeza: No, this shouldn't ever depend on that other module.

pedrobaeza (changes requested):

date_range/readme/ROADMAP.md
yajo: Reworded.

pedrobaeza: Sorry, but you are trying to force something to keep a marginal feature. The solution is to revert it and put it in a different module, like stated in the other PR. This module shouldn't be complicated more, being a very base one.

yajo: I'm fixing a bug. Regarding everything else, I quote myself from https://github.com/OCA/server-ux/pull/1229#issuecomment-3783880428:

> I'm not closed to the idea

etobella (changes requested):
I think we have stated the solution, revert and add a migration script to revert it.

yajo:
Actually, I've been thinking that we don't really need the dependency on `base_multi_company` on v20, because it's present only for backwards compatibility, and we don't need backwards compatibility when migrating to a new version. So, we can merge this PR and we have:

- The benefits of https://github.com/OCA/server-ux/pull/1172, which make management much easier for instances with lots of companies.
- The extra benefit of backwards compatibility.
- A path of migration for future versions that does not require extra dependencies.

It ticks all the boxes, so IMHO this is, technically speaking, the most correct solution.

If we merge this, I'll migrate the changes to v19 so OpenUpgrade is not negatively impacted and the UX is the same.

etobella:
@yajo I think that the solution proposed in #1230 has more consensus. I prefer to continue on that way and leave the space to a new glue module if you want to add this logic.

etobella:
Actually, when doing the change on the scripts as you suggested I found a quite interesting bug with your approach. With the current state, we have a problem because we can set a type  for company A, and set a daterange for company A and B. That is obviously not desired.

LoisRForgeFlow (commented):
Hi all,

FWIW, I will share my opinion about this topic.

For me a new glue module date_range_multi_company is the way to go. I think it is more consistent in behavior to keep date_range as it was before, because a user that does not use base_multi_company will be used to the standard Odoo multi company behavior: no company = shared for all, company set = only that company has access. I think it is best to keep that behavior also in date_range base module, most the of the users will expect a standard multi-company behavior.

Then if someone wants the base_multi_company feature in date_range, date_range_multi_company could be installed to add that behavior.

Kind regards,

yajo:
Thanks. Closing in favor of https://github.com/OCA/server-ux/pull/1230.

PR #1228, opened 2026-01-21
[18.0][IMP] base_tier_validation: handle multiple validations in different states

ArnauCForgeFlow:
This Pull Request enables the assignment of multiple tier definitions across different states. This allows us to validate reviews for a specific state without removing previous reviews. Also, when resetting validation, only the reviews corresponding to that specific state will be deleted.

Note: To achieve full functionality, the following two methods must be extended:

`_allow_to_remove_reviews()`
We override this method to prevent the system from automatically clear the existing reviews when performing the transition corresponding to the second validation step (specific state_from → state_to).

`_check_allow_write_after_validation()`
Standard behavior prevents writing to a record once it is validated. We override this to explicitly whitelist the transition defined for the second validation step, returning True. This ensures the specific state change is not blocked by the validation lock mechanisms.

Please note that this implementation is currently a workaround. While it works correctly now, a small refactor to base_tier_validation will be needed in the future to simplify the process and make it cleaner.

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -40,6 +40,10 @@ class TierValidation(models.AbstractModel):
         domain=lambda self: [("model", "=", self._name)],
         auto_join=True,
     )
+    current_review_ids = fields.Many2many(
+        comodel_name="tier.review",
+        compute="_compute_current_review_ids",
+    )
     # TODO: Delete in v19 in favor of validation_status field
     validated = fields.Boolean(
         compute="_compute_validated_rejected", search="_search_validated"
@@ -82,7 +86,7 @@ class TierValidation(models.AbstractModel):
 
     def _compute_has_comment(self):
         for rec in self:
-            has_comment = rec.review_ids.filtered(
+            has_comment = rec.current_review_ids.filtered(
                 lambda r: r.status in ("waiting", "pending")
                 and self.env.user in r.reviewer_ids
             ).mapped("has_comment")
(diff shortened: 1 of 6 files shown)


LauraCForgeFlow (approved):
Functional test LGTM

LoisRForgeFlow (changes requested):
LG as a first step for multi transition handling!

Could you please add tests for this? you could probably add a new tester model for this specific new case.

StefanRijnhart (commented):
Can you please add a test to illustrate the purpose, and to show that it works well with this refactoring?

ArnauCForgeFlow:
@StefanRijnhart @LoisRForgeFlow added tests! the ci is failing due to this issue not related with the module: https://github.com/odoo/odoo/issues/245835

LoisRForgeFlow:
@ArnauCForgeFlow can you rebase? the issue with the translation has been fixed upstream.

ArnauCForgeFlow:
@LoisRForgeFlow done!

bosd (commented):
Nice refactor — scoping reviews per state with `current_review_ids` reads cleanly, and `test_35` covers the flow well. A few things I'd want sorted before this goes in:

**1. `_compute_current_review_ids` is missing dependencies.** It only depends on the state field, but the result also depends on `review_ids` and their status/definition. So when a review is created, approved or deleted without a state change, the cached value goes stale — and this field now drives the templates and all the write-lock checks. I noticed `test_35` calls `invalidate_recordset()` after almost every step, which is basically hiding this. Could you extend the depends?

```python
@api.depends(lambda self: [self._state_field],
             "review_ids", "review_ids.status", "review_ids.definition_id")
```

**2. `request_validation` quietly changed the definition ordering.** It used to search with `order="sequence desc"`; now it goes through `_get_valid_definitions()`, which has no `order` and falls back to `id`. Since the review `sequence` is assigned in iteration order, this flips the approval order for existing single-state setups using `approve_sequence` — not just the new multi-state case. Was that intended? If not, adding `order="sequence desc"` back would keep behavior stable.

**3. Mixed `review_ids` / `current_review_ids` in `_compute_validation_status`.** `any_rejected` and `any_pending` use the current-state subset, but `any_waiting` still scans the full `review_ids`. Is that deliberate (keeping "waiting" sticky across states), or a missed substitution? A one-line comment either way would help.

PR #1229, opened 2026-01-21
[18.0][REV] "[IMP] date_range: Set multicompanies support for the date.range model"

juancarlosonate-tecnativa:
Due to the multiple conflicts this change is causing https://github.com/OCA/server-ux/pull/1172 I propose reverting it, or at least keeping this approach as a fallback to unblock the situation.

@tecnativa


andreampiovesana (approved):
much better

pedrobaeza (approved):
Yes, better to undo it for now.

yajo (changes requested):
If we merge this, we will break again the modules that already adapted to the change. Example: https://github.com/OCA/account-financial-reporting/pull/1437

It is late for reverting, we need to fix the issue.

date_range/migrations/18.0.2.0.0/post-migration.py
```diff
@@ -1,16 +0,0 @@
-from openupgradelib import openupgrade
-
-
-@openupgrade.migrate()
-def migrate(env, version):
```
Andrii9090-tecnativa, line 5: I think a migration script is needed to return the DB to its previous state.

Andrii9090-tecnativa (changes requested):

date_range/migrations/18.0.2.0.0/post-migration.py
etobella: Yes, maybe we need that, but this script should be deleted.

etobella (approved):

etobella:
> If we merge this, we will break again the modules that already adapted to the change. Example: [OCA/account-financial-reporting#1437](https://github.com/OCA/account-financial-reporting/pull/1437)
> 
> It is late for reverting, we need to fix the issue.

I prefer to revert the changed modules too before applying an extra change where we are not sure what will happen. Sorry.

yajo:
Merging this PR would cause more damage than the one you're trying to avoid, as explained in https://github.com/OCA/server-ux/pull/1227#issuecomment-3783232542.

The fix there is now ready to merge, with both forwards and backwards compatibility included.

IMHO this PR can be closed, to merge the actual fix in #1227.

lmignon (changes requested):
Why don't you give a try to #1227?

etobella:
Because it is adding an extra dependency in the future when we could do it in a separate module (we have other cases in the community).

lmignon:
> Because it is adding an extra dependency in the future when we could do it in a separate module (we have other cases in the community).

What's the proposal? Make a date_range_multi_company addon ?

etobella:
IMO that would be the cleanest. I don't know other people opinion on this topic

lmignon:
> IMO that would be the cleanest. I don't know other people opinion on this topic

@yajo

yajo:
I'm not closed to the idea, but that's something regarding the _future_ of the module, so it is meant to be discussed after merging the fix at https://github.com/OCA/server-ux/pull/1227, which is still the quickest solution for the problem we have at the _present_.

pedrobaeza:
I also think this granular control on date ranges is not a general need. OK to allow to share the same date range across all the companies leaving the field empty, but selecting specific companies should be something in an extra module. 

About the effects of reverting this change, we just need to revert the account_financial_report one. Nothing more needs to be done, while compatibilizing this breaking change requires more effort.

etobella:
I disagree with your conclusion. If we merge your module, we are accepting the change, and the split will not happen. I would prefer to revert, create the scripts if necessary and then you can create the module if you want to

pedrobaeza:
> I'm not closed to the idea, but that's something regarding the _future_ of the module, so it is meant to be discussed after merging the fix at #1227, which is still the quickest solution for the problem we have at the _present_.

The fix shouldn't be to add an extra dependency as said, and more if we are talking to split it. For your customers, just keep the previous branch state while the other actions are done, and let's revert this to the original state.

yajo:
I'm not adding an extra dependency at #1227 anymore.

To perform the reversion appropriately, you need to add a migration script for wherever https://github.com/OCA/server-ux/pull/1172 has been already deployed.

etobella:
> I'm not adding an extra dependency at #1227 anymore.
> 
> To perform the reversion appropriately, you need to add a migration script for wherever #1172 has been already deployed.

Yes, you are not, but it will be done in the future in something that is not really required. there were several solutions proposed. right now, we should think about removing the feature and restore the  data as much as possible

yajo:
Still, I see no problem in merging https://github.com/OCA/server-ux/pull/1227 and doing all of what you say afterwards.

pedrobaeza:
Because as Enric has pointed, once the storm "has passed", nothing will be done, as everyone has a lot of things to do. And that's not all, you are fixing 18, but not thinking on the migration path passing through 19, and you will probably go to 20, leaving that unfinished. Do it in an extra module, and you will only need to migrate that extra module to the versions you want.

@juancarlosonate-tecnativa @victoralmau can you include the migration scripts to convert the m2m back to the m2o?

yajo:
So, IIUC, the reason you prefer to not merge the fix is because that will force everyone else to do what you want right now.

Am I correct?

pedrobaeza:
No, because there are a lot of more things to do to fix the problem that didn't exist before merging this into a stable module, and it's also not on the latest version of Odoo. Other times that this has past, the revert has been done without such opposition. I have reverted or directly not merge things that I have developed in similar cases.

yajo:
I have been also able to fix bugs without so much opposition in the past. We all develop stuff in not-the-last version of Odoo and that's never been a problem. And anyways, still nothing holds you back from doing whatever you need after merging the hotfix.

etobella:
@yajo I understand your frustration, but you should understand that this giving several problems. There is a solution with less stress that allows you to have your result. For me it is a good ending for everyone (you get your functionality, the module will be stable)

The only thing that is missing here is the scripts. When they are done, we can ensure that nothing will be broken

andreampiovesana:
everybody need  a fast solution

pedrobaeza:
Take a look on a recent similar case in partner-contact where the solution was to revert it back: https://github.com/OCA/partner-contact/pull/2263

yajo:
Closing in favor of https://github.com/OCA/server-ux/pull/1230.

PR #1230, opened 2026-01-22
[18.0][REV] date_range: Set multicompany support

etobella:
Removed the dependancy and set the migration scripts already tested

@pedrobaeza @juancarlosonate-tecnativa @yajo @andreampiovesana 

#1229 #1172


pedrobaeza (approved):

andreampiovesana (approved):
LGTM

date_range/migrations/18.0.4.0.0/post-migration.py
```diff
@@ -0,0 +1,20 @@
...
+@openupgrade.migrate()
+def migrate(env, version):
+    if openupgrade.table_exists(env.cr, "date_range_res_company_rel"):
+        openupgrade.logged_query(
+            env.cr,
+            """
+            UPDATE date_range dr
+            SET company_id = drrcr.res_company_id
+            FROM date_range_res_company_rel drrcr
+            WHERE dr.id = drrcr.date_range_id
```
yajo: issue: If one rule had multiple companies, you'll be losing all but one.

yajo (changes requested):

date_range/migrations/18.0.4.0.0/post-migration.py
etobella: I know @yajo but as the change has been there for 2 days, I assume that not many people used it. For this reason I prefered to simplify it. Other option is, if it has multiple ones, I can set the company_id as false. Whatever you prefer, but I don't recommend to duplicate everything

yajo: If there are not many people that used it, then there will be not a lot of duplicates anyways, so the proper fix is always the best solution.

yajo (commented):
Although this solution will be better than #1229, it is still worse than #1227, as explained in https://github.com/OCA/server-ux/pull/1227#issuecomment-3788849963.

You should close this and merge #1227 because it is technically better, the job is already done and it benefits more people.

lmignon:
@yajo @etobella Thank you for your cooperation in finding the best approach that respects everyone's needs. I hope that beyond the revert, a consensus can be reached so that the need expressed by @yajo can be accommodated in the OCA, while taking into account the efforts he already made to consider everyone's opinions. I will be away for the next few weeks and will no longer be able to participate in this conversation.

Regards

etobella:
Yes @lmignon I think we are trying to find a way were both parties are comfortable. I think that both of us are trying to make it in a smooth and educated way even when we disagree.

etobella:
@yajo a full script has been made that should fix your concerns. Also, it duplicates range types when we made something that had no sense.

etobella:
> Although this solution will be better than #1229, it is still worse than #1227, as explained in [#1227 (comment)](https://github.com/OCA/server-ux/pull/1227#issuecomment-3788849963).
> 
> You should close this and merge #1227 because it is technically better, the job is already done and it benefits more people.

BTW, after the security bug I commented [here](https://github.com/OCA/server-ux/pull/1227#issuecomment-3789018390), I think that we need to follow this path. 

Your solution provides an extra functionality that might be interesting in some cases, but as you can see there is still some things to fix/improve. Right now we need to stabilize the module and then we can think of improvements.

etobella:
@yajo can you unblock after the latest changes? this way we can proceed with the merge and we can continue in the evolution.

yajo:
I'm consulting the feasability, hold on.

yajo:
Thanks for waiting.

We have reviewed the subject. There is no problem in creating the glue module. The problem would be to have to create many glue modules, but after deeper analysis, that doesn't seem to be the case. Maybe some modules will need small patches (adding a context key here and there), nothing huge.

Therefore, we can merge this solution, now that the migration script is corrected.

However, I suggest you to have the reversion of https://github.com/OCA/account-financial-reporting/pull/1437 ready to merge before merging this PR. Otherwise, that module will be broken again.

Thanks.

etobella:
I will handle the change once we have merged this one (I will do it right now). For doing the merge, I need yo to approve it as you are PSC and you requested changes :pray:

yajo (approved):
Of course.

/ocabot merge major

Issue #1231, opened 2026-01-27
base_search_custom_field_filter: Error when define filters with related fields


mariancuadranetkia:
## base_search_custom_field_filter

When we define a custom filter with related fields, when we try to access to model we receive error.

## To Reproduce
**Affected versions**: 18.0

1. Define a custom filter like:

<img width="975" height="35" alt="Image" src="https://github.com/user-attachments/assets/b087064e-7775-4b8c-a4e0-77f71d6ba138" />

2. Access to Sales, to check if we can use this filter. And we receive:

```
Odoo Client Error
UncaughtPromiseError > OwlError

Uncaught Promise > An error occured in the owl lifecycle (see this Error's "cause" property)

Occured on oca-server-ux-18-0-3ee738e12efa.runboat.odoo-community.org on 2026-01-27 12:36:51 GMT

OwlError: An error occured in the owl lifecycle (see this Error's "cause" property)
    OwlError@http://oca-server-ux-18-0-3ee738e12efa.runboat.odoo-community.org/web/assets/ce8f788/web.assets_web.min.js:728:1
    handleError@http://oca-server-ux-18-0-3ee738e12efa.runboat.odoo-community.org/web/assets/ce8f788/web.assets_web.min.js:972:101
    handleError@http://oca-server-ux-18-0-3ee738e12efa.runboat.odoo-community.org/web/assets/ce8f788/web.assets_web.min.js:1630:29
    initiateRender@http://oca-server-ux-18-0-3ee738e12efa.runboat.odoo-community.org/web/assets/ce8f788/web.assets_web.min.js:1064:19
    

Caused by: Error: Unknown field partner_id.lang
    visitField@http://oca-server-ux-18-0-3ee738e12efa.runboat.odoo-community.org/web/assets/ce8f788/web.assets_web.min.js:6636:97
    parse/<@http://oca-server-ux-18-0-3ee738e12efa.runboat.odoo-community.org/web/assets/ce8f788/web.assets_web.min.js:6628:296
    visit@http://oca-server-ux-18-0-3ee738e12efa.runboat.odoo-community.org/web/assets/ce8f788/web.assets_web.min.js:4793:59
    visitChildren@http://oca-server-ux-18-0-3ee738e12efa.runboat.odoo-community.org/web/assets/ce8f788/web.assets_web.min.js:4792:176
[...]

PR #1232, opened 2026-01-28
[18.0][FIX] base_search_custom_field_filter: Error when define filters with related fields

CarlosRoca13:
1. Define a custom filter with . notation like partner_id.lang
2. Access to the model defined, to check if we can use this filter.

An error was thrown

With this changes the field is defined in the correct point and the error is not thrown.

ping @pedrobaeza @carlos-lopez-tecnativa @mariancuadranetkia

cc @Tecnativa 

This PR solves the Issue: https://github.com/OCA/server-ux/issues/1231

### DIFF
--- a/base_search_custom_field_filter/models/base.py
+++ b/base_search_custom_field_filter/models/base.py
@@ -7,7 +7,6 @@
 from lxml import etree
 
 from odoo import api, models
-from odoo.tools.translate import _
 
 
 class Base(models.AbstractModel):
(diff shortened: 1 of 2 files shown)


mariancuadranetkia (approved):
I've tested in runboat and now it's working.

pedrobaeza:
@carlosdauden can you check this?

sergioM-S73 (approved):
Functional review 👍🏻
I added this commit in my migration to v19: https://github.com/OCA/server-ux/pull/1238

mariancuadranetkia:
@pedrobaeza 
I'm not sure if the errors detected in the test are only from translations. Is there any way we can help resolve this?

CarlosRoca13:
Let's see if doing a rebase the problem is solved

PR #1235, opened 2026-02-04
[16.0] [IMP] announcement: allow exclude users/groups

luisDIXMIT:
With this change, we also allow excluding specific users or groups from the user list. A page has been added to clearly show the users’ scope.

Imagine a large company with many users. They want to send a generic announcement but also need to exclude certain users. Furthermore, they may want to exclude several users who belong to a specific group.

This way, they can manage a large number of users effortlessly.

@etobella


luisDIXMIT:
If you see correct the change, I also can update the readme to explain the new functionality.

pedrobaeza:
Yes, I think it's a valid improvement of the module to complement the existing things.

luisDIXMIT:
The test error is about connection failed.

announcement/models/announcement.py
```diff
@@ -112,7 +125,10 @@ def _compute_allowed_user_ids(self):
             announcement.allowed_user_ids = announcement.specific_user_ids
             announcement.allowed_users_count = len(announcement.specific_user_ids)
         for announcement in self - specific_user_announcements:
-            announcement.allowed_user_ids = announcement.user_group_ids.users
+            announcement.allowed_user_ids = announcement.user_group_ids.users.filtered(
+                lambda u: u not in announcement.excluded_users_ids
+                and all(g not in announcement.excluded_groups_ids for g in u.groups_id)
```
etobella: all or any? 

With the current case, if you have group A, B, C and D and we set allowed A and B and excluded C and D, if someone is from A and C will be able to see it. only if they are from A, C and D, the exclusion is landed. This is intended?

luisDIXMIT: With any, users who belong to excluded groups aren't excluded.

https://github.com/user-attachments/assets/df6514a6-dd6b-4769-8181-91c68cc0e557

etobella: Ooops, I didn't read the negation inside, ok

etobella (approved):

announcement/models/announcement.py
pedrobaeza: Use sets intersection for better performance and shorter syntax:

```suggestion
                and not (u.groups_id & announcement.excluded_groups_ids))
```

luisDIXMIT: Done, thanks Pedro!

pedrobaeza (approved):
/ocabot merge minor

pedrobaeza:
This is affected by the setuptools problem. This requires https://github.com/OCA/oca-addons-repo-template/pull/333 or to switch to whool.

luisDIXMIT:
@pedrobaeza Is valid the solution I made?

pedrobaeza:
Let's try...

/ocabot merge minor

PR #1236, opened 2026-02-05
[19.0][IMP] Update substates properly when state field is computed

Camille0907:
If the state field on which the substate is based is computed, its update does not goes through `write` or `create` methods. In that cas we need to take care of the substate in ``_compute_field_value``.

As an example, you can check [stock_picking_batch_substate](https://github.com/OCA/stock-logistics-workflow/pull/2244): the batch state is computed in some cases (all picking set to done or cancel).

Unit tests have been adapted to check this case.


vvrossem (approved):

imlopes (approved):

StefanRijnhart (approved):
/ocabot merge patch

PR #1125, opened 2025-07-17
[16.0][ADD] base_tier_validation_allow_disable_restart: new module

maisim:
Add a Base Tier Validation extension module with disable restart feature

PR #1241, opened 2026-02-18
[17.0][IMP] ``base_tier_validation``: restrict validation restart

Camille0907:
Add the possiblity to restrict the validation restart to a specific group. On the tier.definition, add a field `restart_validation_group_id` (default to "base.group_user"). On tier.validation records the button "Restart validation" will be visible only if the current user is part of all `restart_validation_group_id` set on all reviews applying to the record


bosd (approved):

base_tier_validation/models/tier_definition.py
```diff
@@ -112,6 +112,12 @@ def _get_tier_validation_model_names(self):
         help="Bypassed (auto validated), if previous tier was validated "
         "by same reviewer",
     )
+    restart_validation_group_id = fields.Many2one(
+        comodel_name="res.groups",
+        required=True,
```
maisim, line 117: Should be nice to remove the required=True to allow to select no group to totally disallow restart

base_tier_validation/models/tier_validation.py
```diff
@@ -247,6 +251,22 @@ def _compute_hide_reviews(self):
...
+        defined in the implied tier reviews."""
+        for rec in self:
+            restart_groups_xmlids = (
+                rec.review_ids.restart_validation_group_id.get_external_id()
+            )
+            # Check if user belongs to all of these groups
+            rec.can_restart_validation = all(
+                self.env.user.has_group(group_xmlid)
+                for group_xmlid in restart_groups_xmlids.values()
+            )
```
maisim, line 268: This part should be replaced with something like this to totally disable restart if no group is selected
```python
restart_groups = rec.review_ids.restart_validation_group_id
if not restart_groups:
    # No  group = no restart
    rec.can_restart_validation = False
else:
    rec.can_restart_validation = all(
        self.env.user.has_group(group_xmlid)
        for group_xmlid in restart_groups.get_external_id().values()
    )
```

maisim (changes requested):
I suggest modifications so that we can completely disable restarting by not selecting a group

This would cover the need implemented in this module: https://github.com/OCA/server-ux/pull/1125

base_tier_validation/models/tier_definition.py
maisim: And I think it would also be nice to be able to select several groups, wouldn't it?

PR #1242, opened 2026-02-18
[18.0][IMP] date_range: Change where the mock has to be loaded

CarlosRoca13:
With the changes introduced in https://github.com/odoo/odoo/commit/de056cc784a3bbe2575fd3c9e81ca62e73c362d4#diff-b37c7fd5520c97a29ddc59495779e7be61a66e15e85603928e27b3affb9dc31f, an error was occurring because the change validation on the model was being performed before tearDownClass was executed. Therefore, it has been modified so that the mock changes are applied in each test and cleaned up after each test. This way, the error will no longer appear and the test will run normally.

cc @Tecnativa

ping @pedrobaeza @carlos-lopez-tecnativa

### DIFF
--- a/base_cancel_confirm/tests/test_cancel_confirm.py
+++ b/base_cancel_confirm/tests/test_cancel_confirm.py
@@ -8,27 +8,26 @@ from odoo.tests import Form, common, tagged
 
 @tagged("post_install", "-at_install")
 class TestCancelConfirm(common.TransactionCase):
-    @classmethod
-    def setUpClass(cls):
-        super().setUpClass()
+    def setUp(self):
+        super().setUp()
 
-        cls.loader = FakeModelLoader(cls.env, cls.__module__)
-        cls.loader.backup_registry()
+        self.loader = FakeModelLoader(self.env, self.__module__)
+        self.loader.backup_registry()
         from .cancel_confirm_tester import CancelConfirmTester
 
-        cls.loader.update_registry((CancelConfirmTester,))
-        cls.test_model = cls.env[CancelConfirmTester._name]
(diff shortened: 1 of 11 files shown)


carlos-lopez-tecnativa (commented):
Since `check_attrs` is added in `setUp`, loading a `FakeModel` must also be done in `setUp`, and the corresponding cleanup must be moved to `tearDown`. https://github.com/odoo/odoo/pull/247151

I think that, in the meantime, until this PR https://github.com/OCA/odoo-test-helper/pull/39
 is merged and published on PyPI, you need to add it to the requirements.

carlos-lopez-tecnativa (approved):

pedrobaeza (approved):
Thanks for the extensive work to fix it.

/ocabot merge patch

lmignon:
@CarlosRoca13 @pedrobaeza @carlos-lopez-tecnativa  For the record, another approach could have been to remove the `check_attrs` method from the cleanup methods, as I did to fix the storage repository https://github.com/OCA/storage/pull/567/changes. This isn't an easy issue, and the main goal was to find a solution to unblock all the stalled CI. Thank you for taking care of it.

pedrobaeza:
Great, thanks for the investigation. If you consider that this is going to benefit the performance a lot in this case, it can be changed to that formula.

lmignon:
> Great, thanks for the investigation. If you consider that this is going to benefit the performance a lot in this case, it can be changed to that formula.

Thank you for this proposal. You’ve put a lot of time into it. We now know that it can be improved if necessary. At this point, I suggest we leave the code as is.

PR #1239, opened 2026-02-11
[19.0][MIG] base_optional_quick_create

glitchov:
Migration from 18.0 with a new version of _patch_method because api.propagate was removed from Odoo.

PR #1249, opened 2026-03-08
[19.0][IMP] base_optional_quick_create: hide quick create in UI

gdgellatly:
## Summary

`base_optional_quick_create` has now been migrated to 19.0 and merged via #1277.

This PR has been **rebased on top of the merged module** and now contains
**only the UX improvement** — the module-migration commit it previously carried
has been dropped (see note below).

The base module only patches `name_create` server-side, so when a user clicks
"Create" in a many2one dropdown for a model with `avoid_quick_create=True`, they
get an ugly `UserError` dialog. This PR hides the "Create" option in the UI
instead (following the same pattern as `web_m2x_options`):

- Exposes the `avoid_quick_create` model list via `session_info`
- Patches `Many2One.prototype.many2XAutocompleteProps` to set `quickCreate = null`
  for those models
- The "Create and Edit..." option still works normally
- The server-side `name_create` patch is kept as a safety net

## Changes

- `models/ir_http.py` — new: pushes `avoid_quick_create` models into `session_info`
- `static/src/optional_quick_create.esm.js` — new: JS patch for `Many2One`
- `__manifest__.py` — add `web` dependency and register the assets bundle
- `models/__init__.py` — import `ir_http`

## Note

Rebased on 2026-07-22 after #1277 was merged: the branch was reset onto the
current `19.0` and only the improvement commit was cherry-picked on top. This PR
no longer supersedes #1239 (which should be closed as a duplicate migration of
#1277).


gdgellatly:
Why is pushing code, updating text and marking as ready not deemed activity?

Issue #1245, opened 2026-03-02
[18.0] base_tier_validation: Reviews pending indefinitely


victoralmau:
Although I detected this behavior in v18, it probably also occurred in previous versions.

Example use case:

- Install `purchase_request_tier_validation` (or any other)
- Modify the Marc Demo user and define the Purchase Request > Purchase Request Manager permission
- Create a Tier definition called “New Tier Validation 1,” linked to the user Marc Demo, Approve by sequence=True, Sequence=20
- Create a Tier definition called “New Tier Validation 2,” linked to the user Mitchell Admin, Approve by sequence=True, Sequence=10, Send reminder message on pending reviews=3
- Create a test user with permissions in Purchase Request
- Log in with the test user and create a Purchase request + click on the Request validation button. Two reviews will have been created: Marc Demo + Mitchell Admin
- Log in with Marc Demo to the previously created purchase request and click on Reject. The purchase request will be rejected.
- The linked review Mitchell Admin will remain pending
- Change the date (create_date) of the created tier_review records and set it to 3 days less.
- Run the “Send Tier Review Reminder” cron manually. An activity will have been created for Mitchell Admin linked to the pending review of the previously created Purchase Request. Even if the activity is canceled or completed, it will continue to be created indefinitely.

What is the appropriate behavior in this case?

IMO: If there are several pending reviews and one is rejected, the rest of the reviews should be automatically canceled.

Ping @LoisRForgeFlow 

@Tecnativa TT61255

LoisRForgeFlow:
Hi @victoralmau 

I see, what you propose 

> "IMO: If there are several pending reviews and one is rejected, the rest of the reviews should be automatically canceled." 

makes sense. The issue with that is that we would need to either remove the pending reviews loosing the ability to trace or add a new state "cancelled". The second option looks better but requires more code changes.

Thoughts?

victoralmau:
I think the best option is to use the more “complex” option of adding the cancel status. 
If you agree, I will prepare the corresponding changes.

LoisRForgeFlow:
@victoralmau I agree, let's go that way!

LoisRForgeFlow:
closed via https://github.com/OCA/server-ux/pull/1250

Issue #1082, opened 2025-05-20
base_tier_validation: tier definition 'sequence' is used as a priority


gurneyalex:
When using base_tier_definition, there is a field called "sequence" which can be used to request validations in a specific order. 

When I see such a field, I expect the records to be processed by increasing sequence (so sequence = 10 is before sequence = 20). However the [sorting in the code](https://github.com/OCA/server-ux/blob/17.0/base_tier_validation/models/tier_validation.py#L738) is done by decreasing sequence, which is super confusing for end users. Changing the sorting would likely break all instances using the feature; so not a great move. We should at least get a clear explanation in the UI.

LoisRForgeFlow:
Hi @gurneyalex 

Yep, it is an issue that has been generated at the very beggining and never fixed. There has been other attemps to fix this, see for instance https://github.com/OCA/server-ux/pull/1034

I actually think that we should use "sequence asc" as it is expected, the problem is that we need to do it at some point, add migrations scripts and propoerly warn of the change, so likely something to do only in v18. I do not see a non-painful way of fixing this...

Thoughts?

gurneyalex:
@LoisRForgeFlow a path for this could be to:

* rename the current `sequence` field to something like `importance` (which conveys the "higher is better"), have a migration script for this
* re-add a `sequence` field and make it computed, stored, with a formula such as `rec.sequence = 1000 - rec.importance`

maisim:
@gurneyalex @LoisRForgeFlow 
Yes, this is a problem for understanding. When we create the validations steps, we need to define sequences in descending order, and at "execution," the sequences are displayed in ascending order. It's a bit confusing.
Since I'm still on 16 for a while, I'm in favor of fixing this with a script that corrects it when the module is installed.
Perhaps by temporarily adding a sequence_order_fix_applied field somewhere?

If we agree on how to proceed, I can take care of it.

PR #1250, opened 2026-03-09
[18.0][IMP] base_tier_validation: Add cancel status to review

victoralmau:
Add cancel status to review

Related to https://github.com/OCA/server-ux/issues/1245

Please @pedrobaeza and @LoisRForgeFlow can you review it?

@Tecnativa TT61255

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -692,6 +692,13 @@ class TierValidation(models.AbstractModel):
         )
         # We need to notify all pending users if there is approve sequence
         if tier_reviews and any(review.approve_sequence for review in tier_reviews):
+            # If there are waiting reviews that should be approved sequentially,
+            # they must be marked as canceled
+            waiting_reviews = self.review_ids.filtered(
+                lambda r: r.status == "waiting" and r.approve_sequence
+            )
+            if waiting_reviews:
+                waiting_reviews.write({"status": "cancel"})
             reviews_to_notify = self.review_ids.filtered(
                 lambda r: r.status == "pending" and r.definition_id.notify_on_rejected
             )
--- a/base_tier_validation/static/src/components/tier_review_widget/tier_review_widget.xml
+++ b/base_tier_validation/static/src/components/tier_review_widget/tier_review_widget.xml
@@ -50,7 +50,7 @@
                                     t-key="review.id"
                                 >
                                     <t
-                                        t-if="review.status == 'waiting'"
+                                        t-if="review.status == 'waiting' or review.status == 'cancel'"
                                         t-set="status_class"
                                         t-value="'text-muted'"
                                     />
--- a/base_tier_validation/models/tier_review.py
+++ b/base_tier_validation/models/tier_review.py
@@ -23,6 +23,7 @@ class TierReview(models.Model):
             ("pending", "Pending"),
             ("rejected", "Rejected"),
             ("approved", "Approved"),
+            ("cancel", "Cancel"),
         ],
         default="waiting",
     )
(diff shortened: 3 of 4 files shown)


base_tier_validation/static/src/components/tier_review_widget/tier_review_widget.xml
```diff
@@ -50,7 +50,7 @@
                                     t-key="review.id"
                                 >
                                     <t
-                                        t-if="review.status == 'waiting'"
+                                        t-if="review.status == 'waiting' or review.status == 'cancel'"
```
pedrobaeza, line 53: Should they be shown here even if cancelled?

base_tier_validation/models/tier_validation.py
```diff
@@ -692,6 +692,13 @@ def _rejected_tier(self, tiers=False):
         )
         # We need to notify all pending users if there is approve sequence
         if tier_reviews and any(review.approve_sequence for review in tier_reviews):
+            # If there are waiting reviews that should be approved sequentially,
+            # they must be marked as canceled
+            waiting_reviews = self.review_ids.filtered(
+                lambda r: r.status == "waiting" and r.approve_sequence
```
pedrobaeza, line 698: `approve_sequence` shouldn't be higher than the current one?

base_tier_validation/static/src/components/tier_review_widget/tier_review_widget.xml
victoralmau: IMO, yes, because if someone sees a pending review and then it turns out to be confusing, knowing that it has been canceled is useful and necessary, I think.

pedrobaeza: OK, let's see @LoisRForgeFlow's opinion

base_tier_validation/models/tier_validation.py
victoralmau: The sequence works “backwards” from what we would think (definition 1 > sequence 30, definition 2 > sequence 20, definition 3 > sequence 10); In any case, I don't think it's necessary to do that validation because if it's a sequence approval, review 2 cannot be rejected without completing review 1; furthermore, if by chance it could be done, the pending reviews should still be canceled, don't you think? (review 2 is rejected, review 1 and review 3 should be canceled).

pedrobaeza: As I don't know this in-depth, again I let @LoisRForgeFlow to decide.

LoisRForgeFlow: Yes, this is a long standing issue to be honest. I agree that it should be changed to ease understanding. Some ideas have been proposed in this issue to make the transition. https://github.com/OCA/server-ux/issues/1082#issuecomment-2921394749

base_tier_validation/static/src/components/tier_review_widget/tier_review_widget.xml
LoisRForgeFlow: This makes sense from my PoV, having visibility of cancelled reviews will reduce confusion.

base_tier_validation/models/tier_validation.py
victoralmau: Okay, thanks for the clarification and for pointing out the issue, but now that it's clear, do you think it's necessary in this PR to change this filter to only “cancel” revisions with a lower sequence? In my opinion, no, because “by sequence” a revision 2 cannot be rejected if revision 1 has not been completed; therefore, the following revisions (revision 3 and revision 4, for example) will always have to be canceled.

LoisRForgeFlow: Yes, I agree that as it is now it is correct in the scope of this PR, only next review will be cancelled because previous ones are already either approved or reject (not in pending anymore), to answer initial Pedro question in a more specific way.

LoisRForgeFlow (approved):
LGTM

pedrobaeza (approved):
/ocabot merge minor

PR #1255, opened 2026-03-16
[18.0][IMP] base_tier_validation: new user should not have review_ids

ROBBYHOOD9:
### Detail pull request
When duplicating a user that already has review_ids, the new user inherits the same review_ids because the field is copied by default.

This behavior is incorrect since review_ids represent pending review assignments for a specific user. Copying them to a duplicated user can lead to inconsistent review responsibilities.

**Step Error Case**
Test with module `partner_firstname` and `base_tier_validation` together.
1. Install `partner_firstname` and `base_tier_validation`, then create a Tier Definition.
SCR-20260315-pbuh
SCR-20260315-pdkg

2. Request validation on a document so or other that a reviewer is assigned.
SCR-20260315-pdnh

3. Duplicate the user who has pending reviews.
SCR-20260315-pdvu

**Solution**
Set copy=False on the review_ids field to prevent duplicated users from inheriting existing review records.

### DIFF
--- a/base_tier_validation/models/res_users.py
+++ b/base_tier_validation/models/res_users.py
@@ -7,7 +7,9 @@ from odoo import api, fields, models, modules
 class Users(models.Model):
     _inherit = "res.users"
 
-    review_ids = fields.Many2many(string="Reviews", comodel_name="tier.review")
+    review_ids = fields.Many2many(
+        string="Reviews", comodel_name="tier.review", copy=False
+    )
 
     @api.model
     def review_user_count(self):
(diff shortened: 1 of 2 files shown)


Saran440 (approved):
LGTM 👍
When duplicating a user, the new user should not copy `review_ids`.

Pani-k-folk (approved):
code LGTM 👍

LoisRForgeFlow (approved):
Thanks for the explanation and test case 👍 

/ocabot merge minor

PR #1256, opened 2026-03-18
[IMP] base_tier_validation: add configurable position for reviews widget

mathben:
Add `_tier_validation_reviews_position` class attribute to the `TierValidation` mixin to control whether the reviews widget is rendered at the top or bottom of the form sheet.

Default value is `"bottom"` to preserve existing behavior. Set `"top"` on any inheriting model to display reviews above all form fields, which improves usability on mobile views where scrolling to the bottom is less convenient.

Child models can now configure this per-model:

      _tier_validation_reviews_position = "top"

This feature was created to position the validation at the top, as when the record description contains a large amount of text, the validation area becomes hard to find. This option is configurable.

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -27,6 +27,7 @@ class TierValidation(models.AbstractModel):
     _tier_validation_manual_config = True
     _tier_validation_state_field_is_computed = False
     _tier_validation_company_field = "company_id"
+    _tier_validation_reviews_position = "bottom"
 
     _state_field = "state"
     _state_from = ["draft"]
@@ -906,7 +907,10 @@ class TierValidation(models.AbstractModel):
                 new_node = self._add_tier_validation_reviews(node, params)
                 new_arch, new_models = View.postprocess_and_fields(new_node, self._name)
                 new_node = etree.fromstring(new_arch)
-                node.append(new_node)
+                if self._tier_validation_reviews_position == "bottom":
+                    node.append(new_node)
+                else:
+                    node.insert(0, new_node)
                 _merge_view_fields(all_models, new_models)
             excepted_fields = self._get_all_validation_exceptions()
             all_fields = self.fields_get(attributes=("readonly",))


mathben:
It's missing file DEVELOP.md from https://github.com/OCA/server-ux/tree/18.0/base_tier_validation/readme
Information about how develop for specific model with base_tier_validation and customize it. 

@LoisRForgeFlow 
Do you like this quantity of information?
Thanks

```
## Architecture Overview

`base_tier_validation` provides an abstract mixin `TierValidation`
(`tier.validation`) that any Odoo model can inherit to gain a
configurable multi-level approval workflow.

The mixin injects validation buttons, status labels, and a reviews
widget directly into the form view at render time via `get_view()`
override, without requiring manual view changes in inheriting modules
(when `_tier_validation_manual_config = False`).

### Key class attributes

| Attribute | Default | Description |
|---|---|---|
| `_tier_validation_buttons_xpath` | `"/form/header/button[last()]"` | XPath where validation buttons are injected |
| `_tier_validation_manual_config` | `True` | If `False`, buttons/reviews are auto-injected |
| `_tier_validation_state_field_is_computed` | `False` | Set `True` if the state field is a computed field |
| `_tier_validation_company_field` | `"company_id"` | Field used to resolve the company for tier definitions |
| `_tier_validation_reviews_position` | `"bottom"` | Position of the reviews widget in the form sheet (`"top"` or `"bottom"`) |
| `_state_field` | `"state"` | Name of the state field on the model |
| `_state_from` | `["draft"]` | States from which validation can be requested |
| `_state_to` | `["confirmed"]` | States that trigger the validation check |
| `_cancel_state` | `"cancel"` | Cancel state — bypasses validation and removes reviews |

### Overridable hooks

| Method | Purpose |
|---|---|
| `_validated_states()` | List of `tier.review` statuses considered as approved (default: `["approved"]`) |
| `_rejected_states()` | List of `tier.review` statuses considered as rejected (default: `["rejected"]`) |
| `_get_to_validate_message_name()` | Label used in the "needs validation" info banner |
| `_get_under_validation_exceptions()` | Fields allowed to be written while under validation |
| `_get_after_validation_exceptions()` | Fields allowed to be written after validation |
| `_get_all_validation_exceptions()` | All fields excluded from the readonly lock |
| `_prepare_tier_review_vals()` | Customize values when creating `tier.review` records |
| `_allow_to_remove_reviews()` | Control when existing reviews are deleted on state change |
| `_notify_accepted_reviews_body()` | Body of the chatter message posted on approval |
| `_notify_rejected_review_body()` | Body of the chatter message posted on rejection |

## Extending the Mixin

### Minimal example

```python
from odoo import models

class MySaleOrder(models.Model):
    _name = "sale.order"
    _inherit = ["sale.order", "tier.validation"]

    _state_from = ["draft"]
    _state_to = ["sale"]
    _cancel_state = "cancel"

    # Show reviews at the top of the form
    _tier_validation_reviews_position = "top"
```
```

PR #1257, opened 2026-03-18
[18.0][IMP] document_quick_access_folder_auto_classification: add a fallback method to decode with contrast

ArnauCForgeFlow:
Adds a fallback method, that tries to add more contrast to the image and decode it using zbar.


etobella (approved):

LoisRForgeFlow (approved):
This reduces a bit the fail ratio 👍

PR #1223, opened 2026-01-20
[18.0][ADD] base_tier_validation_correction_rules

Saran440:
This module enhance from `base_tier_validation_correction`
1. When user permission Tier Review Correction click Change Reviewer. it will display wizard instead new document and allow to change New Reviewer from rule. If not config in correction rule, it will change New Reviewer to all users
selection_046
selection_047

2. Config rules on menu Settings > Technical > Tier Review Correction Rules
    - User - the user who will be performing the reviewer change.
    - Allowed Reviewers - the users that this specific User is allowed to select as new approvers.
selection_050

PR #1260, opened 2026-03-21
[18.0][IMP] base tier validation correction rules hide change reviewer

ROBBYHOOD9:
### Ref
- Original discussion / reference: https://github.com/OCA/server-ux/pull/1223

### Problem
- In tier validation, when a tier definition is configured with review_type = group, the UI still displays the "Change Reviewer" button. That's make error this:

SCR-20260321-bkmx

### Solution
This PR introduces a UI improvement:

- Add condition to hide "Change Reviewer" button when `is_group_defination == True`
- `is_group_defination` is true when `tier definition review_type == "group"`
- Implemented via template inheritance `<attribute name="invisible">is_group_defination</attribute>`

SCR-20260321-bkar
SCR-20260321-bkcz


Saran440:
Don't create new PR because `base_tier_validation_correction_rules` is not merged in OCA.
Please move your commit into https://github.com/OCA/server-ux/pull/1223

PR #1263, opened 2026-03-24
[15.0][ADD] base_view_action_visibility

AungKoKoLin1997:
This module allows hiding standard view actions (Duplicate and Delete) on a per-model basis through configuration on the ir.model settings. You can hide these actions for all users, or hide them and specify which security groups are allowed to still see them.

@qrtl QT6466


base_view_action_visibility/models/ir_model.py
```diff
@@ -0,0 +1,17 @@
...
+    _inherit = "ir.model"
+
+    hide_action_duplicate = fields.Boolean(
+        string="Hide Duplicate Action",
+        help="Hide the 'Duplicate' action in form views for this model",
+    )
+    hide_action_delete = fields.Boolean(
+        string="Hide Delete Action",
+        help="Hide the 'Delete' action in form and tree views for this model",
+    )
```
yostashiro: Did you consider assigning groups here instead of having boolean fields?

AungKoKoLin1997: No, I didn't. Seem like it is better design. I will update to use groups fields.

AungKoKoLin1997:
pre-commit issue fixes at https://github.com/OCA/server-ux/pull/1264.

base_view_action_visibility/models/models.py
```diff
@@ -0,0 +1,49 @@
...
+        if view_type not in ("form", "tree") or not ir_model:
+            return result
+
+        # Check if we need to apply restrictions based on groups
+        has_duplicate_restriction = bool(ir_model.duplicate_allowed_group_ids)
+        has_delete_restriction = bool(ir_model.delete_allowed_group_ids)
+
+        if not has_duplicate_restriction and not has_delete_restriction:
+            return result
+
```
yostashiro:
```suggestion
        if view_type == "tree" and not has_delete_restriction:
            return result
```

base_view_action_visibility/views/ir_model_views.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="ir_model_view_form" model="ir.ui.view">
+        <field name="model">ir.model</field>
+        <field name="inherit_id" ref="base.view_model_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//notebook" position="before">
```
yostashiro:
```suggestion
            <xpath expr="//sheet/notebook" position="before">
```

base_view_action_visibility/models/models.py
```diff
@@ -0,0 +1,49 @@
...
+    def fields_view_get(
+        self, view_id=None, view_type="form", toolbar=False, submenu=False
+    ):
+        result = super().fields_view_get(
+            view_id=view_id, view_type=view_type, toolbar=toolbar, submenu=submenu
+        )
+        ir_model = (
+            self.env["ir.model"].sudo().search([("model", "=", self._name)], limit=1)
+        )
+        if view_type not in ("form", "tree") or not ir_model:
```
yostashiro: Can we also handle kanban?

yostashiro:
Just attaching some screen shots...
image
image

base_view_action_visibility/models/models.py
AungKoKoLin1997: No. The duplicate and delete action are worked on the selected records. We can't select records in kanban view.

AungKoKoLin1997: @yostashiro I realized that the delete action can also be used in the Kanban view by defining it in the Kanban view template with the `dropdown-menu` class. Odoo standard already uses this dropdown for some models, such as Project Tasks and CRM Pipelines. So, I handled the Kanban view as well.

base_view_action_visibility/models/models.py
```diff
@@ -0,0 +1,44 @@
...
+            return result
+        if view_type in ("tree", "kanban") and not has_delete_restriction:
+            return result
+        hide_duplicate = self._should_hide_action(ir_model.duplicate_allowed_group_ids)
+        hide_delete = self._should_hide_action(ir_model.delete_allowed_group_ids)
+        arch = etree.fromstring(result["arch"])
+        if view_type == "form" and has_duplicate_restriction:
+            arch.set("duplicate", "0" if hide_duplicate else "1")
+        if has_delete_restriction:
+            arch.set("delete", "0" if hide_delete else "1")
```
yostashiro: We can simplify the code.
```suggestion
        hide_duplicate = self._should_hide_action(ir_model.duplicate_allowed_group_ids)
        hide_delete = self._should_hide_action(ir_model.delete_allowed_group_ids)
        if not hide_duplicate and not hide_delete:
            return result
        arch = etree.fromstring(result["arch"])
        if view_type == "form" and hide_duplicate:
            arch.set("duplicate", "0")
        if hide_delete:
            arch.set("delete", "0")
```

The current code has an issue of giving duplicate/delete permission to users even when Odoo's view defines otherwise, and this suggestion should fix it.

base_view_action_visibility/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,6 @@
+This module allows restricting standard view actions (Duplicate and Delete) on a per-model
+basis through configuration on the ir.model settings. You can restrict these actions to
+specific security groups, or leave them unrestricted to allow all users to see them.
+
+The Duplicate action restriction applies to form views, while the Delete action restriction
+applies to form, tree, and kanban views.
```
yostashiro, line 6:
```suggestion
applies to form, tree, and kanban views.

Note: This module controls the visibility of UI actions only. It does not enforce server-side restrictions.
```

yostashiro (changes requested):

base_view_action_visibility/readme/DESCRIPTION.rst
AungKoKoLin1997: Updated!

base_view_action_visibility/models/models.py
AungKoKoLin1997: Done!

yostashiro (approved):
Code review. LGTM.

smorita7749 (approved):
Functional review LGTM

Added an user to a group and set the group in Duplicate Allowed Groups.
Set another group(no user) in Delete Allowed Groups.

Then, the user can see only Duplicate in Action.

Image

Image

nobuQuartile (approved):
Functional review: LGTM

PR #1266, opened 2026-03-25
Updated from template

SirPyTech:
To include https://github.com/OCA/oca-addons-repo-template/pull/333.


StefanRijnhart (approved):

StefanRijnhart:
Fasttracking as the same change is being proposed on many repos

/ocabot merge nobump

PR #1267, opened 2026-03-26
[18.0][FIX] date_range: Tree editor mark 'date end' as end

qgroulard:
Without this fix we skip two days (first and last) when using date range domains, because the first day was set to start at 23:59:59 and the last day to finish at 00:00:00.

Before:

image

After:

image


lmignon (approved):
Thank you for the fix @qgroulard

samirGuesmi (approved):
LGTM

rousseldenis:
@OCA/internal-tools

rousseldenis:
@OCA/tools-maintainers

PR #1275, opened 2026-04-07
[18.0][FIX] date_range: migration script

psugne:
Fix failure when loading migration:
`File ".../odoo/modules/migration.py", line 256, in exec_script TypeError: module date_range: migrate's signature should be (cr, version)`

Migration is being executed by core Odoo migration loader, not by OpenUpgrade’s migration system.


flotho (approved):
easy code review
LGTM
Thnaks

pedrobaeza (commented):
This is not a fix, as it was working previously. You just need openupgradelib.

PR #1220, opened 2026-01-18
[18.0][IMP] base_custom_filter: add configurable insertion point for filter groups, etc.

yostashiro:
- Add insert_xpath, insert_position, and separator_position fields to control filter placement in search views
- Fix custom filters appearing under Favorites
- Fix custom filters not showing when //search/filter node didn't exist
- Make ir.filters name field translatable
- Add date filter support

@qrtl QT6218

PR #1276, opened 2026-04-07
[18.0][IMP] base_custom_filter: Add date filter support

AungKoKoLin1997:
Allow creation of date filters with automatic period options (Today, This Week, This Month, This Quarter, This Year).

Users can now select a date/datetime field when creating Standard Filters as an alternative to domain-based filtering.

@qrtl QT6565


base_custom_filter/models/ir_filters.py
```diff
@@ -37,9 +38,55 @@ def _selection_type(self):
...
+                    )
+                )
+            if not has_domain and not record.date_field:
+                raise ValidationError(
+                    _(
+                        "Filter '%(name)s': You must set either Domain or "
+                        "Date Field for filter type.",
+                        name=record.name,
+                    )
+                )
```
yostashiro, line 73: I think we should drop this. It may cause an issue with module update unless a migration script is provided, and there is not much point being strict with this check.
```suggestion
```

base_custom_filter/models/ir_filters.py
```diff
@@ -37,9 +38,55 @@ def _selection_type(self):
...
+
+    @api.onchange("type")
+    def _onchange_type_clear_date_field(self):
+        if self.type != "filter":
+            self.date_field = False
```
yostashiro, line 78: Can you consider turning it into a compute method?

base_custom_filter/models/ir_filters.py
```diff
@@ -37,9 +38,55 @@ def _selection_type(self):
...
+
+    @api.onchange("domain")
+    def _onchange_domain_clear_date_field(self):
+        if self.domain and self.domain != "[]" and self.date_field:
+            self.date_field = False
+
+    @api.onchange("date_field")
+    def _onchange_date_field_clear_domain(self):
+        if self.date_field:
+            self.domain = "[]"
```
yostashiro, line 88: Can you consider turning it into an inverse method for date_field?

yostashiro (changes requested):
@AungKoKoLin1997 Do you think this enhancement should be included in https://github.com/OCA/server-ux/pull/1220 since we refactor the code there?

AungKoKoLin1997:
> Do you think this enhancement should be included in #1220 since we refactor the code there?

@yostashiro Yes. I think so since the reviewers are only from other team at the moment. I will add as a new commit.

PR #1279, opened 2026-04-07
[18.0][FIX] base_tier_validation_forward: fix forward notification delivery

llabusch93:
## Summary

Two bugs in `base_tier_validation_forward` prevent forwarded review notifications from reaching the target user:

**Bug 1 — Wrong `subtype_xmlid`:** `_get_forwarded_notification_subtype()` returns `"base_tier_validation.mt_tier_validation_forwarded"` but the `mail.message.subtype` record is defined in the `base_tier_validation_forward` module (see `data/mail_data.xml`). When `message_post()` cannot resolve the xmlid, it silently falls back to `mail.mt_note` — internal notes do **not** generate email notifications for followers.

**Bug 2 — Forward target not subscribed as follower:** The forward wizard `add_forward()` creates a `tier.review` for the target user but never calls `message_subscribe()`. Without being a follower of the record, the target user cannot receive email notifications even if the subtype were correct.

**Combined effect:** When a reviewer forwards their approval to another user, the target user is never notified by email. The forward "works" technically (the review record is created), but the target user has no way of knowing they need to act — unless they happen to be actively browsing the record's chatter.

## Changes

- **`models/tier_validation.py`**: Fix the module prefix in `_get_forwarded_notification_subtype()` from `base_tier_validation` → `base_tier_validation_forward`
- **`wizard/forward_wizard.py`**: Call `message_subscribe()` in `add_forward()` to subscribe the forward target as a follower with the forward notification subtype
- **`tests/test_tier_validation.py`**: Add 3 tests — subtype xmlid correctness, message subtype verification, follower subscription after forward

### DIFF
--- a/base_tier_validation_forward/models/tier_validation.py
+++ b/base_tier_validation_forward/models/tier_validation.py
@@ -26,7 +26,7 @@ class TierValidation(models.AbstractModel):
         return res
 
     def _get_forwarded_notification_subtype(self):
-        return "base_tier_validation.mt_tier_validation_forwarded"
+        return "base_tier_validation_forward.mt_tier_validation_forwarded"
 
     def forward_tier(self):
         self.ensure_one()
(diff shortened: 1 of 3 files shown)


Saran440 (approved):
Code Review 👍 

@llabusch93 Could you change Title to add version, please?
`[18.0][FIX] base.....`

llabusch93:
@Saran440 Done — updated the title to `[18.0][FIX] base_tier_validation_forward: fix forward notification delivery`.

PR #1280, opened 2026-04-10
[ADD] announcement: add Chinese (Simplified) translation

laofahai:
## Summary
- Add `zh_CN.po` translation file for the `announcement` module
- Covers all 63 translatable strings from `announcement.pot`
- Includes menus, fields, actions, tags, wizard labels and UI text

## Context
The announcement module currently only has Spanish (es) and Italian (it) translations. This PR adds Chinese (Simplified) to improve accessibility for Chinese-speaking Odoo users.


legalsylvain:
use weblate.

PR #1281, opened 2026-04-10
[18.0][IMP] base_tier_validation: increase review limit for reminder cron

Killance:
The search on pending reviews is currently limited to 1. If we have the case where a company has plenty of reviews, even if the cron runs every minute (the standard minimum in cron configuration), it takes a lot of time to browse all pending reviews. I suggest to increase the limit and adapt the called action to fit with a recordset.


bosd (approved):
LGTM

PR #1287, opened 2026-04-30
[16.0][ADD] base_automation_sales_team: subscribe teams via automated actions

nicolas-delbovier-acsone:
This new module extends the Automated Actions functionality to allow subscribing entire Sales Teams (crm.team) to a record.

image


pedrobaeza:
Shouldn't this be in OCA/automation?

nicolas-delbovier-acsone:
> Shouldn't this be in OCA/automation?

Fair point, I'll move it to there

PR #1289, opened 2026-04-30
[18.0][FIX] server_action_mass_edit: add onchange_helper dependency

fsmw:
Closes: #3368 (reported on OCA/server-tools, but module is in server-ux)

**Problem:**
Installing `server_action_mass_edit` alone and creating a mass editing rule for relational fields (like `categ_id`) crashes with `KeyError: 'selection__categ_id'`. The wizard's `fields_get()` and `onchange()` methods generate dynamic fields, but the underlying `onchange_helper` infrastructure is missing.

**Fix:**
Add `onchange_helper` to the `depends` list in `__manifest__.py`.

**Why not `server_action_mass_edit_onchange`:**
Adding `server_action_mass_edit_onchange` directly would create a circular dependency (`mass_edit` depends on `mass_edit_onchange` depends on `mass_edit`). `onchange_helper` is the common upstream dependency that breaks the cycle.

### DIFF
--- a/server_action_mass_edit/__manifest__.py
+++ b/server_action_mass_edit/__manifest__.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     "name": "Mass Editing",
-    "version": "18.0.1.1.3",
+    "version": "18.0.1.1.4",
     "author": "Serpent Consulting Services Pvt. Ltd., "
     "Tecnativa, "
     "GRAP, "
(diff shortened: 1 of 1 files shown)


legalsylvain (changes requested):
-

fsmw:
Closing this PR — the proposed fix (adding onchange_helper as dependency) is incorrect. The server_action_mass_edit module does not use onchange_helper anywhere in its code. Adding unused dependencies is not the right approach.

The real issue is that a circular dependency prevents adding server_action_mass_edit_onchange as a dependency. The architecture is designed as downstream extension. The proper fix would need a different approach — perhaps fixing the crash at a deeper level or documenting the dependency in README.

PR #1302, opened 2026-06-12
[18.0] [FIX] server_action_mass_edit: delete inactive views in tests

AxelPrel:
If a module deactivates a res partner view, then the test will fail, because the view will not be fetched by the search and this will cause a SQL error, because it is trying to remove the parent view before the inactive view is unlinked


Pani-k-folk (approved):
Code LGMT 👍

pedrobaeza (approved):
/ocabot merge patch

PR #1305, opened 2026-06-25
[15.0][IMP] template_content_swapper: allow manual editing of the name

smorita7749:
This change allows users to edit the mapping name manually.
@qrtl QT6819

### DIFF
--- a/template_content_swapper/models/template_content_mapping.py
+++ b/template_content_swapper/models/template_content_mapping.py
@@ -13,7 +13,7 @@ class TemplateContentMapping(models.Model):
     def _lang_get(self):
         return self.env["res.lang"].get_installed()
 
-    name = fields.Char(compute="_compute_name", store=True, readonly=True)
+    name = fields.Char(required=True)
     report_id = fields.Many2one("ir.actions.report")
     template_id = fields.Many2one(
         "ir.ui.view",
@@ -37,14 +37,18 @@ class TemplateContentMapping(models.Model):
         help="Set your new content (string). e.g. 'Sales Representative'.",
     )
 
-    @api.depends("content_from", "content_to")
-    def _compute_name(self):
-        for record in self:
-            record.name = False
-            if record.content_from:
-                record.name = (
-                    f"{record.content_from or ''} -> {record.content_to or ''}"
-                )
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if not vals.get("name") and vals.get("content_from"):
+                vals["name"] = vals["content_from"]
+        return super().create(vals_list)
+
+    @api.onchange("content_from")
+    def _onchange_content(self):
+        """Auto-fill the name from the contents while empty; keep manual names."""
+        if not self.name and self.content_from:
+            self.name = self.content_from
 
     @api.onchange("report_id")
     def _onchange_report_id(self):


template_content_swapper/models/template_content_mapping.py
```diff
@@ -13,7 +13,7 @@ class TemplateContentMapping(models.Model):
     def _lang_get(self):
         return self.env["res.lang"].get_installed()
 
-    name = fields.Char(compute="_compute_name", store=True, readonly=True)
+    name = fields.Char()
```
AungKoKoLin1997: Isn't it possible just by changing `readonly` to `False`?

AungKoKoLin1997: Do you want to add name before inserting content values? Is that the reason?

yostashiro:
```suggestion
    name = fields.Char(required=True)
```

template_content_swapper/models/template_content_mapping.py
```diff
@@ -37,14 +37,11 @@ def _lang_get(self):
...
-            record.name = False
-            if record.content_from:
-                record.name = (
-                    f"{record.content_from or ''} -> {record.content_to or ''}"
-                )
+    @api.onchange("content_from", "content_to", "name")
+    def _onchange_content(self):
+        """Auto-fill the name from the contents while empty; keep manual names."""
+        if not self.name and self.content_from and self.content_to:
+            self.name = f"{self.content_from} -> {self.content_to}"
```
yostashiro: To keep it simple, as names currently tend to look cluttered.
```suggestion
    @api.onchange("content_from")
    def _onchange_content(self):
        """Auto-fill the name from the contents while empty; keep manual names."""
        if not self.name and self.content_from:
            self.name = self.content_from
```

template_content_swapper/models/template_content_mapping.py
smorita7749: The intent is to preserve a manually entered name even after content_from / content_to are changed.
I reflected the review for Yoshi-san, so please take a look when you have time.

template_content_swapper/models/template_content_mapping.py
smorita7749: This change makes sense. I have updated the code.

template_content_swapper/models/template_content_mapping.py
smorita7749: Just a question, but is `required=True` necessary? CI shows failing with NOT NULL constraint violation since the name is empty when a record is created. I think override create() method is too much to keep `required=True`, so I think `required=True` is not necessary.

yostashiro: Not a must, but we want to have a reasonable way to ensure the field is updated. I actually think overriding create is a reasonable arrangement, similar to how it's handled in https://github.com/oca/ocb/blob/58e7814/addons/stock/models/stock_picking.py#L721-L723.

smorita7749: Thank you for your point! Updated the code to override the create method.

template_content_swapper/models/template_content_mapping.py
```diff
@@ -37,14 +37,18 @@ def _lang_get(self):
...
-            record.name = False
-            if record.content_from:
-                record.name = (
-                    f"{record.content_from or ''} -> {record.content_to or ''}"
-                )
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if not vals.get("name") and vals.get("content_from"):
+                vals["name"] = vals["content_from"]
```
AungKoKoLin1997, line 44: Shouldn't we keep original practice `f"{record.content_from or ''} -> {record.content_to or ''}"`? Is there any reason to use only `content_from`?

template_content_swapper/models/template_content_mapping.py
```diff
@@ -37,14 +37,18 @@ def _lang_get(self):
...
+        return super().create(vals_list)
+
+    @api.onchange("content_from")
+    def _onchange_content(self):
+        """Auto-fill the name from the contents while empty; keep manual names."""
+        if not self.name and self.content_from:
+            self.name = self.content_from
```
AungKoKoLin1997, line 51: Same as above.
Isn't better to use `write` method?

AungKoKoLin1997 (commented):
Instead of adding logic in both `create` and `onchange`, I am wondering why can't we keep original approach just by adding readonly=False attribute. The original compute method depends on `content_from` and `content_to`. So, user can still change the name after these values are filled. IMO, changing the name when `content_from` or `content_to` is make sense and the name should be updated.

smorita7749:
> Instead of adding logic in both `create` and `onchange`, I am wondering why can't we keep original approach just by adding readonly=False attribute. The original compute method depends on `content_from` and `content_to`. So, user can still change the name after these values are filled. IMO, changing the name when `content_from` or `content_to` is make sense and the name should be updated.

The editable-computed way would overwrite the name, but that's what we want to avoid. `content_from` / `content_to` can hold long text or markup, so a name built from them is often hard to read. So we want to let users set their own clear name, and keep it even when `content_from` / `content_to` are changed later. The auto-fill is just a helper to set a default name while it's empty, and once the user sets a name, we keep it. That's why `name` is a normal field filled in `create` / `onchange` instead of a computed one.

template_content_swapper/models/template_content_mapping.py
smorita7749: In the original computed version the name was always rebuilt from `content_from` / `content_to`, so including `content_to` made sense.

Now the record is created as soon as `content_from` (a required field) is filled, and `content_to` is optional and often not set yet at that point. So I guess there's no need to build the name from `content_to` here.

template_content_swapper/models/template_content_mapping.py
smorita7749: `name` is `required=True`, so a saved record always has a name, and a `write` override would have nothing to fill. The `onchange` fills a default in the form while the record's name is still empty before saving.

template_content_swapper/models/template_content_mapping.py
AungKoKoLin1997: What I am not comfortable is current improvement breaks the original behavior and always forces us to put the manual name. If we don't do it, the name will always show "content_from" only. I understand the name field in the config is not in important role at the moment but I don't want to break existing behavior and forcing a manual name.

PR #1306, opened 2026-06-26
[19.0][MIG] filter_multi_user

samirGuesmi:


filter_multi_user/models/ir_filters.py
```diff
@@ -24,46 +21,6 @@ class IrFilters(models.Model):
...
-                "is_default",
-                "domain",
-                "context",
-                "user_id",
-                "sort",
-                "embedded_action_id",
-                "embedded_parent_res_id",
-            ]
-        )
+            rec.user_ids = rec.manual_user_ids | rec.group_ids.user_ids
```
petrus-v: I've never seen the usage of `api.constrains` to change the state of recordset, are you sure it's good idea ?

I would prefer to explicitly make the field compute/store with  @api.depends instead.

samirGuesmi: I've changed it to synchronize the field from create() and write() instead

petrus-v: I guess api.depends would have updated fields in the UI for better experience but this version seems ok to me

petrus-v (approved):

lmarion-source (approved):

BhaveshHeliconia (approved):

prayagrk (approved):

samirGuesmi:
Hello @OCA/tools-maintainers,
This PR has the Approved label. Would it be possible to merge it please?

PR #1307, opened 2026-07-02
[16.0][IMP] base_tier_validation: extract _remove_reviews hook

smorita7749:
Extract the review removal in `write()` / `restart_validation()` into a `_remove_reviews()` method, so other modules can override it to do something other than delete the reviews (e.g. archive them as history).

@qrtl QT6878

### DIFF
--- a/base_tier_validation/models/tier_validation.py
+++ b/base_tier_validation/models/tier_validation.py
@@ -444,7 +444,7 @@ class TierValidation(models.AbstractModel):
                     }
                 )
             if rec._allow_to_remove_reviews(vals):
-                rec.mapped("review_ids").unlink()
+                rec._remove_reviews()
         return super(TierValidation, self).write(vals)
 
     def _allow_to_remove_reviews(self, values):
@@ -754,7 +754,7 @@ class TierValidation(models.AbstractModel):
                         .mapped("partner_id")
                         .ids
                     )
-                rec.mapped("review_ids").unlink()
+                rec._remove_reviews()
                 if to_update_counter:
                     self._update_counter({"review_deleted": True})
             if partners_to_notify_ids:
@@ -785,6 +785,10 @@ class TierValidation(models.AbstractModel):
         channel = "base.tier.validation/updated"
         notifications.append([self.env.user.partner_id, channel, review_counter])
         self.env["bus.bus"]._sendmany(notifications)
+
+    def _remove_reviews(self):
+        """Remove the reviews. Hook to allow overriding the removal behavior."""
+        self.review_ids.unlink()
 
     def unlink(self):
         self.mapped("review_ids").unlink()


base_tier_validation/models/tier_review.py
```diff
@@ -73,6 +79,7 @@ class TierReview(models.Model):
         related="definition_id.approve_sequence_bypass", readonly=True
     )
     last_reminder_date = fields.Datetime(readonly=True)
+    active = fields.Boolean(default=True)
```
yostashiro: I think this approach could be a bit risky, as we will need to consider this field in various compute methods, such as  _compute_validation_status, _compute_need_validation, and etc., and it might break the logic of some extension modules. Can we have a separate model to keep the history instead of keeping "archived" reviews? In that case, we should probably create a new module base_tier_validation_review_history rather than working on the base module.

base_tier_validation/models/tier_validation.py
```diff
@@ -444,9 +451,29 @@ def write(self, vals):
                     }
                 )
             if rec._allow_to_remove_reviews(vals):
-                rec.mapped("review_ids").unlink()
+                rec._remove_reviews()
```
yostashiro, line 447: This hook will be necessary, so we may shrink the scope of this PR to this minor refactoring.

yostashiro (changes requested):

base_tier_validation/models/tier_review.py
smorita7749: I decided to extract the history feature into #1311.

base_tier_validation/models/tier_validation.py
smorita7749: Done!

base_tier_validation/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Base Tier Validation",
     "summary": "Implement a validation process based on tiers.",
-    "version": "16.0.4.0.2",
+    "version": "16.0.4.1.0",
```
AungKoKoLin1997: Better to let ocabot handle that.

smorita7749: Done!

base_tier_validation/models/tier_validation.py
```diff
@@ -786,6 +786,10 @@ def _update_counter(self, review_counter):
         notifications.append([self.env.user.partner_id, channel, review_counter])
         self.env["bus.bus"]._sendmany(notifications)
 
+    def _remove_reviews(self):
+        """Remove the reviews. Hook to allow overriding the removal behavior."""
```
AungKoKoLin1997, line 790: Shouldn't this method be above `write()`?
```suggestion
    def _remove_reviews(self):
        """Remove the reviews. Hook to allow overriding the removal behavior."""
        self.ensure_one()
```

smorita7749: I moved down since the method was called from up and down.
Is it better to be close to write()?

AungKoKoLin1997: The ordering in this file is not following “callee before caller” pattern. So, you can keep your order but I prefer above because it is better for readability.

yostashiro (approved):
Code review. LGTM.

AungKoKoLin1997 (approved):

PR #1188, opened 2025-11-11
[19.0][MIG] base_export_manager: Migration to 19.0

lef-adhoc:

PR #1309, opened 2026-07-03
[19.0][MIG] base_export_manager: Migration to 19.0

lbisiach:
## Summary

Migration of `base_export_manager` to Odoo 19.0, including a fix for a crash during module data loading.

### Changes

- **Migration to 19.0**: Updated module structure, views, and JS assets for Odoo 19 compatibility.
- **Fix `_inverse_name` crash during data loading**: When Odoo loads data for modules like `project_enterprise`, it creates `ir.exports.line` records via XML. If the parent `ir.exports` record's `model_id` is not yet computed (e.g. version mismatch between CE and EE, or records pre-dating this module), `_inverse_name` would raise a `ValidationError` that propagated as a `ParseError` and prevented the registry from loading entirely. The fix skips field resolution gracefully when the model is unknown.
- **Improved `post_init_hook`**: Now recomputes `model_id` for existing `ir.exports` records that have a `resource` but no `model_id` stored, before processing lines.

## Test plan

- [ ] Install `base_export_manager` on a fresh Odoo 19 database with `project` and `project_enterprise` installed — server should start without errors
- [ ] Verify existing export profiles load correctly after module installation
- [ ] Verify creating a new export profile from the UI works as expected
- [ ] Run module tests: `base_export_manager`


victoralmau:
I see there's another migration PR: https://github.com/OCA/server-ux/pull/1188

PR #1313, opened 2026-07-15
[18.0][MIG] base_user_locale: Migration to 18.0

anusriNPS:
Migrating `base_user_locale` module to 18.0


anusriNPS:
**Note:** One of the testcase is removed `test_uninstalled_lang` from the commit, as `ValueError` is no longer raised when an user is created with `uninstalled language`.

base_user_locale/views/res_users.xml
```diff
@@ -0,0 +1,21 @@
...
+        <field name="name">res.users.preferences.form</field>
+        <field name="model">res.users</field>
+        <field name="inherit_id" ref="base.view_users_form_simple_modif" />
+        <field name="arch" type="xml">
+            <field name="tz" position="before">
+                <field name="date_format" invisible="lang == 'False'" />
+                <field name="time_format" invisible="lang == 'False'" />
+                <field name="week_start" invisible="lang == 'False'" />
+                <field name="decimal_point" invisible="lang == 'False'" />
+                <field name="thousands_sep" invisible="lang == 'False'" />
```
HekkiMelody: **issue**: don't check if lang is equal to the string False


```suggestion
                <field name="date_format" invisible="not lang" />
                <field name="time_format" invisible="not lang" />
                <field name="week_start" invisible="not lang" />
                <field name="decimal_point" invisible="not lang" />
                <field name="thousands_sep" invisible="not lang" />
```

HekkiMelody (commented):
Functional review, LGTM
Code review ok aside from the one issue (couldn't reproduce functionally since you can't empty the `lang` field from the interface)

base_user_locale/views/res_users.xml
anusriNPS: Wrong value was used as a `string` while adapting to version 18.0. Corrected now as suggested

HekkiMelody (approved):

Pani-k-folk (approved):
Functional Review👍 
Code LGTM 👍

LoisRForgeFlow (approved):
👍

PR #1320, opened 2026-08-03
[19.0][ADD] base_field_length_constraint

AungKoKoLin1997:
This module enforces a maximum length on `char`, `text` and `html` fields,
defined as configuration data rather than in code.

A rule declares the limit of one field of one model, counted in characters or
in the bytes of a given encoding, and optionally restricted to a company or to
the records matching a domain. An over-long value is refused when the record
is saved, or only reported if the rule is set to warn.

@qrtl QT7079


AungKoKoLin1997:
I added `mute_logger` on the test that exercises the violation cap: that path logs a warning by design, and the warning showing up in the Odoo log makes CI fail. Muting the module's logger keeps the run clean.

kanda999:
Conditional rules can be bypassed when only a field used by the condition is updated — validation is skipped if the constrained field itself is not present in field_names.

For example:

1. Create a rule on res.partner.ref with max = 5 and condition [('is_company', '=', True)].
2. Save ref = "123456" on a non-company partner. This correctly succeeds because the rule does not apply yet.
3. Change only is_company to True. The form write sends only {'is_company': True}.
4. At this point, field_names == {'is_company'} and does not contain ref, so the error rule is silently skipped even though the record now matches the condition and ref violates the rule.

AungKoKoLin1997:
> Conditional rules can be bypassed when only a field used by the condition is updated — validation is skipped if the constrained field itself is not present in field_names.

@kanda999 I updated the code to retrigger for the case.

base_field_length_constraint/security/base_field_length_rule_security.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2026 Quartile
+    License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
```
kanda999, line 3:
```suggestion
```

base_field_length_constraint/views/base_field_length_rule_views.xml
```diff
@@ -0,0 +1,160 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2026 Quartile
+    License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
```
kanda999, line 3:
```suggestion
```

kanda999 (approved):
Code and functioanl review

PR #1322, opened 2026-08-04
[18.0][IMP] base_export_manager: Add properties compatibility

victoralmau:
Add properties compatibility

Use case example:
- Create a property (for example, in a task)
- In the task list, try to export data, select a property field, and add it
- Create a new export template and save

@Tecnativa TT63959


carlos-lopez-tecnativa (approved):
LGTM

pedrobaeza (approved):
/ocabot merge patch

PR #1326, opened 2026-08-26
[18.0][IMP] base_export_manager: allow denying export access by default

smorita7749:
## Summary

Adds a "Deny export access by default" setting under *Settings > General Settings > Permissions*
- When enabled, newly created `ir.model.access` rows get `perm_export = False` by default, instead of the current hardcoded `True`.

This does not change the core `base.group_allow_export` behavior, and `perm_export` remains a UI-only restriction on the Export button (not enforced in `ir.model.access.check()`), same as before this change.

Assisted-by: Claude Sonnet 5
@qrtl QT6643


base_export_manager/readme/CONFIGURE.md
```diff
@@ -4,3 +4,16 @@
 - Uncheck the "Export Access" box on the object of your choice and save
 
 You can also go to Settings \> Technical \> Security \> Access Rights.
+
+To change the default value of "Export Access" for newly created access
+rights:
+
+- Go to Settings \> General Settings \> Export Manager.
```
kanda999: Image

I think the explanation in this section is incorrect. 

Could you please correct the code or the explanation?

base_export_manager/__manifest__.py
```diff
@@ -8,12 +8,13 @@
     "name": "Manage model export profiles",
     "category": "Personalization",
     "version": "18.0.1.0.3",
```
kanda999:
```suggestion
    "version": "18.0.1.1.0",
```

kanda999 (changes requested):

base_export_manager/readme/CONFIGURE.md
smorita7749: Moved it to Permissions section and updated README as well.

base_export_manager/models/res_config_settings.py
```diff
@@ -0,0 +1,12 @@
...
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    export_denied_by_default = fields.Boolean(
+        string="Deny export access by default",
+        config_parameter="base_export_manager.export_denied_by_default",
```
AungKoKoLin1997, line 11: Isn't better to have a relation with company?

smorita7749: Do you mean making it a company field (`related="company_id...."`) instead of a system parameter?
`ir.model.access` applies to all companies, so even if this setting was per company, the result would still be shared across all companies.

Please let me know if I am missing something.

AungKoKoLin1997 (commented):
LG 👍

PR #1327, opened 2026-08-26
[18.0][IMP] base_export_manager: restrict export from pivot view

smorita7749:
The "Export Access" access right already hides the list view's Export
action, but the pivot view's "Download xlsx" button remained visible
and usable regardless of that setting. This PR hides it the same way,
based on the models the current user is allowed to export.

Assisted-by: Claude Sonnet 5


yostashiro:
@smorita7749 Can you squash commits please.

PR #1331, opened 2026-09-04
[19.0][ADD] module list_write_date: add write_date to all list views

bealdav:
ADD module list_write_date: add write_date to all list


tonygalmiche (approved):
Tested on a local Odoo 19.0 instance:
- Installed the module, write_date column appears (hidden by default, toggleable via the optional columns selector) on standalone list views opened from a menu action.
- No duplicate field added on views already declaring write_date explicitly (e.g. stock.quant editable list).
Works as expected. Approving.

Possible improvements (not blocking):
- Not applied to x2many subviews (inline, e.g. invoice/order lines, or non-inline, e.g. ir.model > Views tab), since their resolution bypasses get_view() (uses the private _get_view() directly, or embeds inline arch). Would need overriding _get_view() + recursively injecting into nested <list> nodes.
- A write_date already present as column_invisible="True" is treated as "handled" and skipped, even though it stays permanently hidden.
- No automated tests.
- Could optionally cover create_date/create_uid/write_uid too, each toggleable via a config parameter.

rvalyi:
May be it's a bit invasive... Some ideas:

- an ir.config.parameter (or company based) to active/deactivate the feature
- a group to see it?
- model flags to enable or disable it per model

bealdav:
@rvalyi not so invasive I think, write date isn't seen by default. You may not install this module ;-)
And module can be uninstalled without any data loss
Do you approve it or not ?

@tonygalmiche Thanks for your review I finalized the full perimeter of the module adding test.

tonygalmiche (approved):
Thanks for the update and for documenting the one2many scope limitation in the README, that's clear enough. LGTM.

PaulGoubert (approved):
LGTM

sebastienbeau (approved):
LGTM (code review)

PR #1333, opened 2026-09-15
[FIX] date_range_account: repair corrupted Turkish translation

volkantasci:
**Description of the issue**

The Turkish translation of `date_range_account` is corrupted: several non-ASCII characters have been replaced with `??`, for example:

- `Tarih Aral??klar??` instead of `Tarih Aralıkları`
- `Tarih Aral?????? T??rleri` instead of `Tarih Aralığı Türleri`
- `Tarih Aral??klar?? Olu??tur` instead of `Tarih Aralıkları Oluştur`

As a consequence, the date range menu entries are displayed with broken characters in the Turkish user interface.

**What this PR does**

Restores the missing Turkish characters in `i18n/tr.po` (including the `Last-Translator` header). It only changes translation strings; no code or functional behavior is modified.

**Possible root cause**

This looks like an encoding issue in the tooling that generated the file (Weblate, 2025-07-22). It might be worth fixing the corresponding entries on Weblate as well, so the correction is not lost on the next automatic synchronization.

Thanks in advance for your review.


CristianoMafraJunior (changes requested):
To translate, you have to use Weblate, here's the link
https://translation.odoo-community.org/projects/server-ux-18-0/

volkantasci:
Closing this in favour of fixing the strings on Weblate, as @CristianoMafraJunior suggested.

I checked where the corruption actually comes from: the Turkish file downloaded from Weblate itself for this component contains the same `??` sequences (e.g. `Tarih Aral??klar??`). The broken text is therefore stored in Weblate and pushed into this repository, so a direct `.po` change here would be overwritten by the next Weblate sync.

The 19.0 component to fix is https://translation.odoo-community.org/projects/server-ux-19-0/server-ux-19-0-date_range_account/ — I will correct the Turkish strings there so the fix lands through Weblate. (The link in the review points to the 18.0 project, while this PR targets 19.0.)

Thanks for the pointer!

PR #1334, opened 2026-09-16
[18][base_tier_validation][FIX] Error on accessing simple form view, without header or sheet

florian-dacosta:
If the form view does not contain header, it does not contains the `validation_status` field, but the fields still have the condition `readonly="validation_status..."`, giving an owl error `validation_status` is not defined.

A simple way to reproduce:
Install partner_tier_validation
Install sale_management
Create a sale order with a customer that does not have an email.
Click on button "send by email". Odoo will open a small form view to set partner email and will crash doing it.


LoisRForgeFlow (commented):
@florian-dacosta Thanks for the PR, could you add a test case to avoid future regressions?

rvalyi (approved):

LoisRForgeFlow (approved):
Thanks!

/ocabot merge patch

PR #1336, opened 2026-09-18
[FIX] server_action_mass_edit: bind the wizard's dynamic fields

moylop260:
`onchange()` adds one throwaway field per mass edit line to the wizard's own `_fields`, calls `super()`, and removes them again. Those fields are built outside a class body, so `Field.__set_name__()` never runs on them and their `model_name` stays `None`.

`_fields` is a class attribute shared by every request the worker serves, and `Registry._field_triggers` resolves the dependencies of every field of every model in a single pass that starts with `registry[field.model_name]`. While the dynamic fields are installed, that pass ends in `KeyError: None`, and it does so on whatever record the next `create()` or `write()` touches, not on the wizard.

The removal was not protected either, so an onchange that failed left the fields behind, and every later write kept raising until the registry was reloaded.

What it looks like in production: the wizard is nowhere in the traceback. This one comes from a public donation page of an 18.0 instance, several hours after someone had used mass edit, and it was raised for 45 people before the worker recycled.

    File "odoo/http.py", line 2585, in __call__
    File "odoo/http.py", line 2141, in _serve_ir_http
    File "odoo/addons/base/models/ir_http.py", line 333, in _dispatch
      [...the addon's own controller and create() overrides...]
    File "addons/website_sale/models/website.py", line 439, in
        sale_get_order
      sale_order_sudo = SaleOrder.with_user(SUPERUSER_ID).create(...)
    File "odoo/models.py", line 5052, in create
      records = self._create(data_list)
    File "odoo/models.py", line 5289, in _create
      records.modified(self._fields, create=True)
    File "odoo/models.py", line 7224, in modified
      todo = [self._modified([self._fields[fname] ...], create)]
    File "odoo/models.py", line 7275, in _modified
      tree = self.pool.get_trigger_tree(fields, select=select)
    File "odoo/modules/registry.py", line 416, in get_trigger_tree
      if field in self._field_triggers
    File "odoo/tools/func.py", line 42, in __get__
      value = self.fget(obj)
    File "odoo/modules/registry.py", line 510, in _field_triggers
      dependencies = list(field.resolve_depends(self))
    File "odoo/fields.py", line 800, in resolve_depends
      Model0 = registry[self.model_name]
    File "odoo/modules/registry.py", line 234, in __getitem__
      return self.models[model_name]
    KeyError: None

How to reproduce it:

1. Install `server_action_mass_edit`.
2. In Settings > Technical > Actions > Server Actions, create an action on Contact with type "Mass Edit Records" and one line on the field Notes (`comment`).
3. Put the worker in the state it is in right after a registry reload, where the trigger tree is not cached yet and the next caller is the one that builds it. In a shell: `env.registry.__dict__.pop("_field_triggers", None)`.
4. Open the wizard, which is what installs the dynamic fields: pick contacts in the list view and run the action. It raises the traceback above, so the fields are never removed.
5. Create or write any record at all, on any model, for instance a new contact. It raises `KeyError: None` too, and keeps doing so until the registry is reloaded.

With this commit step 4 succeeds, step 5 succeeds, and `_fields` is left with no field whose `model_name` is `None`.

Bind the fields to the model before installing them, and remove them in a `finally` block. Two regression tests cover both halves. 19.0 is not affected: it no longer touches `_fields`.


legalsylvain:
could you credit the IA, respecting the OCA rules ? thanks !

